PR #7, opened 2016-03-28
Added states for setting up old nodes with Fluentd

blarghmatey:
- Add the brightbox PPA for newer ruby versions
- Install up-to-date ruby version with dev dependencies for installing gems
- Restart syslog after updating configuration


bdero:
:+1:

PR #18, opened 2016-03-29
Added state for resizing EBS root partition

blarghmatey:
On Debian AMIs the root partition is automatically set to the default
size of 8 GB, regardless of the actual EBS volume capacity. In order to
take advantage of the full volume it is necessary to resize the
partition and filesystem.


bdero:
:+1:

PR #19, opened 2016-03-29
Updated profile and orchestration for ES capacity

blarghmatey:
Increased the root volume size on the elasticsearch profile to
accomodate a larger volume of log data. Added the state to expand the
root partition to the orchestration state for running on the
elasticsearch nodes.


bdero:
:+1:

PR #24, opened 2016-04-01
Run datadog formula for all hosts

bdero:


salt/top.sls
```diff
@@ -1,5 +1,7 @@
 base:
-  '*': []
+  '*':
+    - match: grain
```
blarghmatey: The match parameter is not needed here since the match is being done with globbing.

bdero:
@blarghmatey Oops, I accidentally merged before you gave a thumbs up. I addressed your comment - If you have other suggestions I can either revert this or fix in a new PR.

blarghmatey:
:+1:

PR #54, opened 2016-06-24
Ignore the formulas directory

bdero:


blarghmatey:
👍 
Where is the formulas directory coming from?

bdero:
On our salt master, there exists `/home/admin/salt-ops/formulas/master-formula`. I'm not sure if it's there arbitrarily; if it is, we should probably just close this PR.

blarghmatey:
On the master it is there because of the bootstrap script, but there shouldn't be any git operations happening on the master anyway aside from a pull.

PR #57, opened 2016-06-30
Updated bootstrap scripts for salt master

blarghmatey:
**Updated bootstrap scripts for salt master**
Updated the bootstrap script for salt master to make it a bit more streamlined.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,19 +1,19 @@
 #!/bin/bash
 
-sudo apt-get -y install git
-mkdir formulas
-git clone https://github.com/mitodl/master-formula formulas/master-formula
-sh formulas/master-formula/scripts/build.sh
-sudo cp -r pillar/* /srv/pillar
-sudo cp -r salt/* /srv/salt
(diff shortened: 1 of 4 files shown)


bootstrap.sh
```diff
@@ -1,19 +1,19 @@
...
-# local_repo:
-#   url: file:///srv/spm" | sudo tee /etc/salt/spm.repos.d/local_repo.conf
-# sudo spm update_repo
-# sudo spm install master
-sudo cp -r formulas/master-formula/master /srv/salt
+sudo apt-get -y install curl
+if [ -z `which salt-minion` ]
+then
+    curl -o install_salt.sh -L https://bootstrap.saltstack.com
+    sudo sh install_salt.sh -M -U -Z -L -P -A 127.0.0.1 stable
```
bdero: Is the -P flag recent? It seems like the bootstrap script used to always install a bunch of Python packages to the system Python environment using pip, even though there were system packages available for them - like jinja2.

bdero:
After merge conflict resolved, :+1:

PR #59, opened 2016-07-01
Updated mechanism for handling cloud profiles

blarghmatey:
**Updated mechanism for handling cloud profiles**
Moved all cloud profiles to `orchestrate/aws/cloud_profiles` directory
and updated orchestrate states to do a `file.managed` on them to place
them in the `/etc/salt/cloud.profiles.d` directory at the beginning of
the run.

@bdero


bdero:
:+1: 
Is this the last step for not having to run the bootstrap script while iterating on salt-ops?

blarghmatey:
I believe so, yes

PR #60, opened 2016-07-07
Orchestration for edX apps

bdero:


cloud.profiles.d/edx.yml
```diff
@@ -0,0 +1,21 @@
+edx:
```
blarghmatey: Should move this to the `orchestrate/aws/cloud_profiles` directory (I may be misremembering the proper path).

cloud.profiles.d/edx.yml
```diff
@@ -0,0 +1,21 @@
...
+  provider: mitx
+  size: m3.xlarge
```
blarghmatey: Can we get away with running on t2.large? Even if we have up to 6 of them it's still cheaper than 3 of the xlarge.

salt/orchestrate/edx/init.sls
```diff
@@ -0,0 +1,89 @@
...
+          - edx
+        securitygroupid: {{ salt.boto_secgroup.get_group_id(
+            'edx-dogwood_qa', vpc_name='Dogwood QA') }}
+        subnetids: {{ subnet_ids }}
+
+ensure_instance_profile_exists_for_edx:
+  boto_iam_role.present:
+    - name: edx-instance-role
+
+deploy_logging_cloud_map:
```
blarghmatey: Need to update the id here :)

salt/orchestrate/edx/init.sls
```diff
@@ -0,0 +1,89 @@
...
+    tgt='roles:edx', fun='grains.item', tgt_type='grain'
+    ).items() %}
+update_edx_instance:
+  boto_ec2.instance_present:
+    - vpc_name: 'Dogwood QA'
+    - instance_id: {{ grains[1]['ec2:instance_id'] }}
+    - instance_profile_name: edx
+    - security_group_names: edx-dogwood_qa
+    - target_state: running
+{% endfor %}
```
blarghmatey: We can get rid of this block given the use of the templated map file.

salt/orchestrate/edx/init.sls
```diff
@@ -0,0 +1,89 @@
...
+
+resize_root_partitions_on_edx_nodes:
+  salt.state:
+    - tgt: 'G@roles:edx and G@environment:dogwood-qa'
+    - tgt_type: compound
+    - sls: utils.grow_partition
+    - require:
+        - salt: deploy_edx_cloud_map
```
blarghmatey: No longer necessary with the updated Debian AMI.

blarghmatey:
We should move the edx state, templates, and script into an edx state directory (e.g. `salt-ops/salt/edx/...`) and then reference that state from the orchestrate plan. That will make it a bit clearer and maintain the use of the orchestrate directory for purely orchestration.

salt/orchestrate/edx/deploy.yml
```diff
@@ -0,0 +1,103 @@
...
+  virtualenv.managed:
+    - name: {{ venv_path }}
+    - requirements: {{ repo_path }}/requirements.txt
+    - require:
+      - git: clone_edx_configuration
+
+place_sandbox_ansible_configuration:
+  file.managed:
+    - name: {{ conf_file }}
+    - source: salt://orchestrate/edx/templates/ansible_env_config.yaml.j2
```
blarghmatey: This should be relocated to `salt-ops/salt/edx/...` along with this sls file.

salt/orchestrate/edx/templates/ansible_env_config.yml.j2
```diff
@@ -0,0 +1,533 @@
...
+    {% for key, value in xqueue.queues.items() %}
+      {{ key }}: {{ value }}
+    {% endfor %}
+{% raw %}
+  'RABBIT_VHOST': '/prod'
+  'XQUEUE_WORKERS_PER_QUEUE': 2
+  'LOGGING_ENV' : 'prod'
+  'LOG_DIR' : '/logs'
+  'SYSLOG_SERVER' : "{{ SYSLOG_SERVER }}"
+  'RABBIT_HOST' : 'localhost'
```
blarghmatey: Is this node actually going to have a local rabbit service?

salt/orchestrate/edx/templates/ansible_env_config.yml.j2
```diff
@@ -0,0 +1,533 @@
...
+      KEY_PREFIX: "integration_mongo_metadata_inheritance"
+      LOCATION: "/var/tmp/mongo_metadata_inheritance"
+      TIMEOUT: 300
+    staticfiles:
+      BACKEND: "django.core.cache.backends.memcached.MemcachedCache"
+      KEY_FUNCTION: "util.memcache.safe_key"
+      KEY_PREFIX: "integration_static_files"
+      LOCATION:
+        - "localhost:11211"
+  CELERY_BROKER_HOSTNAME: "localhost"
```
blarghmatey: Is this just a default value that gets overridden somewhere else? (I haven't read through the whole file yet...)

salt/orchestrate/edx/templates/ansible_env_config.yml.j2
bdero: Ah, good catch. This should be changed to the consul service DNS since we won't be using haproxy.

cloud.profiles.d/edx.yml
bdero: We're using xlarge currently in production, but we can try having lighter instances like this.

Note that this is the current way that the number of gunicorn workers is calculated: https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/templates/lms_gunicorn.py.j2#L20
Currently in production, `worker_core_mult.lms` is set to 4.

My biggest concern would be the celery workers (defined in this file), which need to be scaled back for this - we've had to scale them back a couple of times in the past due to memory constraints on our current instances.

salt/orchestrate/aws/map_templates/edx.yml
```diff
@@ -0,0 +1,14 @@
+edx:
+  {% for id_num in range(3) %}
+  - edx-{{ environment_name }}-{{ id_num }}:
+      network_interfaces:
+        - DeviceIndex: 0
+          AssociatePublicIpAddress: True
+          SubnetId: {{ subnetids[id_num] }}
```
blarghmatey: This needs to be something like
```salt
{{ subnetids[id_num % len(subnetids)] }}
```
to prevent index out of bounds errors.

blarghmatey:
👍

bdero:
:tada:

PR #61, opened 2016-07-08
Updated to deploy consul first

blarghmatey:
**Updated to deploy consul first**
- Moved edx state under `salt/edx`
- Added NFS package to allow for mounting EFS
- Added deploy of Consul agent before deploying EdX so that DNS
  resolution will work properly during deploy.

@bdero


bdero:
:+1:

PR #62, opened 2016-07-08
Added EFS mount for course data

blarghmatey:
**Added EFS mount for course data**
Added state to mount EFS at `/mnt/data` for course assets.


bdero:
:+1:

PR #63, opened 2016-07-08
Moved templates to `salt/edx/templates`

blarghmatey:
**Moved templates to `salt/edx/templates`**
Moved the templates for the edx state to be under `salt/edx/templates`
instead of orchestrate. Updated references in salt state.


bdero:
:+1:

PR #64, opened 2016-07-08
Moved profile and added file.manage state for it

blarghmatey:
**Moved profile and added file.manage state for it**
Moved handling of the cloud profile to the orchestrate state to ensure
that the profile is in place before trying to execute the map file.


bdero:
:+1:

PR #70, opened 2016-07-15
Added orchestrate routine for launching ES

blarghmatey:
**Added orchestrate routine for launching ES**
In order for Forums to work we need an Elasticsearch cluster. This will
launch a three node cluster using EC2 discovery.


bdero:
:+1:

PR #71, opened 2016-07-15
Added config for elasticsearch

blarghmatey:
**Added config for elasticsearch**


salt/edx/templates/ansible_env_config.yml.j2
```diff
@@ -550,6 +550,10 @@ FORUM_MONGO_HOSTS:
   - "{{ FORUM_MONGO_HOST }}"
 FORUM_API_KEY: "{{ COMMENTS_SERVICE_KEY }}"
 
+EDXAPP_ELASTIC_SEARCH_CONFIG:
+  - host: elasticsearch.service.consul
+    port: 9200
```
bdero: I think EDXAPP_ELASTIC_SEARCH_CONFIG is used for stuff like the courseware search feature, which we don't have turned on at the moment. Is this the ES config value that's used in the forums role?

bdero:
:+1:

PR #75, opened 2016-07-19
Added gitreload state

bdero:
@blarghmatey This PR adds a state for installing gitreload, which is intended to be run on edX instances.

### DIFF
--- a/salt/edx/gitreload.sls
+++ b/salt/edx/gitreload.sls
@@ -33,3 +33,9 @@ create_gitreload_config:
     - group: www-data
     - contents: {{ salt.pillar.get('edx:gitreload:gr_env') | json }}
     - makedirs: True
+
+install_gitreload:
+  pip.install:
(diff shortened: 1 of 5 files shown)


salt/edx/gitreload.sls
```diff
@@ -0,0 +1,84 @@
...
+  cmd.run:
+    - name: ssh-keyscan {{ item }} >> /var/www/.ssh/known_hosts
+    - runas: www-data
+    - require:
+      - file: create_empty_known_hosts
+{% endfor %}
+
+create_gitreload_config:
+  file.managed:
+    - name: {{ salt.pillar.get('edx:gitreload:gr_dir') }}/gr.env.json
```
blarghmatey: If you can specify the default value as the second parameter to the `pillar.get` function that would be useful. That way the pillar is just for overriding the defaults.

bdero:
I addressed the comment, moved the pillar data grabs to the top of the file, and added a comment to the slight known_hosts hack to explain the empty `context`.

blarghmatey:
👍

PR #80, opened 2016-07-21
Updated pillar references, some simplifications

blarghmatey:
**Updated pillar references, some simplifications**
- Simplified handling of ssh_known_hosts
- Moved pillar data for SSH settings to be directly under edx

@bdero


bdero:
:+1: After rebase

PR #81, opened 2016-07-21
Adding routines for enabling git export

blarghmatey:
In order for git export to work we need to make sure that:
- [x] The target directory exists
- [ ] Git is the proper version
- [x] The private SSH key for the user with auth for github.mit.edu is present
- [x] The host keys for the target servers are present


blarghmatey:
@bdero this is ready for review.

bdero:
:+1:

PR #100, opened 2016-07-22
Updating mongo metadata cache to use memcache

blarghmatey:
Updating mongo_metadata_inheritance cache to use memcached instead of
file based in an attempt to fix pickle errors.


bdero:
Oh yes, I remember this. It seems ok to just try this with LMS, but we might want to make sure this is set for CMS as well :+1:

blarghmatey:
Ah, forgot about that split in the settings. I'll update that too.

bdero:
Cool, looks good to me :+1:

PR #101, opened 2016-07-22
Added preview domain to ELB settings

blarghmatey:
**Added preview domain to ELB settings**

@bdero


bdero:
My only slight concern with this is how the yaml merge will behave and the possibility that the default var actually won't make it into the FEATURES list. I think it will because it's dictionaries all the way down, but it's something to watch out for during the next run of this. :+1:

PR #102, opened 2016-07-22
Updated mongo settings to use replicaset

blarghmatey:
**Updated mongo settings to use replicaset**
Updated the MongoDB connection options to specify the replica set and
read preference for connecting to MongoDB so that DNS round robin will
work via Consul while still having a reliable connection to Mongo from EdX.

@bdero

### DIFF
--- a/salt/edx/templates/ansible_env_config.yml.j2
+++ b/salt/edx/templates/ansible_env_config.yml.j2
@@ -115,8 +115,8 @@ lms_auth_config:  &lms_auth
       password: "{{ MONGODB_CONTENTSTORE_PASSWORD }}"
       port: "{{ MONGODB_PORT }}"
       user: "{{ MONGODB_CONTENTSTORE_USER }}"
-    ADDITIONAL_OPTIONS:
       replicaset: "{{ MONGODB_REPLICASET }}"
+      read_preference: "PRIMARY_PREFERRED"
     DOC_STORE_CONFIG:
       db: "{{ MONGODB_CONTENTSTORE_DB }}"
       host:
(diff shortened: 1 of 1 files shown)


salt/edx/templates/ansible_env_config.yml.j2
```diff
@@ -115,8 +115,8 @@ lms_auth_config:  &lms_auth
       password: "{{ MONGODB_CONTENTSTORE_PASSWORD }}"
       port: "{{ MONGODB_PORT }}"
       user: "{{ MONGODB_CONTENTSTORE_USER }}"
-    ADDITIONAL_OPTIONS:
```
bdero: Why is this not required anymore? I think ADDITIONAL_OPTIONS is passed on as the kwargs to the pymongo constructor in the modulestore.

blarghmatey: When I was looking through it appears that it gets that from the OPTIONS settings. I didn't find a reference to ADDITIONAL_OPTIONS outside of one location in the config repo, though I could have missed something.

My thinking is that we can try it with just putting the settings in the `OPTIONS` block and if it doesn't pick it up then we can add back in the `ADDITIONAL_OPTIONS`.

blarghmatey: Digging a bit deeper it appears as though `ADDITIONAL_OPTIONS` just gets merged back into `OPTIONS` anyway. https://github.com/mitodl/edx-platform/blob/mitx/dogwood.3-1rc1/common/lib/xmodule/xmodule/contentstore/django.py#L27

bdero: Thanks for looking into it, I must have missed this when I was trying to fix the mongo connection issue.

bdero:
:+1:

PR #103, opened 2016-07-22
Updated preview domain

blarghmatey:
**Updated preview domain**
Changed `.` to `-` because of issues with second-level subdomains on the
wildcard certificate.

@bdero


bdero:
:+1:

PR #109, opened 2016-07-26
Consolidated common vars using YAML references

blarghmatey:
**Consolidated common vars using YAML references**
- Consolidated duplicate variables from LMS and CMS configs to common
  bases and then added those common settings into the specific sections
  using YAML reference syntax.

- Added MongoDB replicaset configs to common mongo settings.

### DIFF
--- a/salt/edx/templates/ansible_env_config.yml.j2
+++ b/salt/edx/templates/ansible_env_config.yml.j2
@@ -96,9 +96,26 @@ edxapp_workers:
     monitor: False
     max_tasks_per_child: 1
 
+
+
+edxapp_generic_doc_store_config: &edxapp_generic_default_docstore
+  db: "{{ EDXAPP_MONGO_DB_NAME }}"
+  host:
+    - "{{ MONGODB_HOST }}"
+  password: "{{ EDXAPP_MONGO_PASSWORD }}"
+  port: "{{ MONGODB_PORT }}"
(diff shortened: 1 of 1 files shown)


salt/edx/templates/ansible_env_config.yml.j2
```diff
@@ -238,37 +294,44 @@ lms_env_config:  &lms_env
...
   GRADES_DOWNLOAD:
     'STORAGE_TYPE': 'S3'
     'BUCKET': "{{ AWS_GRADES_BUCKET_NAME }}"
     'ROOT_PATH': "{{ AWS_GRADES_ROOT_PATH }}"
-  LMS_BASE: "{{ LMS_NAME }}"
-  LOCAL_LOGLEVEL: "DEBUG"
   LOGGING_ENV: "{{ LMS_LOG_ENV }}"
   LOG_DIR: "{{ LMS_LOG_DIR }}"
-  MEDIA_URL: ""
   FEATURES:
```
bdero: Nit: These feature flags are mostly have to be the same between LMS and CMS, so moving these out to `generic_env_config` or a separate hash merged in (if necessary) can trim more repetition.

bdero:
I didn't check line-by-line to make sure every piece of info is still in the config, but this seems like the right way to break down and compose these configs. :+1:

PR #113, opened 2016-07-28
Added script for using with Consul check

blarghmatey:
**Added script for using with Consul check**
In order to verify whether a node is master, added a script to run via
Consul. This will allow us to expose a service of
`master.mongodb.service.consul` for easier connection to replica sets.


bdero:
:+1:

PR #118, opened 2016-07-28
Install custom theme

bdero:
@blarghmatey


salt/edx/prod.sls
```diff
@@ -117,6 +120,17 @@ mount_efs_filesystem_for_course_assets:
...
+{% if theme_name is not none %}
+install_edxapp_theme:
+  git.latest:
+    - name: {{ theme_repo }}
+    - branch: {{ theme_branch }}
+    - target: /edx/app/edxapp/themes/{{ theme_name }}
+    - user: www-data
+    - require_in:
+      - cmd: run_ansible
+{% endif %}
```
blarghmatey: This will likely require a state to create the directory path. You should be able to just add:
```salt
  file.directory:
    - name: /edx/app/edxapp/themes
    - makedirs: True
    - user: www-data
    - group: www-data
```

bdero: Done, thanks.

blarghmatey:
👍

PR #119, opened 2016-07-29
Added python 2.7 ppa back in

blarghmatey:
**Added python 2.7 ppa back in**
In order to allow for a full unattended build from scratch we need to
have a more up to date version of Python installed. Added the deadsnakes
PPA back in and updated the 'common_debian_pkgs' so that version 2.7.12
will be used for everything.

This had been removed because of suspicions that it was causing SciPy to
fail to build but that was not actually attributable to the newer
version of Python.


blarghmatey:
Merging this in because it was already tested previously and found working.

PR #120, opened 2016-08-01
Added template override for course static assets

blarghmatey:
**Added template override for course static assets**
Added template override for static-files fragment so that our courses
will render properly.

### DIFF
--- a/salt/edx/files/nginx_static_assets.j2
+++ b/salt/edx/files/nginx_static_assets.j2
@@ -1,0 +1,20 @@
+  location ~ /static/((?P<dir>[^/]+)/(?P<file>.*)|(?P<rfile>[\w\d\-.]+)) {
+    root {{ edxapp_app_dir }};
+    try_files /staticfiles/$dir/$file /course_static/$dir/static/$file /staticfiles/$rfile =404;
+
+    # return a 403 for static files that shouldn't be
+    # in the staticfiles directory
+    location ~ ^/static/(?:.*)(?:\.xml|\.json|README.TXT) {
+        return 403;
+    }
+    # Set django-pipelined files to maximum cache time
+    location ~ "/static/(?P<collected>.*\.[0-9a-f]{12}\..*)" {
+        expires max;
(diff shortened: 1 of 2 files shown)


salt/edx/prod.sls
```diff
@@ -159,6 +167,26 @@ run_ansible:
       - virtualenv: create_ansible_virtualenv
```
bdero: This should require `file: replace_nginx_static_asset_template_fragment`

salt/edx/files/nginx_static_assets.j2
```diff
@@ -0,0 +1,20 @@
+  location ~ /static/((?P<dir>[^/]+)/(?P<file>.*)|(?P<rfile>[\w\d\-.]+)) {
+    root {{ edxapp_app_dir }};
+    try_files /staticfiles/$dir/$file /course_static/$dir/static/$file /staticfiles/$rfile =404;
```
bdero: This line needs to have `/prod_course_static/$dir/static/$file` if it's a live instance, and `/course_static/$dir/static/$file` if it's a draft instance.

In fact, we can get away with one symlink here and have it point to the right place. I will leave another comment with details.

salt/edx/prod.sls
```diff
@@ -159,6 +167,26 @@ run_ansible:
...
+
+create_course_prod_asset_symlink:
+  file.absent:
+    - name: /edx/var/edxapp/prod_course_static/
+  file.symlink:
+    - name: /edx/var/edxapp/prod_course_static
+    - target: /mnt/data/prod_repos
+    - makedirs: True
+    - user: edxapp
+    - group: edxapp
```
bdero: We can get away with having one symlink. The only reason we had two in our old stack was because the hosts in our old stack serving static assets were the load balancers, which serve both "draft" and "live".

Maybe just set the target to `salt.pillar.get('edx:edxapp:GIT_REPO_DIR')` and use the name `/edx/var/edxapp/course_static/` both here and in the template.

PR #121, opened 2016-08-02
Changed directory ownership for custom theme

blarghmatey:
**Changed directory ownership for custom theme**

@bdero


bdero:
Currently the modes of all directories and files leading up to the themes directory is 755 and 644 respectively, so the user of the directory shouldn't matter.

PR #126, opened 2016-08-04
Added proper path for xqueuewatcher requirements

blarghmatey:
**Added proper path for xqueuewatcher requirements**
As of the end of 2015, the requirements file for xqueue watcher is now
located at `<src_dir>/requirements/prod.txt` instead of `<src_dir>/requirements.txt`

@bdero


bdero:
:+1:

PR #151, opened 2016-08-09
Place gitreload above edx.prod in top.sls

bdero:
This is an attempt to make gitreload execute *before* edx.prod.


blarghmatey:
:+1:

PR #154, opened 2016-08-10
Fix xsiftx errors

bdero:
This PR has fixes necessary for the xsiftx state to run.

### DIFF
--- a/salt/edx/xsiftx.sls
+++ b/salt/edx/xsiftx.sls
@@ -6,7 +6,7 @@
                               'https://github.com/mitodl/sifters') -%}
 {% set extra_sifter_dir = salt.pillar.get('edx:xsiftx:extra_sifter_dir',
                               '/usr/local/share/xsiftx/sifters') -%}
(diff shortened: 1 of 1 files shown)


salt/edx/xsiftx.sls
```diff
@@ -28,11 +28,11 @@
 
 install_xsiftx:
   pip.installed:
-    - name: git+https://{{ repo }}@{{ version }}#egg=xsiftx
+    - name: git+{{ repo }}@{{ version }}#egg=xsiftx
     - exists_action: w
 
-{% if extra_sifter_repo is not None %}
-install_extra_sifters
+{% if extra_sifter_repo is not none %}
```
blarghmatey: You should be able to just do an if statement and use the truth in essence of the variable.

blarghmatey:
:+1:

salt/edx/xsiftx.sls
bdero: Thanks, I'll change this. I tend habitually be more explicit when using Jinja2 due to some weird Ansible behavior with vars.

PR #157, opened 2016-08-11
Run pip as edxapp when installing gitreload

bdero:
Otherwise, it installs the package + it's dependencies as root in the virtualenv!


blarghmatey:
👍

PR #159, opened 2016-08-11
Added google analytics setting

bdero:
Allows setting the Google Analytics tracking ID for client-side tracking on LMS and CMS.


blarghmatey:
👍

PR #164, opened 2016-08-15
Move GA setting from env to auth config

bdero:
I overlooked that `aws.py` grabs the key from `AUTH_TOKENS`. This PR fixes the GA config.


blarghmatey:
:+1:

PR #165, opened 2016-08-15
Added status monitor config for Nginx

blarghmatey:
**Added status monitor config for Nginx**
Added a state to enable status monitoring for nginx so that we can tie
that into Datadog integrations.

@bdero

### DIFF
--- a/salt/edx/prod.sls
+++ b/salt/edx/prod.sls
@@ -224,3 +224,23 @@ add_{{ host.name }}_to_known_hosts_for_edxapp:
     - user: www-data
     - fingerprint: {{ host.fingerprint }}
 {% endfor %}
+
+configure_nginx_status_module_for_edx:
+  file.managed:
+    - name: /etc/nginx/sites-enabled/status_monitor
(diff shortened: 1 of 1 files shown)


salt/edx/prod.sls
```diff
@@ -224,3 +224,23 @@ add_{{ host.name }}_to_known_hosts_for_edxapp:
...
+  file.managed:
+    - name: /etc/nginx/sites-enabled/status_monitor
+    - contents: |
+        server {
+            listen 80;
+            location /nginx_status {
+                stub_status on;
+                access_log off;
+                allow 127.0.0.1;
+                deny all;
```
bdero: So we're only allowing requests from localhost because we plan on only having the datadog agent pinging this for now?

bdero:
:+1:

salt/edx/prod.sls
blarghmatey: Correct, but it's also best practice from a security perspective to not expose this endpoint to external clients without having it locked down.

PR #170, opened 2016-08-17
Updated ELB orchestration to add sticky sessions

blarghmatey:
**Updated ELB orchestration to add sticky sessions**
Updated the orchestration states for the Dogwood ELBs to enable sticky
sessions for HTTP(S) connections.

@bdero

### DIFF
--- a/salt/orchestrate/aws/dogwood_rp_elb.sls
+++ b/salt/orchestrate/aws/dogwood_rp_elb.sls
@@ -8,24 +8,31 @@
 {% endfor %}
 
 {% for edx_type in ['draft', 'live'] %}
+{% set elb_name = 'edx-{0}-dogwood-rp'.format(edx_type) %}
+
 create_elb_for_edx_{{ edx_type }}:
   boto_elb.present:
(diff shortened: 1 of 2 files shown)


salt/orchestrate/aws/dogwood_rp_elb.sls
```diff
@@ -8,24 +8,31 @@
...
         - elb_port: 80
           instance_port: 80
           elb_protocol: HTTP
           instance_protocol: HTTP
+          policies:
+            - {{ elb_name }}-sticky-cookie-policy
     - attributes:
         cross_zone_load_balancing:
           enabled: True
-        sticky_sessions:
```
pdpinch: I'm confused. We're removing sticky sessions?

blarghmatey: No, that was added previously as an attempt to enable them but it doesn't actually do anything. The piece that actually enables them is the `policies` attribute on the listeners. The policy itself is created further down on line 63.

PR #175, opened 2016-09-02
Added orchestration for Micromasters Elasticsearch

blarghmatey:
Added an orchestration state for a Micromasters Elasticsearch cluster. Includes creating a VPC.

### DIFF
--- a/salt/orchestrate/edx/services/elasticsearch.sls
+++ b/salt/orchestrate/edx/services/elasticsearch.sls
@@ -25,6 +25,9 @@ generate_elasticsearch_cloud_map_file:
           - {{ salt.boto_secgroup.get_group_id(
             'salt_master-dogwood_qa', vpc_name='Dogwood QA') }}
         subnetids: {{ subnet_ids }}
+        volume_size: 200
+        tags:
+          escluster: edx-dogwood-rp
     - require:
         - file: load_elasticsearch_cloud_profile
 
@@ -84,7 +87,7 @@ build_dogwood_elasticsearch_nodes:
 
 remove_broken_line_from_elasticsearch_init_script:
   salt.function:
-    - tgt: 'G@roles:elasticsearch and G@environment:dogwood-rp'
+    - tgt: 'G@roles:elasticsearch and G@environment:dogwood-qa'
     - tgt_type: compound
     - name: file.comment_line
     - kwarg:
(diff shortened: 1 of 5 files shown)


salt/orchestrate/edx/services/elasticsearch.sls
```diff
@@ -25,6 +25,9 @@ generate_elasticsearch_cloud_map_file:
           - {{ salt.boto_secgroup.get_group_id(
             'salt_master-dogwood_qa', vpc_name='Dogwood QA') }}
         subnetids: {{ subnet_ids }}
+        volume_size: 200
+        tags:
+          escluster: edx-dogwood-rp
```
pdpinch: Why is this edx-dogwod-rp?

salt/orchestrate/edx/services/elasticsearch.sls
```diff
@@ -84,7 +87,7 @@ build_dogwood_elasticsearch_nodes:
 
 remove_broken_line_from_elasticsearch_init_script:
   salt.function:
-    - tgt: 'G@roles:elasticsearch and G@environment:dogwood-rp'
+    - tgt: 'G@roles:elasticsearch and G@environment:dogwood-qa'
```
pdpinch: why the change from -rp to -qa -- here and on L99?

pdpinch:
Still mostly over my head.

What are we using letsencrypt for?

salt/orchestrate/edx/services/elasticsearch.sls
blarghmatey: Good catch, that should be qa. The reason it got added in here is because I modified the map file template to look for a `tags` variable and I am using the `escluster` tag for the micromasters cluster discovery to allow for multiple clusters in the same VPC so I'm just trying to be consistent.

salt/orchestrate/edx/services/elasticsearch.sls
blarghmatey: That's because this orchestration file is for the instances in the QA environment and they were accidentally targeting RP instances which means that it wasn't performing those last couple of steps on the right instances.

blarghmatey:
Letsencrypt is for generating certificates to be used with the Nginx services that are proxying traffic to the ES servers on the same box. I'm using a domain of `es.micromasters.odl.mit.edu` so the wildcard cert won't work due to the nested subdirectories. I could modify it and use the wildcard, but for a proof of concept I am just using letsencrypt. I've used the formula that is being targeted for some of my own projects and it works well.

pdpinch:
It still seems odd to me that the generic configuration, in salt/orchestrate/edx/services/elasticsearch.sls is for dogwood-qa, but that's irrelevant for the issue at hand which is about getting the micromasters cluster running. 

AFAICT 👍

PR #177, opened 2016-09-07
Added orchestration for xqwatcher

blarghmatey:
**Added orchestration for xqwatcher**
Added state, cloud profile, map, and orchestration for building and
configuring an instance of xqwatcher for running Python 3 graders to be
used by 6.00x courses.


blarghmatey:
Verified working 👍

PR #178, opened 2016-09-13
Added state to modify upload size limit

blarghmatey:
**Added state to modify upload size limit**
Added a state to modify the upload size limit enforced by Nginx to a
configurable value.

@pdpinch


blarghmatey:
Applied changes in QA and validated that they are functioning properly. 👍

Issue #196, opened 2016-12-28
Update automated notifications to use Slack instead of Hipchat


blarghmatey:
There are a couple of notifications that are triggered by automated processes to deliver messages to Hipchat. Update them to use Slack.

- [ ] Elastalert mail delivery failure notifications
      Update pillar data to reflect proper configuration for Slack channel
- [ ] edX backup failures
      Update orchestration to use Slack module for `onfail` action instead of Hipchat

shaidar:
[PR](https://github.com/mitodl/salt-ops/pull/197) submitted and merged.

PR #197, opened 2017-01-04
Changed hipchat post message to slack

shaidar:
A couple of services are still using hipchat notifications and given that the team has moved to Slack, we wanted to make sure that those notifications are now going to slack channels instead. This is related to [Issue #196](https://github.com/mitodl/salt-ops/issues/196). Interesting thing about the change is that we've had to use both a slack bot with its api token and the slack web hook url for Elastalert as this is how that service implemented their slack integration.


blarghmatey:
👍

PR #198, opened 2017-01-05
Updated states to support Eucalyptus release

blarghmatey:
- Updated the edx states to make the ansible run more generic so that we can start decomposing the edxapp servers
- Updated AMI targets for cloud profiles
- Removed ansible vars template and replaced with rendering data from pillar.


salt/edx/run_ansible.sls
```diff
@@ -0,0 +1,73 @@
...
+
+{% set playbooks = salt.pillar.get('edx:playbooks', salt.environ.get('EDX_PLAYBOOKS', ['edx-east/common.yml']).split(',')) -%}
+
+clone_edx_configuration:
+  file.directory:
+    - name: {{ repo_path }}
+    - makedirs: True
+  git.latest:
+    - name: {{ salt.pillar.get('edx:config:repo', 'https://github.com/edx/configuration.git') }}
+    - rev: {{ salt.pillar.get('edx:config:branch', 'named-release/dogwood.3') }}
```
shaidar: Shouldn't we change this to eucalyptus? Wonder if it's easier going forward to define the openedx version variable that we want to install and just use it everywhere.

blarghmatey: That's certainly something worth considering. In this particular case the branch is for the version of the ansible repo to use. I updated the default value to point to eucalyptus as well.

shaidar:
:thumbsup:

PR #199, opened 2017-01-09
Updating security group linkings

blarghmatey:
- Added the Consul security group to have access to the MySQL RDS
  instance for health checking
- Added the `default` security group to edX instances to enable access
  to EFS


shaidar:
:thumbsup:

PR #217, opened 2017-01-19
Updated the edX orchestrate states to DRY them up

blarghmatey:
Extracted a Jinja macro for setting the VPC attributes to simplify
writing of reusable orchestrate states. Consolidated the orchestrate
states to work on multiple VPCs by passing a `VPC_NAME` environment
variable when calling the orchestrate run.

This declutters the repository as well as making it more explicit when
running orchestrate states against the production environment since
the default VPC is MITx QA.


shaidar:
:thumbsup: - Looks good, and was just wondering whether we should also include the aws_env_macro.jinja to [backup.sls](https://github.com/mitodl/salt-ops/blob/master/salt/orchestrate/edx/backup.sls) as well

PR #224, opened 2017-01-28
Updated libgit version

shaidar:
Changed libgit tag to latest version along with its checksum


blarghmatey:
👍

blarghmatey:
Thanks for that. I'm sure it would have bitten us sooner or later.

Issue #187, opened 2016-12-13
Restore production backup data on QA servers


blarghmatey:
Write orchestrate state to reload backup prod data onto QA servers.

PR #227, opened 2017-02-02
Added states and scripts for restoring data from production

blarghmatey:
In order to make QA more useful and help with rapid recovery we need a
way to automate data restoration. This will also serve as backup
verification. This PR adds scripts and states for automatically
loading data backed up from production into our QA environment.


salt/backups/restore.sls
```diff
@@ -0,0 +1,30 @@
...
+
+install_duplicity_backend_requirements:
+  pip.installed:
+    - name: boto
+
+{% for service in salt.pillar.get('backups:enabled', []) %}
+{% if service.get('pkgs') %}
+install_packages_for_{{ service.title }}_backup:
+  pkg.installed:
+    - pkgs: {{ service.pkgs }}
```
shaidar: What are the expected values here?

salt/backups/templates/restore_course_assets.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -e
+
+{% set backupdir = '/mnt/efs/{}'.format(settings.get('directory', 'course_assets')) %}
+
+sudo mkdir -p /mnt/efs
```
shaidar: Is sudo needed?

blarghmatey: Yes, because the /mnt directory is owned by root. It's possible that we could mount it somewhere else, but given that this instance is destroyed as soon as the restore is finished I don't think that it would gain us too much.

shaidar: But aren't those scripts run by the salt-master who's running as root?

salt/backups/restore.sls
blarghmatey: This would be a list of packages that need to be installed for the script to work. e.g. mysql-client, curl, etc.

salt/backups/templates/restore_course_assets.sh
blarghmatey: Fair point. I'll update.

shaidar:
:thumbsup:

blarghmatey:
closes #187

PR #236, opened 2017-02-16
Updating instance sizes to reflect actual usage patterns

blarghmatey:
After viewing the actual usage of disk and memory resources it was
evident that most of the services are running on oversized
instances. Updated the profiles to use smaller and cheaper instance types.

### DIFF
--- a/salt/orchestrate/aws/cloud_profiles/elasticsearch.conf
+++ b/salt/orchestrate/aws/cloud_profiles/elasticsearch.conf
@@ -6,6 +6,9 @@ elasticsearch:
   ssh_interface: private_ips
   block_device_mappings:
     - DeviceName: /dev/xvda
+      Ebs.VolumeSize: 20
+      Ebs.VolumeType: gp2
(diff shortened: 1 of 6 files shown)


salt/orchestrate/aws/cloud_profiles/elasticsearch.conf
```diff
@@ -6,6 +6,9 @@ elasticsearch:
   ssh_interface: private_ips
   block_device_mappings:
     - DeviceName: /dev/xvda
+      Ebs.VolumeSize: 20
+      Ebs.VolumeType: gp2
+    - DeviceName: /dev/xvdb
       Ebs.VolumeSize: 400
```
shaidar: Do we still need the other volume to be that size?

blarghmatey: I left it that size for when we are deploying things like our logging cluster. For other instances we can override that value in the map file.

shaidar:
:thumbsup:

PR #237, opened 2017-02-22
Updated Consul security group to fix DNS

blarghmatey:
There was an intermittent issue with DNS recursor lookups failing due
to the fact that port 8300 was not exposed for UDP traffic. This fixes
that issue so that we should no longer have MySQL lookup errors.

Also updated the CIDR ranges for security groups in QA.


shaidar:
:thumbsup:

PR #238, opened 2017-02-22
Added consul server security group to allowed sources

blarghmatey:
Updated the consul agent security group to also allow traffic from the
Consul servers.


shaidar:
:thumbsup:

Issue #166, opened 2016-08-16
Install and configure Hubblestack


blarghmatey:
hubblestack.io

pdpinch:
?

blarghmatey:
Sorry, I put this in as a placeholder so I could fill it out more.

Hubblestack is a tool built on top of Saltstack for doing security auditing of running systems. It includes things like alerting on file modifications, compliance auditing, etc. https://hubblestack.io/

shaidar:
Ended up using a basic saltstack beacon to do FIM and deployed to rp.

PR #239, opened 2017-03-01
Configure Hubblestack Pulsar

shaidar:
[Issue#166](https://github.com/mitodl/salt-ops/issues/166)

### What does this PR do?
Hubblestack Pulsar is used for file integrity monitoring and in this case, we will be using it for the mitx instances. This PR does the following:
- Created a hubblestack folder that includes a simple state file to install python-pyinotify on the minion
- Added a reactor for edx that posts a slack message anytime there's a modification to a list of watched files

### DIFF
--- a/salt/hubblestack/pulsar_dependency.sls
+++ b/salt/hubblestack/pulsar_dependency.sls
@@ -1,0 +1,3 @@
+install_pulsar_dependency:
+  pkg.installed:
+    - name: python-pyinotify
--- a/salt/hubblestack/pulsar_dependency.sls
+++ b/salt/hubblestack/pulsar_dependency.sls
@@ -1,3 +1,0 @@
-install_pulsar_dependency:
-  pkg.installed:
-    - name: python-pyinotify
--- a/salt/reactors/edx/pulsar_mitx.sls
+++ b/salt/reactors/edx/pulsar_mitx.sls
@@ -1,0 +1,8 @@
+pulsar_mitx:
+  local.slack.post_message:
+    - tgt: {{ data['id'] }}
+    - kwarg:
+       channel: "#devops"
+       message: "Hubblestack Pulsar FMI - Detected change - Host:`{{ data['id'] }}` - File modified:`{{ data['path'] }}` - User:`{{ data['stats']['user'] }}`"
+       from_name: "saltbot"
+       api_key: {{ salt.pillar.get('slack_api_token') }}
--- a/salt/reactors/edx/pulsar_mitx.sls
+++ b/salt/reactors/edx/pulsar_mitx.sls
@@ -1,8 +1,9 @@
 pulsar_mitx:
   local.slack.post_message:
-    - tgt: {{ data['id'] }}
+    - tgt: 'roles:master'
+    - expr_form: grain
     - kwarg:
-       channel: "#devops"
-       message: "Hubblestack Pulsar FMI - Detected change - Host:`{{ data['id'] }}` - File modified:`{{ data['path'] }}` - User:`{{ data['stats']['user'] }}`"
-       from_name: "saltbot"
-       api_key: {{ salt.pillar.get('slack_api_token') }}
+        channel: "#devops"
+        message: "Hubblestack Pulsar FMI - Detected change - Host:`{{ data['id'] }}` - File modified:`{{ data['path'] }}` - Details: `{{ data }}`"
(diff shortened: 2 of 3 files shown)


salt/hubblestack/pulsar_dependency.sls
```diff
@@ -0,0 +1,3 @@
+install_pulsar_dependency:
+  pkg.installed:
+    - name: python-pyinotify
```
blarghmatey: We have a state file in `salt/utils/install_pip.sls` that allows for installation of arbitrary python packages driven by pillar data in the form
```yaml
python_dependencies:
    python_libs:
        - pyinotify
        - someotherpackage
```
I believe that state is already included in the `top.sls` for all of the minions.

salt/reactors/edx/pulsar_mitx.sls
```diff
@@ -0,0 +1,8 @@
+pulsar_mitx:
+  local.slack.post_message:
+    - tgt: {{ data['id'] }}
```
blarghmatey: This will cause the slack message to be posted by the minion which we don't necessarily want. We probably want the message to be sent from the master so that we don't have to add the api token to the pilllar data for all of the minions.

salt/reactors/edx/pulsar_mitx.sls
```diff
@@ -0,0 +1,8 @@
...
+    - kwarg:
+       channel: "#devops"
+       message: "Hubblestack Pulsar FMI - Detected change - Host:`{{ data['id'] }}` - File modified:`{{ data['path'] }}` - User:`{{ data['stats']['user'] }}`"
```
blarghmatey: Typo: `FMI` should be `FIM`

blarghmatey: It might be useful to also post the entirety of the data that is sent from hubblestack so that we don't lose that information.

salt/hubblestack/pulsar_dependency.sls
blarghmatey: The install_pip state is run on all of the minions, but adding the pyinotify package to pillar data can be done on specific minions that we are targeting, which would only install it on those minions that have it in the pillar data.

salt/reactors/edx/pulsar_mitx.sls
shaidar: If u do cmd.slack.post_message: ... Would that run locally or is that running on the minion?

blarghmatey: It's not so much that you're using the `local` prefix, it's the `tgt: {{ data['id'] }}` line that will cause the `post_message` to occur on the minion that triggered the event. You could set it to 
```salt
  - tgt: 'roles:master'
  - tgt_type: grain
```
(I believe, might need to double check the syntax)

salt/utils/install_pip.sls
```diff
@@ -1,6 +1,6 @@
 {% set python_dependencies = salt.grains.filter_by({
     'default': {
-      'python_libs': ['testinfra']
+      'python_libs': ['testinfra', 'python-pyinotify']
```
blarghmatey: This should be the name of the package as it is represented in PyPI. The `python-pyinotify` package is how it is represented in the Debian apt repository. For PyPI it's `pyinotify`

salt/reactors/edx/pulsar_mitx.sls
```diff
@@ -1,8 +1,9 @@
...
     - kwarg:
-       channel: "#devops"
-       message: "Hubblestack Pulsar FMI - Detected change - Host:`{{ data['id'] }}` - File modified:`{{ data['path'] }}` - User:`{{ data['stats']['user'] }}`"
-       from_name: "saltbot"
-       api_key: {{ salt.pillar.get('slack_api_token') }}
\ No newline at end of file
+        channel: "#devops"
+        message: "Hubblestack Pulsar FMI - Detected change - Host:`{{ data['id'] }}` - File modified:`{{ data['path'] }}` - Details: `{{ data }}`"
+        from_name: "saltbot"
+        api_key: {{ salt.pillar.get('slack_api_token') }}
```
blarghmatey: Is this necessary still? Are we going to set the api_key in the minion pillar data?

blarghmatey:
👍

blarghmatey:
I think the last piece is to update the pillar data with the slack API key.

PR #241, opened 2017-03-01
Added state to create Vault roles based on pillar data

blarghmatey:
In order for us to be able to handle multiple vault roles across
different environments and business needs I created a templated state
that will use information from pillar to create the appropriate state
definitions for the various backend roles that are required.

### DIFF
--- a/salt/vault/roles.sls
+++ b/salt/vault/roles.sls
@@ -1,0 +1,22 @@
+{% set roles = salt.pillar.get('vault:roles') %}
+{% for role_id, role in roles.items() %}
+{% for environment in role.environments %}
+{% set env_data = salt.pillar.get('environments:{}'.format(environment), {}) %}
+  {% for purpose in env_data.purposes %}
+create_{{ role.name }}_role_in_{{ role.backend }}_for_{{ purpose }}_in_{{ environment }}:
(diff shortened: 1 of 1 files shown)


salt/vault/roles.sls
```diff
@@ -0,0 +1,18 @@
+{% set roles = salt.pillar.get('vault:roles') %}
+{% for role_id, role in roles.items() %}
+{% for environment in role.environments %}
+{% set environment = salt.pillar.get('environments:{}'.format(environment), {}) %}
+  {% for purpose in env_info.purposes %}
```
shaidar: Where is env_info.purposes coming from?

blarghmatey: That's coming from the different environment blocks in the `environment_settings.sls` pillar file. For example:
```yaml
environments:
  mitx-qa:
    purposes:
      current-residential-live: mitx/dogwood.3-2rc2
      current-residential-draft: mitx/dogwood.3-2rc2
...
```

blarghmatey: I started writing one answer and then realized that `env_info` is a remnant of one of my earlier approaches. Good catch :)

shaidar:
:thumbsup:

PR #247, opened 2017-03-06
Updated restore scripts to use force option

blarghmatey:
Updated the restore scripts to use the `--force` option to fix errors
when loading course assets on QA environment. Also updated the restore
script for Consul to actually perform a restore.


shaidar:
:thumbsup:

PR #250, opened 2017-03-08
Using inotify beacon without pulsar

shaidar:
- Renamed pulsar to inotify
- Added a state to utils to increase number of inotify watches
- Added utils.inotify_watches state to base role
- Deleted utils.grow_partition state as part of cleanup since it was not working


blarghmatey:
👍

Issue #248, opened 2017-03-07
Write tests for status of edX application


blarghmatey:
In order to ensure that our edX instances are healthy we need to be able to run tests to verify the state of the system both at creation and on a periodic schedule so that we can be notified of soft-fail situations.

blarghmatey:
Some things to test:
- [ ] edxapp supervisor process is running
- [ ] edxapp listen port is accessible
- [ ] edxapp instance can reach services including:
  - [ ] RabbitMQ
  - [ ] Elasticsearch
  - [ ] MySQL
  - [ ] Elastic File Store (NFS)
- [ ] edxapp can reach forum service
- [ ] edxapp can reach xqueue endpoint

shaidar:
This is mainly complete. We tried going beyond a simple socket connection to test connectivity to several of the dependent services, however the socket module in testinfra does not support that. We submitted a [PR](https://github.com/philpep/testinfra/pull/190) to the testinfra library, however our use cases is somewhat different and PR will not be merged. For enhanced testing, we might need to revisit [goss](https://github.com/aelsabbahy/goss) and use that going forward.

PR #252, opened 2017-03-15
Tests to verify status and connectivity of edx application

shaidar:
[Issue#248](https://github.com/mitodl/salt-ops/issues/248)

### What does this PR do?
Run tests specified in issue#248 to verify that the edxapp is running as expected. This PR relies on a new module (sockethost) added to the testinfra library and thus needs to be in place for all tests to work properly.

### How to test?
Run state file on one of the edx nodes

### DIFF
--- a/salt/edx/tests/test_edxapp_running.sls
+++ b/salt/edx/tests/test_edxapp_running.sls
@@ -22,7 +22,7 @@ our configurations. Test the following:
     'lms': 'tcp://0.0.0.0:8000',
     'gitreload_service': 'tcp://0.0.0.0:8095',
     'xqueue': 'tcp://0.0.0.0:18040',
-    'forum': 'unix:///edx/var/forum/forum.sock',
+    'forum': 'tcp://0.0.0.0:4567',
     'rabbitmq': 'tcp://rabbitmq.service.consul:15672',
     'elasticsearch': 'tcp://elasticsearch.service.consul:9200',
     'mysql':'tcp://mysql.service.consul:3306',
@@ -51,7 +51,7 @@ test_edxapp_supervisor_{{ sv_service }}:
 
 {% for connection in socket_connections %}
 test_edxapp_{{ connection }}:
-  testinfra.sockethost:
+  testinfra.socket_host:
     - name: {{ connection }}
     - is_listening: True
 {% endfor %}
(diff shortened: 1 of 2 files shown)


salt/edx/tests/test_edxapp_running.sls
```diff
@@ -0,0 +1,86 @@
...
+    'xqueue',
+    'xqueue_consumer'
+  ] %}
+
+{% set socket_connections = {
+    'ssl': 'tcp://0.0.0.0:443',
+    'lms': 'tcp://0.0.0.0:8000',
+    'gitreload_service': 'tcp://0.0.0.0:8095',
+    'xqueue': 'tcp://0.0.0.0:18040',
+    'forum': 'unix:///edx/var/forum/forum.sock',
```
blarghmatey: This should use the TCP port that the forum is listening on since we aren't using the unix socket. You may need to refer to edx/configuration for the appropriate port number.

salt/edx/tests/test_edxapp_running.sls
```diff
@@ -0,0 +1,86 @@
...
+
+test_edxapp_cms:
+  testinfra.supervisor:
+      - name: 'edxapp:cms'
+      - is_running: True
+
+# Check if AWS EFS is mounted
+test_edxapp_efs_mount:
+  testinfra.mountpoint:
+    - name: '/mnt/data'
```
blarghmatey: I would add an `exists: True` here as well.

salt/edx/tests/test_edxapp_running.sls
```diff
@@ -0,0 +1,86 @@
...
+    - filesystem:
+        expected: nfs4
+        comparison: eq
+
+{% for entry in lms_env %}
+test_edxapp_lms_env_{{ entry }}:
+  testinfra.file:
+    - name: '/edx/app/edxapp/lms.env.json'
```
blarghmatey: Add an `exists: True` here as well.

salt/edx/tests/test_edxapp_running.sls
```diff
@@ -0,0 +1,86 @@
...
+    - contains: {{ entry }}
```
blarghmatey: The `for` loop should be more like `for attribute, value in lms_env.items()`. Just doing a for loop through the dictionary will only give you the keys and not the values.

The `contains` entry should look like:
```salt
- contains:
  parameter: {{ value }}
  expected: True
  comparison: is_
```
See https://github.com/mitodl/rabbitmq-formula/blob/master/rabbitmq/tests/test_configure.sls#L11 for an example.

blarghmatey:
👍 Good start :)

PR #256, opened 2017-03-28
Additional options to backup scripts

shaidar:
Included the following new duplicity options:
- archive-dir pointed archive dir to backup disk since that's the larger one
- full-if-older-than doing full backups once a week
- s3-use-ia infrequent access s3 and is geared for backup data


blarghmatey:
Looks like you accidentally included your MySQL db changes, but once that's cleared up 👍

PR #257, opened 2017-03-28
Added documentation about vault key spaces

blarghmatey:
Added a writeup of our discussion about how to manage static secrets
in Vault.


docs/standards/vault.md
```diff
@@ -0,0 +1,64 @@
...
+- micromasters
+
+### Putting It All Together
+Putting all of this together, a keypath can be templated as:
+
+```
+secret-{{ business_unit }}/{{ environment || global }}/{{ purpose? }}/{{ app_name }}-{{ key_name }}
+```
+
+## Data Conventions
```
shaidar: This section is a bit unclear to me. When writing a key-value should the full path not be specified? Same thing for the salt state.

shaidar: Guess what I'm missing is how does the GET request to that URL know what key you are requesting the value for.

blarghmatey: The GET request will just return all of the data for a given path. For example, if the path `secret/foo/bar` has the data `foo=bar` then when you run the GET request you will get back `foo=bar`. If the data for that path is `foo=bar, baz=quux` then that is what gets returned when you send the GET request.

shaidar:
:thumbsup:

PR #258, opened 2017-03-29
Bugfixes

shaidar:
- removed the --s3-use-ia option as this is only available in the newer version of duplicity which is not the one that gets installed from the repo
- fixed a typo in cachedir


blarghmatey:
👍

PR #259, opened 2017-03-29
Updated role state to use new structure for pillar data

blarghmatey:
In order to simplify the management of roles and make them more
flexible I refactored the pillar data and updated the state file to
take advantage of that new structure.

### DIFF
--- a/salt/vault/roles.sls
+++ b/salt/vault/roles.sls
@@ -1,22 +1,12 @@
 {% set roles = salt.pillar.get('vault:roles') %}
 {% for role_id, role in roles.items() %}
-{% for environment in role.environments %}
-{% set env_data = salt.pillar.get('environments:{}'.format(environment), {}) %}
-  {% for purpose in env_data.purposes %}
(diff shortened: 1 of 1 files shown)


salt/vault/roles.sls
```diff
@@ -1,22 +1,12 @@
 {% set roles = salt.pillar.get('vault:roles') %}
 {% for role_id, role in roles.items() %}
-{% for environment in role.environments %}
-{% set env_data = salt.pillar.get('environments:{}'.format(environment), {}) %}
-  {% for purpose in env_data.purposes %}
-create_{{ role.name }}_role_in_{{ role.backend }}_for_{{ purpose }}_in_{{ environment }}:
+create_{{ role_id }}:
```
shaidar: What's the difference between role_id and role.name?

blarghmatey: role_id is the arbitrary keypath in the pillar file that is purely for the purpose of differentiating between other entries in the pillar file. role.name is the actual role name that will be used as part of the URL for retrieving the credentials (e.g. `mysql-mitx-qa/creds/admin`)

shaidar:
:thumbsup:

PR #261, opened 2017-04-03
Added step to backup routines to alert on successful completion

blarghmatey:
In order to know that the lack of a failure message doesn't just mean
that the backup didn't get executed then we would like to be notified
on successful completion of the orchestrate routines as well. This PR
adds a step to the backup and restore scripts to notify the devops
channel on completion whether the result is success or failure.


shaidar:
:thumbsup:

PR #262, opened 2017-04-03
Added state to handle generic backends for business units

blarghmatey:
In order to properly segregate our generic secrets I added a state to enumerate the business units that we are dealing with and mount generic secret backends for each one.


shaidar:
:thumbsup:

PR #263, opened 2017-04-04
Removing datadog integration files

shaidar:
The files were experimental to start with and are not really needed given that Datadog integrations are handled through pillar data.


blarghmatey:
👍

PR #265, opened 2017-04-05
Updated the states for the secret backends to use new params

blarghmatey:
The vault state extension was updated to expose the `ttl_max` and
`ttl_default` parameters for tuning the mount points. Updated
orchestrate states to take advantage of that fact.


shaidar:
:thumbsup:

PR #266, opened 2017-04-05
Minor bug fixes

MDPolaris:
RHEL/CentOS python packages are changing names and it is being rolled out in phases. The pip package has been changed to python2-pip. 

libgit now requires python-cffi, which is still on the old naming scheme.

Salt has replaced tar_options with options. tar_options is deprecated now.

The pip install command was missing the version number which could lead to a failure during installation if the libgit version being built is not the most recent.

### DIFF
--- a/salt/master_utils/libgit.sls
+++ b/salt/master_utils/libgit.sls
@@ -22,7 +22,8 @@
             'make',
             'openssl-devel',
             'libssh-devel',
-            'python-pip',
+            'python2-pip',
+            'python-cffi',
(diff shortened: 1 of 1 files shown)


pdpinch:
@mitodl/devops what should we do with this PR?

MDPolaris:
From my perspective, I realized there was another bug with my changes. I
had hoped to get back to this sooner, but you can reject the PR. I'll
resubmit when i resolve that issue.

PR #269, opened 2017-04-10
Refactor edx map template

shaidar:
- Added business unit and purpose to tags
- Added business unit and purposes as grains
- Made sure to match role with proper purpose when it comes to live and draft instances. For example, if role is edx-live, then purpose will have a 'live' suffix.

### DIFF
--- a/salt/orchestrate/aws/map_templates/edx.yml
+++ b/salt/orchestrate/aws/map_templates/edx.yml
@@ -11,16 +11,12 @@ edx:
       tag:
         business_unit: {{ business_unit }}
         environment: {{ environment_name }}
-        purpose: {{ purpose }}
+        purpose_prefix: {{ purpose_prefix }}
       grains:
         business_unit: {{ business_unit }}
         environment: {{ environment_name }}
-        {% if app_type == 'live' or 'draft' %}
         {% set modified_purpose = purpose + '-' + app_type %}
         purpose: {{ modified_purpose }}
-        {% else %}
-        purpose: {{ purpose }}
-        {% endif %}
         roles:
           - edx
           - edx-{{ app_type }}
--- a/salt/orchestrate/aws/map_templates/edx.yml
+++ b/salt/orchestrate/aws/map_templates/edx.yml
@@ -16,9 +16,10 @@ edx:
         business_unit: {{ business_unit }}
         environment: {{ environment_name }}
         {% if app_type not in purpose %}
-        {% set purpose = purpose[:-len(purpose.split('-')[-1])] +app_type %}
-        purpose:
-          - {{ purpose }}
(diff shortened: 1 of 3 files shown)


salt/orchestrate/aws/map_templates/edx.yml
```diff
@@ -9,9 +9,17 @@ edx:
...
           SecurityGroupId: {{ securitygroupid }}
       tag:
+        business_unit: {{ business_unit }}
         environment: {{ environment_name }}
+        purpose: {{ purpose }}
       grains:
+        business_unit: {{ business_unit }}
         environment: {{ environment_name }}
+        {% if app_type not in purpose %}
+        {% set purpose = purpose[:-len(purpose.split('-')[-1])] +app_type %}
```
blarghmatey: It's potentially dangerous to overwrite the variable used in the conditional, particularly if it is being used in a loop. This should probably be set to a different name for reference inside the `if` block.

blarghmatey: The `if` block has the effect that, if the `app_type` is included in the `purpose` then the purpose grain will never be set.

salt/orchestrate/aws/map_templates/edx.yml
```diff
@@ -9,9 +9,17 @@ edx:
...
+        purpose:
+          - {{ purpose }}
```
blarghmatey: This should be a string rather than a list. I don't think we have a use case for needing more than one purpose on a given host.

salt/orchestrate/edx/init.sls
```diff
@@ -13,7 +13,9 @@ generate_edx_cloud_map_file:
     - template: jinja
     - makedirs: True
     - context:
+        business_unit: {{ BUSINESS_UNIT }}
         environment_name: {{ ENVIRONMENT }}
+        purpose: {{ PURPOSE }}
```
blarghmatey: I would propagate the `_prefix` nomenclature throughout so that it's easier to read and understand what it is intended to represent.

salt/orchestrate/aws/map_templates/edx.yml
```diff
@@ -9,9 +9,14 @@ edx:
           SubnetId: {{ subnetids[id_num % num_subnets] }}
           SecurityGroupId: {{ securitygroupid }}
       tag:
+        business_unit: {{ business_unit }}
         environment: {{ environment_name }}
+        purpose_prefix: {{ purpose_prefix }}
       grains:
+        business_unit: {{ business_unit }}
         environment: {{ environment_name }}
+        {% set modified_purpose = purpose + '-' + app_type %}
```
blarghmatey: This needs to reference `purpose_prefix`. As it is right now it will throw an undefined variable exception because `purpose_prefix` is the variable getting set via the `context` in the orchestrate state.

blarghmatey:
👍

PR #270, opened 2017-04-10
Added state to handle refreshing vault token on master

blarghmatey:
In order to be able to use Vault the master needs to have a valid
token. This state will execute a token refresh and then restart the
salt processes to reread the config.


shaidar:
:thumbsup:

PR #271, opened 2017-04-11
Delete old edx log files

shaidar:
- gzipped files older than 30 days should be deleted on a weekly schedule given the limited storage restrictions we have on the edx instances
- updated top file


shaidar:
Tobias said this is :thumbsup: and to go ahead and merge.

PR #272, opened 2017-04-12
Splitting out edx_worker into its own instance

shaidar:
- Removed edx_worker role from ansible run
- Fixed edx_worker cloud profile
- Created edx_worker map_template
- Fixed typo in ike_edx map template
- Created generic mitx orchestrate that can be used for both qa and rp dependent on VPC_RESOURCE_SUFFIX


salt/orchestrate/aws/map_templates/ike_edx.yml
```diff
@@ -1,4 +1,4 @@
-edx:
+ike_edx:
```
blarghmatey: This needs to stay the same. The identifier here is used to determine which cloud profile will be deployed.

salt/orchestrate/aws/mitx.sls
```diff
@@ -0,0 +1,369 @@
...
+{% set cidr_block_public_subnet_3 = 10.5.3.0/24 %}
+{% set cidr_ip = 10.5.0.0/22 %}
+{% set cidr_block = 10.5.0.0/16 %}
+{% elif VPC_RESOURCE_SUFFIX == 'mitx-rp' %}
+{% set cidr_block_public_subnet_1 = 10.6.1.0/24 %}
+{% set cidr_block_public_subnet_2 = 10.6.2.0/24 %}
+{% set cidr_block_public_subnet_3 = 10.6.3.0/24 %}
+{% set cidr_ip = 10.6.0.0/22 %}
+{% set cidr_block = 10.5.0.0/16 %}
+{% endif %}
```
blarghmatey: Rather than using this mapping you can use the network prefix values stored in the `environments` pillar data.

salt/orchestrate/aws/map_templates/edx_worker.yml
```diff
@@ -0,0 +1,24 @@
+edx_worker:
+  {% set num_subnets = subnetids|length %}
```
blarghmatey: We should just include the worker nodes in the same map template as the edX instances so that creation of all of them can be executed in parallel.

salt/orchestrate/aws/map_templates/edx.yml
```diff
@@ -20,5 +20,23 @@ edx:
         roles:
           - edx
           - edx-{{ app_type }}
+
+  - edx-worker-{{ environment_name }}-{{ app_type }}-{{ id_num }}:
```
blarghmatey: This needs to be outside of the loop and scoped to the `edx_worker` namespace so that it will use the proper cloud profile.

salt/orchestrate/edx/init.sls
```diff
@@ -72,24 +71,34 @@ populate_mine_with_edx_node_data:
...
   salt.state:
-    - tgt: 'G@roles:edx and G@environment:{{ ENVIRONMENT }}'
+    - tgt: 'P@roles:(edx|edx-worker) and G@environment:{{ ENVIRONMENT }}'
     - tgt_type: compound
     - highstate: True
     - require:
         - salt: deploy_consul_agent_to_edx_nodes
+
+stop_non_edx_worker_services:
+{% for service in ['edxapp:', 'forum', 'xqueue', 'xqueue_consumer'] %}
```
blarghmatey: The for loop needs to be above the state ID and the ID needs to be templatized so that it doesn't violate the uniqueness constraint.

salt/orchestrate/edx/init.sls
```diff
@@ -72,24 +71,34 @@ populate_mine_with_edx_node_data:
...
+  salt.function:
+    - name: supervisord.dead
+    - tgt: 'G@roles:edx-worker and G@environment:{{ ENVIRONMENT }}'
+    - tgt_type: compound
+    - bin_env: '/edx/bin/supervisorctl'
+    - name: {{ service }}
```
blarghmatey: `name` and `bin_env` need to be nested under `kwargs` as dictionary entries (no `- `)

salt/orchestrate/aws/map_templates/edx.yml
```diff
@@ -20,5 +20,23 @@ edx:
...
+      network_interfaces:
+        - DeviceIndex: 0
+          AssociatePublicIpAddress: False
```
blarghmatey: This needs to be `True` or else the instance won't be able to access the internet to download dependencies.

salt/orchestrate/edx/init.sls
```diff
@@ -16,13 +16,12 @@ generate_edx_cloud_map_file:
...
         environment_name: {{ ENVIRONMENT }}
         purpose_prefix: {{ PURPOSE_PREFIX }}
-        roles:
-          - edx
-          - log-forwarder
         securitygroupid:
           - {{ salt.boto_secgroup.get_group_id(
               'edx-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
           - {{ salt.boto_secgroup.get_group_id(
+              'edx-worker{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
```
blarghmatey: This will be applied to all of the nodes in the map template. You might want to break out the different groups into individual ids in the context object and then compile the list of groups via the named references in the map template.

salt/orchestrate/aws/map_templates/edx.yml
```diff
@@ -20,5 +20,23 @@ edx:
...
+        business_unit: {{ business_unit }}
+        environment: {{ environment_name }}
+        purpose_prefix: {{ purpose_prefix }}
+      grains:
+        business_unit: {{ business_unit }}
+        environment: {{ environment_name }}
+        {% set modified_purpose = purpose_prefix + '-' + app_type %}
+        purpose: {{ modified_purpose }}
+        roles:
+          - edx-worker-{{ app_type }}
```
blarghmatey: This should also have a generic `edx-worker` role

salt/orchestrate/aws/map_templates/edx.yml
shaidar: I'm under the impression that public IP is mainly for inbound traffic and outbound traffic for an instance without a public IP will go through the gateway. I was looking at some of my old terraform stuff and instances like rabbitmq weren't assigned a public IP since they only were accessed from the edxapp nodes. Maybe I'm missing something though and need to test it out in qa.

shaidar: We'd need a NAT gateway if we were to set this to false in order for that instance to go beyond the VPC.

salt/orchestrate/aws/map_templates/edx.yml
shaidar: Why not use the same cloud profile? Think the only difference I noticed was that the EBS volume size was a tiny bit bigger for the edxapp. 
Also, if we use a different cloud profile, wouldn't we need to load it in edx/init.sls where we're already loading the edx.conf one?

salt/orchestrate/aws/mitx.sls
```diff
@@ -6,7 +6,7 @@
 {% from "orchestrate/aws_env_macro.jinja" import VPC_NAME, VPC_RESOURCE_SUFFIX,
  ENVIRONMENT, subnet_ids with context %}
 
-{% import_yaml "environment_settings.yml" as env_settings %}
+{% set env_settings = salt.pillar.get('environment_settings') %}
```
blarghmatey: This should be `salt.pillar.get('environments:{}'.format(ENVIRONMENT))`

salt/orchestrate/edx/init.sls
```diff
@@ -93,12 +102,19 @@ build_edx_nodes:
...
-    - bin_env: '/edx/bin/supervisorctl'
-    - name: {{ service }}
+    - kwargs:
+        bin_env: '/edx/bin/supervisorctl'
+        name: {{ service }}
+{% endfor %}
+
+{% for bucket in bucket_prefixes %}
+create_edx_s3_bucket_{{ bucket }}_{{ PURPOSE_PREFIX }}:
+  boto_s3_bucket.create:
```
blarghmatey: This should use the `.present` state function instead of the `create` function from the execution module.

salt/orchestrate/edx/init.sls
```diff
@@ -72,24 +80,41 @@ populate_mine_with_edx_node_data:
...
+    - name: {{ bucket }}_{{ PURPOSE_PREFIX }}
+{% endfor %}
```
blarghmatey: This block should go toward the top of the file or add a `require` setting so that the buckets are present by the time the edX instances are launched and configured.

salt/orchestrate/aws/map_templates/edx.yml
```diff
@@ -7,7 +7,9 @@ edx:
         - DeviceIndex: 0
           AssociatePublicIpAddress: True
           SubnetId: {{ subnetids[id_num % num_subnets] }}
-          SecurityGroupId: {{ securitygroupid }}
+          SecurityGroupId: |
+            {{ securitygroupids.edxapp, securitygroupids.edx-worker, securitygroupids.default,
```
blarghmatey: This should be formatted as a list, and I don't believe that calling the multiple values in a single Jinja reference will work...

salt/orchestrate/aws/map_templates/edx.yml
```diff
@@ -7,9 +7,12 @@ edx:
         - DeviceIndex: 0
           AssociatePublicIpAddress: True
           SubnetId: {{ subnetids[id_num % num_subnets] }}
-          SecurityGroupId: |
-            {{ securitygroupids.edxapp, securitygroupids.edx-worker, securitygroupids.default,
-            securitygroupids.salt-master, securitygroupids.consul-agent }}
+          SecurityGroupId:
+            - {{ securitygroupids.edxapp }}
+            - {{ securitygroupids.edx-worker }}
```
blarghmatey: This isn't necessary on the edxapp instance since it's exposing a port that's already present in the edxapp security group.

salt/orchestrate/aws/map_templates/edx.yml
```diff
@@ -22,3 +27,31 @@ edx:
           - edx-{{ app_type }}
 {% endfor %}
 {% endfor %}
+
+edx-worker:
+  {% set num_subnets = subnetids|length %}
+  {% for app_type, num_instances in app_types.items() %}
+  {% for id_num in range(num_instances) %}
```
blarghmatey: This needs to reference the edx-worker. e.g. `for id_num in range(num_instances.edx-worker)`, same for edxapp above.

salt/orchestrate/edx/init.sls
```diff
@@ -1,6 +1,11 @@
 {% from "orchestrate/aws_env_macro.jinja" import VPC_NAME, VPC_RESOURCE_SUFFIX,
  ENVIRONMENT, BUSINESS_UNIT, PURPOSE_PREFIX, subnet_ids with context %}
 
+ {% set env_settings = salt.pillar.get('environment_settings') %}
+ {% set bucket_prefixes = |
```
blarghmatey: The pipe character is only for multiline YAML strings. You can remove it here and bring the next line up to the same line.

salt/orchestrate/edx/init.sls
```diff
@@ -1,6 +1,11 @@
...
+     env_settings.edxapp_secret_backends.aws.bucket_prefixes %}
+{% set purposes = env_settings.[ENVIRONMENT].purposes %}
```
blarghmatey: Remove the `.` before the square brackets.

salt/orchestrate/aws/cloud_profiles/edx-worker.conf
```diff
@@ -14,7 +14,7 @@ edx_forum:
     role: edx
   grains:
     roles:
-      - edx-worker
+      - edx
```
blarghmatey: This should default to a role of `edx-worker`.

blarghmatey:
👍

PR #273, opened 2017-04-14
Create multiple mysql databases based on purpose

shaidar:
As part of the ficus rollout and some re-architecting, this creates multiple mysql databases on the same host based on purpose.


salt/edx/edx_mysql.sls
```diff
@@ -0,0 +1,20 @@
...
+{% set mysql_user = salt.pillar.get('edx:ansible_vars:EDXAPP_MYSQL_USER') %}
+{% set mysql_password = salt.pillar.get('edx:ansible_vars:EDXAPP_MYSQL_PASSWORD') %}
+
+{% for db,name in edxlocal_databases.iteritems() %}
+{% for purpose in purposes %}
+edxapp_create_db_{{ name }}-{{ purpose }}:
+  mysql_database.present:
+    - name: {{ name }}-{{ purpose }}
+    - connection_user: {{ mysql_user }}
+    - connection_pass: {{ mysql_password }}
```
blarghmatey: This will need to have the host and port parameters set as well. Additionally, the credentials referenced here won't have privileges to create a database. For that we will want to leverage the `mysql-{env}/creds/admin` Vault role.

salt/edx/edx_mysql.sls
```diff
@@ -0,0 +1,20 @@
+#!jinja|yaml
+{% from "shared/edx/mitx.jinja" import edx with context %}
```
blarghmatey: This won't render properly because that jinja file is in the pillar repo.

salt/edx/edx_mysql.sls
```diff
@@ -1,20 +1,29 @@
...
   'XQUEUE_MYSQL_DB_NAME': 'xqueue',
   'EDXAPP_MYSQL_DB_NAME': 'edxapp',
   }
 %}
-{% set mysql_user = salt.pillar.get('edx:ansible_vars:EDXAPP_MYSQL_USER') %}
-{% set mysql_password = salt.pillar.get('edx:ansible_vars:EDXAPP_MYSQL_PASSWORD') %}
+{% set edxapp_mysql_creds = salt.vault.read(
+    'mysql-{env}/creds/edxapp-{purpose}'.format(
+        env=environment,
+        purpose=purpose)) %}
```
blarghmatey: This needs to be inside the for loop because the purpose variable hasn't been set at this point.

salt/edx/edx_mysql.sls
```diff
@@ -1,20 +1,29 @@
...
+{% set edxapp_mysql_host = salt.pillar.get('edx:ansible_vars:EDXAPP_MYSQL_HOST') %}
+{% set edxapp_mysql_port = salt.pillar.get('edx:ansible_vars:EDXAPP_MYSQL_PORT') %}
```
blarghmatey: If this is going to be run on the master then this pillar data won't be available. So, we will either need to use the consul dns for the host or run this on an edX node.

salt/edx/edx_mysql.sls
```diff
@@ -0,0 +1,30 @@
...
+
+{% for db,name in edxlocal_databases.iteritems() %}
+{% for purpose in purposes %}
+{% set edxapp_mysql_creds = salt.vault.read(
+    'mysql-{env}/creds/edxapp-{purpose}'.format(
+        env=environment,
+        purpose=purpose)) %}
+edxapp_create_db_{{ name }}-{{ purpose }}:
+  mysql_database.present:
+    - name: {{ name }}-{{ purpose }}
```
blarghmatey: The database names should use `_` instead of `-`

salt/edx/edx_mysql.sls
```diff
@@ -0,0 +1,30 @@
...
+
+{% for db,name in edxlocal_databases.iteritems() %}
+{% for purpose in purposes %}
+{% set edxapp_mysql_creds = salt.vault.read(
+    'mysql-{env}/creds/edxapp-{purpose}'.format(
+        env=environment,
+        purpose=purpose)) %}
+edxapp_create_db_{{ name }}_{{ purpose }}:
+  mysql_database.present:
+    - name: {{ name }}_{{ purpose }}
```
blarghmatey: You will need to do a replace in the purpose string as well. `|replace('-', '_')`

PR #274, opened 2017-04-14
Updated elb orchestrate to be more generic and flexible

blarghmatey:


salt/orchestrate/aws/mitx_qa_elb.sls
```diff
@@ -1,24 +1,13 @@
...
-             'mitx-qa.mitx.mit.edu.',
-             'preview-mitx-qa.mitx.mit.edu.',
-             'prod-gr-qa.mitx.mit.edu.']
-} %}
-{% set environment = 'mitx-qa' %}
+{% from "orchestrate/aws_env_macro.jinja" import VPC_NAME, VPC_RESOURCE_SUFFIX,
+ ENVIRONMENT, PURPOSE_PREFIX, subnet_ids with context %}
+{% set env_settings = salt.pillar.get('environments:{}'.format(ENVIRONMENT)) %}
+
 {% set security_groups = salt.pillar.get('edx:lb_security_groups', ['default', 'edx-mitx-qa']) %}
```
shaidar: I know this line didn't change, but are we sure about lb_security_groups, cause I don't see it in pillar data

salt/orchestrate/aws/mitx_qa_elb.sls
```diff
@@ -43,8 +32,10 @@ create_elb_for_edx_{{ edx_type }}:
           enabled: True
           timeout: 300
     - cnames:
-        {% for domain in domains[edx_type] %}
-        - name: {{ domain }}
+        {% for domain_key, domain in purpose.domains.items()  %}
+        {% if (app_type == 'live' and domain_key in ['lms', 'gitreload'])
+           or app_type == 'draft' %}
+        - name: {{ domain }}.
```
shaidar: Is the dot necessary here?

blarghmatey: Yes, because that is how the domain is represented in the zone file. If the `.` isn't present then it won't work properly.

salt/orchestrate/aws/mitx_qa_elb.sls
blarghmatey: If the value is not present in pillar then it will use the default values passed as the second parameter of the `get` call.

salt/orchestrate/aws/mitx_qa_elb.sls
```diff
@@ -1,24 +1,13 @@
...
+
 {% set security_groups = salt.pillar.get('edx:lb_security_groups', ['default', 'edx-mitx-qa']) %}
-{% set subnet_ids = [] %}
-{% for subnet in salt.boto_vpc.describe_subnets(subnet_names=[
-    'public1-mitx-qa', 'public2-mitx-qa', 'public3-mitx-qa'])['subnets'] %}
-{% do subnet_ids.append('{0}'.format(subnet['id'])) %}
-{% endfor %}
 
 {% for edx_type in ['draft', 'live'] %}
 {% set elb_name = 'edx-{0}-mitx-qa'.format(edx_type) %}
```
shaidar: Should the elb_name not be similar to the instance names - edx-{env}-edx_type?

blarghmatey: Yeah, not a bad idea. Not terribly critical, but still a good thing to have.

shaidar:
:thumbsup:

salt/orchestrate/aws/mitx_elb.sls
```diff
@@ -43,8 +33,10 @@ create_elb_for_edx_{{ edx_type }}:
...
           timeout: 300
     - cnames:
-        {% for domain in domains[edx_type] %}
-        - name: {{ domain }}
+        {% for domain_key, domain in purpose.domains.items()  %}
+        {% if (app_type == 'live' and domain_key in ['lms', 'gitreload'])
+           or app_type == 'draft' %}
+        - name: {{ domain }}.
           zone: mitx.mit.edu.
           ttl: 60
```
shaidar: Missing {% endif %}

PR #277, opened 2017-04-18
Add ami id's to support different Ubuntu 

shaidar:


salt/orchestrate/aws/map_templates/edx.yml
```diff
@@ -12,6 +12,7 @@ edx:
             - {{ securitygroupids.default }}
             - {{ securitygroupids.salt-master }}
             - {{ securitygroupids.consul-agent }}
+      image: {{ ami_id }}
```
blarghmatey: This needs to be added to the worker definition as well.

blarghmatey:
👍

PR #281, opened 2017-04-25
Refactor backup/restore and slack secret from vault

shaidar:
Following changes were made:
- Changed duplicity tempdir to /backups/tmp
- Added slack api_token and webhook_url to vault and changed slack calls to read values from vault
- Added the mysql database name to cover the edxapp and xqueue db
- Pointed archive dir to cache dir
- Created 400GB EBS volume and pointed backup host to mount that and use as persistent storage for duplicity cache. Had to specify host's availability zone since EBS volume is not multi-zone.

### DIFF
--- a/salt/backups/templates/backup_mysql.sh
+++ b/salt/backups/templates/backup_mysql.sh
@@ -18,7 +18,6 @@ PASSPHRASE={{ settings.duplicity_passphrase }} /usr/bin/duplicity \
           --archive-dir {{ cachedir }} \
           --full-if-older-than 1W \
           --allow-source-mismatch --tempdir /backups/tmp/ \
-          s3+http://odl-operations-backups/{{ settings.get('directory', 'mysql') }}/ \
-          {{ settings.database }}/
+          s3+http://odl-operations-backups/{{ settings.get('directory', 'mysql') }}
 
 rm -rf {{ backupdir }}
--- a/salt/backups/templates/backup_mysql.sh
+++ b/salt/backups/templates/backup_mysql.sh
@@ -17,7 +17,8 @@ PASSPHRASE={{ settings.duplicity_passphrase }} /usr/bin/duplicity \
           --s3-use-server-side-encryption {{ backupdir }} \
           --archive-dir {{ cachedir }} \
           --full-if-older-than 1W \
-          --allow-source-mismatch --tempdir /backups \
-          s3+http://odl-operations-backups/{{ settings.get('directory', 'mysql') }}/
+          --allow-source-mismatch --tempdir /backups/tmp/ \
+          s3+http://odl-operations-backups/{{ settings.get('directory', 'mysql') }}/ \
+          {{ settings.database }}/
 
 rm -rf {{ backupdir }}
--- a/salt/backups/templates/restore_mongodb.sh
+++ b/salt/backups/templates/restore_mongodb.sh
@@ -2,12 +2,13 @@
 set -e
 
 {% set backupdir = '/backups/{}'.format(settings.get('directory', 'mongodb')) %}
+{% set cachedir = '/backups/.cache/{}'.format(settings.get('directory', 'mongodb')) %}
 mkdir -p {{ backupdir }}
 
 PASSPHRASE={{ settings.duplicity_passphrase }} /usr/bin/duplicity restore\
           --s3-use-server-side-encryption s3+http://odl-operations-backups/{{ settings.get('directory', 'mongodb') }}/ \
-          --archive-dir /backups/.cache \
-          --force --tempdir /backups {{ backupdir }}/
+          --archive-dir {{ cachedir }}
+          --force --tempdir /backups/tmp/ {{ backupdir }}
 
 /usr/bin/mongorestore --host {{ settings.host }} \
                       --port {{ settings.get('port', 27017) }} \
--- a/salt/backups/templates/restore_mysql.sh
+++ b/salt/backups/templates/restore_mysql.sh
@@ -2,12 +2,13 @@
 set -e
 
 {% set backupdir = '/backups/{}'.format(settings.get('directory', 'mysql')) %}
+{% set cachedir = '/backups/.cache/{}'.format(settings.get('directory', 'mysql')) %}
 mkdir -p {{ backupdir }}
 
 PASSPHRASE={{ settings.duplicity_passphrase }} /usr/bin/duplicity restore \
           --s3-use-server-side-encryption s3+http://odl-operations-backups/{{ settings.get('directory', 'mysql') }}/ \
-          --archive-dir /backups/.cache \
-          --force --tempdir /backups  {{ backupdir }}/
+          {{ settings.database }}/ --archive-dir {{ cachedir }} \
+          --force --tempdir /backups/tmp/ {{ backupdir }}/{{ settings.database }}
 
 /usr/bin/mysql --host {{ settings.host }} \
                --port {{ settings.get('port', 3306) }} \
--- a/salt/backups/templates/restore_mysql.sh
+++ b/salt/backups/templates/restore_mysql.sh
@@ -6,8 +6,8 @@ set -e
 mkdir -p {{ backupdir }}
 
 PASSPHRASE={{ settings.duplicity_passphrase }} /usr/bin/duplicity restore \
-          --s3-use-server-side-encryption s3+http://odl-operations-backups/{{ settings.get('directory', 'mysql') }}/ \
-          {{ settings.database }}/ --archive-dir {{ cachedir }} \
+          --s3-use-server-side-encryption s3+http://odl-operations-backups/{{ settings.get('directory', 'mysql') }} \
+          --archive-dir {{ cachedir }} \
           --force --tempdir /backups/tmp/ {{ backupdir }}/{{ settings.database }}
 
 /usr/bin/mysql --host {{ settings.host }} \
--- a/salt/orchestrate/aws/cloud_profiles/backup_host.conf
+++ b/salt/orchestrate/aws/cloud_profiles/backup_host.conf
@@ -2,6 +2,7 @@ backup_host:
   provider: mitx
   size: t2.medium
   image: ami-49e5cb5e
+  availability_zone: us-east-1b
   ssh_username: admin
   ssh_interface: private_ips
   script_args: -Z -A salt.private.odl.mit.edu
@@ -15,9 +16,8 @@ backup_host:
     - DeviceName: /dev/xvda
       Ebs.VolumeSize: 8
       Ebs.VolumeType: gp2
-    - DeviceName: /dev/xvdb
-      Ebs.VolumeSize: 400
-      Ebs.VolumeType: gp2
+    - device: /dev/xvdb
+      volume_id: vol-0bd5c579e0336bcef
   minion:
     master:
       - salt.private.odl.mit.edu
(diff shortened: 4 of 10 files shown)


salt/orchestrate/aws/cloud_profiles/backup_host.conf
```diff
@@ -15,9 +16,8 @@ backup_host:
     - DeviceName: /dev/xvda
       Ebs.VolumeSize: 8
       Ebs.VolumeType: gp2
-    - DeviceName: /dev/xvdb
-      Ebs.VolumeSize: 400
-      Ebs.VolumeType: gp2
+    - device: /dev/xvdb
+      volume_id: vol-0bd5c579e0336bcef
```
blarghmatey: Rather than hardcoding the volume ID it would be better to use a labelled volume that we mount and unmount as part of the orchestrate state. If the volume is launched as part of the profile I would be worried about it being terminated along with the instance...

salt/backups/templates/restore_mysql.sh
```diff
@@ -2,12 +2,13 @@
...
 
 {% set backupdir = '/backups/{}'.format(settings.get('directory', 'mysql')) %}
+{% set cachedir = '/backups/.cache/{}'.format(settings.get('directory', 'mysql')) %}
 mkdir -p {{ backupdir }}
 
 PASSPHRASE={{ settings.duplicity_passphrase }} /usr/bin/duplicity restore \
           --s3-use-server-side-encryption s3+http://odl-operations-backups/{{ settings.get('directory', 'mysql') }}/ \
-          --archive-dir /backups/.cache \
-          --force --tempdir /backups  {{ backupdir }}/
+          {{ settings.database }}/ --archive-dir {{ cachedir }} \
```
blarghmatey: What is the purpose of the `settings.database` reference on this line?

salt/backups/templates/restore_mongodb.sh
```diff
@@ -2,12 +2,13 @@
...
 {% set backupdir = '/backups/{}'.format(settings.get('directory', 'mongodb')) %}
+{% set cachedir = '/backups/.cache/{}'.format(settings.get('directory', 'mongodb')) %}
 mkdir -p {{ backupdir }}
 
 PASSPHRASE={{ settings.duplicity_passphrase }} /usr/bin/duplicity restore\
           --s3-use-server-side-encryption s3+http://odl-operations-backups/{{ settings.get('directory', 'mongodb') }}/ \
-          --archive-dir /backups/.cache \
-          --force --tempdir /backups {{ backupdir }}/
+          --archive-dir {{ cachedir }}
+          --force --tempdir /backups/tmp/ {{ backupdir }}
```
blarghmatey: If we are going to use a subdirectory of /backups/ then we should add a call to `file.directory` as part of the orchestrate to ensure that the directory exists before we try using it.

salt/backups/templates/backup_mysql.sh
```diff
@@ -17,7 +17,8 @@ PASSPHRASE={{ settings.duplicity_passphrase }} /usr/bin/duplicity \
           --s3-use-server-side-encryption {{ backupdir }} \
           --archive-dir {{ cachedir }} \
           --full-if-older-than 1W \
-          --allow-source-mismatch --tempdir /backups \
-          s3+http://odl-operations-backups/{{ settings.get('directory', 'mysql') }}/
+          --allow-source-mismatch --tempdir /backups/tmp/ \
+          s3+http://odl-operations-backups/{{ settings.get('directory', 'mysql') }}/ \
+          {{ settings.database }}/
```
blarghmatey: What is the intent for this reference to `settings.database`?

salt/backups/templates/restore_mongodb.sh
shaidar: Added file.directory to create the /backups dir after the backup drive is mounted

salt/backups/templates/restore_mysql.sh
shaidar: My thinking is to map each database (edxapp, xqueue) separately in case we want to backup/restore one or the other separately.

salt/backups/templates/backup_mysql.sh
shaidar: My thinking is to map each database (edxapp, xqueue) separately in case we want to backup/restore one or the other separately.

salt/orchestrate/edx/backup.sls
```diff
@@ -62,6 +63,24 @@ deploy_backup_instance_to_{{ ENVIRONMENT }}:
...
+    - name: saltutil.runner
+    - arg:
+        - cloud.action
+    - kwarg:
+        func: ec2.attach_volume
+        kwargs:
+          instance_id: {{ instance_id }}
+          volume_name: {{ backup_volume_name }}
+          zone: us-east-1b
+          size: 400
```
blarghmatey: This is also not creating the volume... I think that what we want as an action instead is to use `ec2.create_attach_volume` which will allow us to specify the device name so that we don't have to attempt to find it via grains (which I don't believe will resolve properly anyway...).

salt/orchestrate/edx/backup.sls
```diff
@@ -71,7 +90,7 @@ mount_backup_drive:
         - mount.mounted
     - kwarg:
         name: /backups
-        device: /dev/xvdb
+        device: /dev/{{ salt.grains.get('ec2:block_device_mapping:ebs2') }}
```
blarghmatey: If we use `ec2.create_attach_volume` then we can just use the same device name as what is specified in the call to that action.

salt/orchestrate/edx/backup.sls
```diff
@@ -88,6 +107,33 @@ execute_enabled_backup_scripts:
...
+unmount_backup_drive:
+  salt.function:
+    - tgt: 'G@roles:backups and G@environment:{{ ENVIRONMENT }}'
+    - tgt_type: compound
+    - name: state.single
+    - arg:
+        - mount.unmounted
+    - kwarg:
+        name: /backups
+        device: /dev/{{ salt.grains.get('ec2:block_device_mapping:ebs2') }}
```
blarghmatey: This needs a require statement to tie it to the success state of the backup scripts.

salt/backups/templates/backup_mysql.sh
blarghmatey: In that case you will want to move this up a line and remove the space after the `/` following the `settings.get` call.

salt/orchestrate/edx/backup.sls
```diff
@@ -61,18 +63,29 @@ deploy_backup_instance_to_{{ ENVIRONMENT }}:
...
+    - name: saltutil.runner
     - arg:
-        - blockdev.formatted
+        - cloud.action
     - kwarg:
-        name: /dev/xvdb
-        fs_type: ext4
+        func: ec2.create_attach_volume
+        kwargs:
+          instance_id: {{ instance_id }}
```
blarghmatey: Looking again, we don't want to pass `instance_id`, instead we want to pass the instance _name_ (e.g. backups-mitx-rp) because the instance_id will be looked up based on that name if it's not explicitly provided. Since we are creating the instance as part of this orchestrate state we won'd have any way to resolve the instance_id ahead of time because it doesn't yet exist when the state file is being rendered.

salt/orchestrate/aws/cloud_profiles/backup_host.conf
```diff
@@ -2,6 +2,7 @@ backup_host:
   provider: mitx
   size: t2.medium
   image: ami-49e5cb5e
+  availability_zone: us-east-1b
```
blarghmatey: I don't think we need to specify the AZ because that would be implicit from the subnet that it gets placed into.

salt/orchestrate/edx/backup.sls
```diff
@@ -7,6 +7,8 @@
     'public1-{}'.format(VPC_RESOURCE_SUFFIX), 'public2-{}'.format(VPC_RESOURCE_SUFFIX), 'public3-{}'.format(VPC_RESOURCE_SUFFIX)])['subnets'] %}
 {% do subnet_ids.append('{0}'.format(subnet['id'])) %}
 {% endfor %}
+{% set slack_api_token = salt.vault.read('secret-operations/global/slack/slack_api_token.data.value') %}
```
blarghmatey: `.data.value` needs to be outside the parentheses for the `.read` call.

salt/orchestrate/operations/backups.sls
```diff
@@ -6,6 +6,7 @@
     'public1-{}'.format(VPC_RESOURCE_SUFFIX), 'public2-{}'.format(VPC_RESOURCE_SUFFIX), 'public3-{}'.format(VPC_RESOURCE_SUFFIX)])['subnets'] %}
 {% do subnet_ids.append('{0}'.format(subnet['id'])) %}
 {% endfor %}
+{% set slack_api_token = salt.vault.read('secret-operations/global/slack/slack_api_token.data.value') %}
```
blarghmatey: `.data.value` needs to be oustside the parentheses for the `.read` function call.

salt/orchestrate/edx/backup.sls
```diff
@@ -102,6 +124,35 @@ execute_enabled_backup_scripts:
...
+    - tgt: 'G@roles:backups and G@environment:{{ ENVIRONMENT }}'
+    - tgt_type: compound
+    - name: saltutil.runner
+    - arg:
+        - cloud.action
+    - kwarg:
+        func: boto_ec2.detach_volume
+        kwargs:
+          tags: {{ backup_volume_name }}
+          instance_id: {{ instance_id }}
```
blarghmatey: We want to be able to pass the instance _name_ here too instead of needing the ID.

salt/orchestrate/aws/cloud_profiles/backup_host.conf
shaidar: I think I had put it there because the EBS volume that we're attaching has to be in the same AZ as the host.

salt/orchestrate/edx/backup.sls
shaidar: instance_id is set based on value returned from `boto_ec2.find_instances('name=backup-..)`
[Line#69](https://github.com/mitodl/salt-ops/blob/e1d6d884fcb7476d870814c4c20999a863459692/salt/orchestrate/edx/backup.sls#L69)

blarghmatey: Right, but because we are creating the instance as part of this run the `instance_id` will not be resolved because the`find_instances` call will be executed before the instance is created.

salt/orchestrate/aws/cloud_profiles/backup_host.conf
blarghmatey: Makes sense.

salt/backups/templates/backup_mysql.sh
blarghmatey: This line still needs to be addressed... I don't think that it is needed at all since each database is dumped to its own file already. As it stands right now, the syntax is invalid.

salt/backups/templates/restore_mysql.sh
blarghmatey: This line needs to be addressed as well.

salt/orchestrate/edx/backup.sls
```diff
@@ -87,8 +112,17 @@ mount_backup_drive:
...
+
+create_backup_directory:
+  salt.function:
+    - tgt: 'G@roles:backups and G@environment:{{ ENVIRONMENT }}'
+    - tgt_type: compound
+    - name: state.single
+    - arg:
+        - file.directory
+    - kwargs:
+        name: backups
```
blarghmatey: This should be `/backups/tmp` so that the `tmp` subdirectory is created. Also, without the leading `/` the backups directory would be wherever the process happened to be running from. Granted, that is most likely to be the `/` directory, but no guarantees.

blarghmatey:
:+1:

PR #283, opened 2017-05-01
Fix for master render errors

shaidar:
Noticed errors in master log about "Jinja variable 'dict object' has no attribute 'data'" on draft_disk _cleanup.sls file. This should fix the problem.


blarghmatey:
👍

PR #288, opened 2017-05-08
Changed DeviceName

shaidar:
Instance provisioning is adding another block device instead of sizing the root volume which is what we want. It appears that the device name needs to be '/dev/sda1' for Ubuntu AMI's instead of '/dev/xvda'.


shaidar:
Closing this as a I added commits from a different branch

PR #289, opened 2017-05-08
Changed device name

shaidar:
Instance provisioning is adding another block device instead of sizing the root volume which is what we want. It appears that the device name needs to be '/dev/sda1' for Ubuntu AMI's instead of '/dev/xvda'.


blarghmatey:
That makes sense and is likely an artifact of us using the marketplace AMI instead of the Amazon one.

blarghmatey:
👍

PR #291, opened 2017-05-09
Added environment variable

shaidar:
Added environment variable for grain matching so that tasks targeting elasticsearch role does not also apply to all elasticsearch instances.


blarghmatey:
Hard code this for operations. It is specific to logging so we dont want to accidentally run this in the mitx env

blarghmatey:
👍

PR #296, opened 2017-05-11
Added cluster tag in order to use it in discovery

shaidar:


shaidar:
This is not necessary since an elasticsearch tag is already specified and the cluster is in its own VPC

PR #298, opened 2017-05-12
Updated backup instances to have the consul-agent security group

blarghmatey:
In order for the backup/restore nodes to be able to resolve the
location of the resources they are backing up they need to be able to
properly connect their Consul agents for DNS lookups.


shaidar:
:thumbsup:

PR #299, opened 2017-05-12
Updated restore script for course assets

blarghmatey:
Updated the restore script for course assets so that it will execute
properly by downloading the backed up files and then rsyncing them to
their final destination.

### DIFF
--- a/salt/backups/templates/restore_course_assets.sh
+++ b/salt/backups/templates/restore_course_assets.sh
@@ -1,12 +1,15 @@
 #!/bin/bash
 set -e
 
-{% set backupdir = '/mnt/efs/{}'.format(settings.get('directory', 'course_assets')) %}
+{% set backupdir = '{}'.format(settings.get('directory', 'course_assets')) %}
 
 mkdir -p /mnt/efs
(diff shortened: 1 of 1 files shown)


salt/backups/templates/restore_course_assets.sh
```diff
@@ -1,12 +1,15 @@
...
 
 mkdir -p /mnt/efs
 mountpoint /mnt/efs || mount -t nfs4 -o nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2 $(curl -s http://169.254.169.254/latest/meta-data/placement/availability-zone).{{ settings.efs_id }}.efs.us-east-1.amazonaws.com:/ /mnt/efs
 
 PASSPHRASE={{ settings.duplicity_passphrase }} /usr/bin/duplicity restore \
-          --s3-use-server-side-encryption s3+http://odl-operations-backups/{{ settings.get('directory', 'course_assets') }}/ \
+          --s3-use-server-side-encryption s3+http://odl-operations-backups/{{ backupdir }}/ \
           --archive-dir /backups/.cache \
-          --force --tempdir /backups {{ backupdir }}/
+          --force --tempdir /backups /backups/{{ backupdir }}_copy/
```
shaidar: Why not restore directly to backupdir?

blarghmatey: Because Duplicity won't overwrite any of the existing data, which means that it won't be a full restore. The only way to overwrite data using Duplicity is to restore to an empty directory and then rsync.

shaidar:
:thumbsup:

PR #300, opened 2017-05-12
Format and mount 2nd drive

shaidar:
Added commands to format 2nd drive and then mount it. Also removed code that relates to creating Route53 records to Pillar data.

### DIFF
--- a/salt/orchestrate/aws/logging.sls
+++ b/salt/orchestrate/aws/logging.sls
@@ -12,8 +12,6 @@
 create_elasticsearch_index_template:
   elasticsearch_index_template.present:
     - name: 'logstash-*'
-    - tgt: 'G@roles:elasticsearch and G@environment:operations'
-    - tgt_type: compound
     - definition:
         template: 'logstash-*'
         settings:
--- a/salt/orchestrate/aws/logging.sls
+++ b/salt/orchestrate/aws/logging.sls
@@ -1,0 +1,89 @@
+#!jinja|yaml
+
+# Obtain the grains for one of the elasticsearch nodes
+{% set hosts = [] %}
+{% for host, grains in salt.saltutil.runner(
+    'mine.get',
+    tgt='G@roles:elasticsearch and G@environment:operations', fun='grains.item', tgt_type='compound'
+    ).items() %}
+{% do hosts.append(grains['external_ip']) %}
+{% endfor %}
+# PUT the mapper template into the ES _template index
+put_elasticsearch_mapper_template:
+  http.query:
+    - name: http://{{ hosts[0] }}:9200/_template/logstash
+    - data: '
+      {
+        "template":   "logstash-*",
+        "settings" : {
+          "index.refresh_interval" : "5s"
+        },
+        "mappings": {
+          "_default_": {
+            "_all": {
+              "enabled": false
+            },
+            "dynamic_templates": [
+              {
+                "strings": {
+                  "match_mapping_type": "string",
+                  "mapping": {
+                    "type": "string",
+                    "fields": {
+                      "raw": {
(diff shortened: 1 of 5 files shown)


salt/orchestrate/logging/init.sls
```diff
@@ -53,63 +51,50 @@ deploy_logging_cloud_map:
...
-    - tgt: 'P@roles:(elasticsearch|kibana|fluentd) and G@environment:operations'
+    - tgt: 'P@roles:(elasticsearch|kibana|fluentd) and G@environment:{{ ENVIRONMENT }}'
     - tgt_type: compound
     - require:
       - salt: load_pillar_data_on_logging_nodes
 
 build_logging_nodes:
   salt.state:
-    - tgt: 'P@roles:(elasticsearch|kibana|fluentd) and G@environment:operations'
+    - tgt: 'P@roles:(elasticsearch|kibana|fluentd) and G@environment:'
```
blarghmatey: This is missing the `{{ ENVIRONMENT }}` piece after the `:`

salt/orchestrate/aws/logging.sls
```diff
@@ -0,0 +1,74 @@
...
+  elasticsearch_index_template.present:
+    - name: 'logstash-*'
+    - definition:
+        template: 'logstash-*'
+        settings:
+          index.refresh_interval : '5s'
+        mappings:
+          _default_:
+            _all_:
+              enabled: false
```
blarghmatey: Do we have a particular reason for disabling the __all__ mapping, or is that just a carryover from the old map?

salt/orchestrate/logging/init.sls
```diff
@@ -53,63 +51,57 @@ deploy_logging_cloud_map:
...
-{% do hosts.append(grains['ec2:local_ipv4']) %}
-{% endfor %}
-register_log_aggregator_internal_dns:
-  boto_route53.present:
-    - name: log-input.private.odl.mit.edu
-    - value: {{ hosts }}
-    - zone: private.odl.mit.edu.
-    - record_type: A
+# Elasticsearch Curator is used for elasticsearch snapshots
+install_elasticsearch_curator:
```
blarghmatey: Do we have anything set up somewhere to configure curator?

salt/orchestrate/aws/logging.sls
shaidar: Mostly a carryover from the old map, but sounds like a good idea to keep it there:
https://github.com/elastic/elasticsearch/issues/19784

salt/orchestrate/logging/init.sls
shaidar: Yeah there are two config files required in addition to running it as cron job which I just need to add to this PR

salt/orchestrate/logging/configure_curator.sls
```diff
@@ -0,0 +1,38 @@
...
+install_elasticsearch_curator:
+  pip.installed:
+    - name: elasticsearch-curator
+    - require:
+      - salt: build_logging_nodes
+
+put_elasticsearch_snapshot_config:
+  http.query:
+    - name: http://elasticsearch.service.consul:9200/_snapshot/s3_repository/
+    - data: '
```
blarghmatey: Rather than quoting the whole thing you could use the `|` or `>-` operators for multiline YAML.

blarghmatey:
I think that the curator stuff fits better as an addition to the elasticsearch formula which we can then add to the top file for the Kibana node so that it gets executed in the highstate. Also, it looks like with this orchestrate configuration that the salt master would be responsible for executing the curator which I think is a better responsibility for the Kibana instance.

PR #301, opened 2017-06-01
Add gitreload as a systemd service for newer Ubuntu versions

shaidar:
This PR addresses [Issue#302](https://github.com/mitodl/salt-ops/issues/302)


blarghmatey:
Rather than using the `if/else` block for placing the upstart vs systemd configs you can use an approach similar to that found [here](https://github.com/mitodl/consul-formula/blob/master/consul/map.jinja#L23) and [here](https://github.com/mitodl/consul-formula/blob/master/consul/install.sls#L43)

salt/edx/gitreload.sls
```diff
@@ -155,6 +184,6 @@ start_gitreload:
   service.running:
     - name: gitreload
     - enable: True
+    - reload: True
```
blarghmatey: I'm not sure whether the gitreload service supports `reload` as an option. I think we should be fine leaving it as a restart operation.

blarghmatey:
Other than the `reload: True` I think this is ready to go. Good job :) 👍

PR #325, opened 2017-06-20
Fixed bug in AWS VPC orchestration and added tagging

blarghmatey:
- Added `business_unit` tag to a number of resources for expense tracking
- Updated variable setting in `mitx.sls` to sidestep bug associated with trying to populate ids for non-existent subnets that was happening in the `aws_env_macro` jinja template
- Added consul-agent security group to operations VPC


shaidar:
:thumbsup:

PR #336, opened 2017-06-22
Fix call to set subnet_id

shaidar:
Fix call to set subnet_id to use public2 which is in the same AZ as the EBS volume that we want to attach


blarghmatey:
👍

PR #338, opened 2017-06-22
Added missing tgt for salt.function

shaidar:


salt/orchestrate/edx/backup.sls
```diff
@@ -69,6 +69,8 @@ from s3 buckets when called. In order to accomodate that, we have an EBS volume
 by the ephemeral instance that is destroyed once backups are complete. #}
 create_attach_backup_volume:
   salt.function:
+    - tgt: 'G@roles:backups and G@environment:{{ ENVIRONMENT }}'
+    - tgt_type: compound
```
blarghmatey: This is the wrong target. This function needs to be executed on the master because it is the host hat has appropriate permissions to execute this action.

blarghmatey:
:+1:

PR #341, opened 2017-06-23
Added salt.function call since cloud.action is a function

shaidar:
cloud.action failed with the following error:
`State 'cloud.action' was not found in SLS 'orchestrate.edx.backup'
Reason: 'cloud.action' is not available.`
Added a salt.function and tried verifying syntax through the Online YAML parser


blarghmatey:
:+1:

Issue #260, opened 2017-03-30
Backup features


shaidar:
To improve our existing backup process for mitx-rp, it would be helpful to do the following:

- Add a slack notification on successful completion of backups. Currently, we only get notified if backups fail, but it would be helpful to get notified that they succeeded and are actually running.
- Setup a persistent AWS EBS volume that the backup instance mounts on creation and is used to store duplicity's archive data. Duplicity requires this data and if it does not find it, it'll have to download it from remote source everytime the backup instance is created.

blarghmatey:
While going through the current state of the backup orchestration I noticed that there are a number of `state.single` calls. I think that it might be easier to maintain and reason about if we group those into pre and post state files under the `backups` module here: https://github.com/mitodl/salt-ops/tree/master/salt/backups

Then we can use the `salt.state` function to call those SLS files with proper targeting, rather than trying to nest 3 levels of abstraction into a single function call. For instance, this section:
```salt
format_backup_drive:
  salt.function:
    - tgt: 'G@roles:backups and G@environment:{{ ENVIRONMENT }}'
    - tgt_type: compound
    - name: state.single
    - arg:
        - blockdev.formatted
    - kwarg:
        name: /dev/xvdb
        fs_type: ext4
    - require:
        - salt: create_attach_backup_volume

mount_backup_drive:
  salt.function:
    - tgt: 'G@roles:backups and G@environment:{{ ENVIRONMENT }}'
    - tgt_type: compound
    - name: state.single
    - arg:
        - mount.mounted
    - kwarg:
        name: /backups
        device: /dev/xvdb
        fstype: ext4
        mkmnt: True
        opts: 'relatime,user'

create_backup_directory:
  salt.function:
    - tgt: 'G@roles:backups and G@environment:{{ ENVIRONMENT }}'
    - tgt_type: compound
    - name: state.single
    - arg:
        - file.directory
    - kwargs:
        name: /backups/tmp
        makedirs: True
```
could be encapsulated in a single state file (e.g. `backups.prepare_disk` or `backups.mount_drive`) and then the orchestrate call could be:
```
mount_and_format_backup_drive:
  salt.state:
    - tgt: foo
    - sls:
      - backups.mount_drive
```

PR #345, opened 2017-06-28
Changes based on comments in Issue#260

shaidar:
[Issue#230](https://github.com/mitodl/salt-ops/issues/260#issuecomment-311486746) comments addressed and two files files have been created to prepare backup volume and then once backup scripts have completed, unmount and detach volume.

### DIFF
--- a/salt/backups/unmount_drive.sls
+++ b/salt/backups/unmount_drive.sls
@@ -10,7 +10,7 @@ detach_backup_volume:
     - func: boto_ec2.detach_volume
     - kwargs:
         tags:
-          Name: {{ backup.backup_volume_name }}
+          Name: {{ backup_volume_name }}
           device: /dev/xvdb
     - require:
         - salt: unmount_backup_drive
--- a/salt/backups/unmount_drive.sls
+++ b/salt/backups/unmount_drive.sls
@@ -1,3 +1,5 @@
+{% from "orchestrate/edx/backup" import backup with context %}
+
 unmount_backup_drive:
   mount.unmounted:
     - name: /backups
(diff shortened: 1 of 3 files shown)


salt/backups/unmount_drive.sls
```diff
@@ -0,0 +1,14 @@
...
+  mount.unmounted:
+    - name: /backups
+    - device: /dev/xvdb
+
+detach_backup_volume:
+  cloud.action:
+    - func: boto_ec2.detach_volume
+    - kwargs:
+        tags:
+          Name: {{ backup_volume_name }}
```
blarghmatey: This needs to be set somewhere. Right now it will throw an error for an undefined variable.

salt/backups/unmount_drive.sls
```diff
@@ -0,0 +1,16 @@
+{% from "orchestrate/edx/backup" import backup with context %}
```
blarghmatey: This is trying to import a `backup` jinja variable from the orchestrate state which doesn't appear to be set. I'm also not in love with the idea of importing anything from the orchestrate state, I think that introduces too much tight coupling between them.
It should be possible to use something like ` {% set backup_volume_name = 'odl-operations-backups-cache-{}'.format(salt.grains.get('environment')) %}`

salt/backups/unmount_drive.sls
```diff
@@ -0,0 +1,16 @@
...
+  mount.unmounted:
+    - name: /backups
+    - device: /dev/xvdb
+
+detach_backup_volume:
+  cloud.action:
+    - func: boto_ec2.detach_volume
+    - kwargs:
+        tags:
+          Name: {{ backup.backup_volume_name }}
```
blarghmatey: The variable set above needs to be reflected here.

blarghmatey:
:+1:

PR #347, opened 2017-06-28
Consolidated cloud maps to (mostly) use a single template

blarghmatey:
In order to remove a certain amount of WET (Write Everything Twice)
code I consolidated our cloud map templates to use a single template
since they were mostly a direct copy of each other and push some of
the variable pieces to the specific orchestrate states, thereby
increasing the data locality.

This also adds a `business_unit` tag and grain to the deployed instances.


salt/orchestrate/edx/backup.sls
```diff
@@ -88,41 +88,38 @@ create_attach_backup_volume:
...
+    - name: file.directory
     - arg:
-        - file.directory
-    - kwargs:
-        name: /backups/tmp
+        - /backups/tmp
+    - kwarg:
         makedirs: True
 
 execute_enabled_backup_scripts:
```
shaidar: Should this whole file be here?

salt/orchestrate/edx/services/mongodb.sls
```diff
@@ -28,6 +33,8 @@ generate_mongodb_cloud_map_file:
           - {{ salt.boto_secgroup.get_group_id(
             'vault-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
         subnetids: {{ subnet_ids }}
+        profile_overrides:
+          size: t2.medium
```
shaidar: Why are we overriding it here with the same instance type that's already specified?

shaidar:
:thumbsup:

PR #348, opened 2017-06-28
Stop instance on completion

shaidar:
Resized instance primary volume and instead of terminating it on completion, will be stopping it once backups are complete. Next time it runs, instance is powered on and backup scripts are called.

### DIFF
--- a/salt/orchestrate/edx/backup.sls
+++ b/salt/orchestrate/edx/backup.sls
@@ -60,42 +60,27 @@ deploy_backup_instance_to_{{ ENVIRONMENT }}:
                      'edx-{}'.format(VPC_RESOURCE_SUFFIX), vpc_name=VPC_NAME) }}
                 - {{ salt.boto_secgroup.get_group_id(
                      'consul-agent-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
+          block_device_mappings:
+            - DeviceName: /dev/xvda
+              Ebs.VolumeSize: 400
+              Ebs.VolumeType: gp2
     - require:
         - file: load_backup_host_cloud_profile
         - boto_iam_role: ensure_instance_profile_exists_for_backups
 
-{# Duplicity requires an archive directory otherwise it will have to create it and download files
-from s3 buckets when called. In order to accomodate that, we have an EBS volume that will be mounted
-by the ephemeral instance that is destroyed once backups are complete. #}
-create_attach_backup_volume:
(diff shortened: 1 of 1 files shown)


salt/orchestrate/edx/backup.sls
```diff
@@ -60,42 +60,27 @@ deploy_backup_instance_to_{{ ENVIRONMENT }}:
                      'edx-{}'.format(VPC_RESOURCE_SUFFIX), vpc_name=VPC_NAME) }}
                 - {{ salt.boto_secgroup.get_group_id(
                      'consul-agent-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
+          block_device_mappings:
+            - DeviceName: /dev/xvda
+              Ebs.VolumeSize: 400
+              Ebs.VolumeType: gp2
```
blarghmatey: We should still use a secondary drive because the Debian AMI doesn't properly adjust the size of the root volume if the underlying volume is larger. It is also simpler to handle unmounting/remounting a secondary drive for troubleshooting than if it is the root volume.

salt/orchestrate/edx/backup.sls
```diff
@@ -109,28 +94,17 @@ execute_enabled_backup_scripts:
...
-terminate_backup_instance_in_{{ ENVIRONMENT }}:
+stop_backup_instance_in_{{ ENVIRONMENT }}:
   salt.function:
-    - name: saltutil.runner
+    - name: cloud.action
     - tgt: 'roles:master'
     - tgt_type: grain
     - arg:
-        - cloud.destroy
+        - stop
```
blarghmatey: Since we are stopping the instance we should update the profile so that the stop action is to stop and not terminate. The ec2 cloud provider docs in salt should have examples.

salt/orchestrate/edx/backup.sls
```diff
@@ -61,13 +61,23 @@ deploy_backup_instance_to_{{ ENVIRONMENT }}:
                 - {{ salt.boto_secgroup.get_group_id(
                      'consul-agent-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
           block_device_mappings:
-            - DeviceName: /dev/xvda
+            - DeviceName: /dev/xvdb
```
blarghmatey: When we specify the block device mappings it's necessary to list both drives, otherwise it will still only create one.

blarghmatey:
:+1:

PR #350, opened 2017-06-30
Mimic restore to backup instance provisioning

shaidar:
Made the following changes:

- Setup restore instance orchestration similar to the way the backup instance is provisioned and configured. 
- Also reverted previous change to backup instance whereby specific subnet was specified to match the AZ of the EBS volume.  This is no longer needed, given that we're powering off the instance as opposed to terminating it and the EBS volume would be created along with the instance during initial provisioning.
- Renamed backups/init file to backups/backup for clarity and made the necessary changes throughout
- Created a new role called 'restores'
- Added a statement to create the hidden cache folder used by duplicity


salt/backups/templates/restore_mongodb.sh
```diff
@@ -4,14 +4,16 @@ set -e
...
+          --archive-dir {{ cachedir }} \
           --force --tempdir /backups/tmp/ {{ backupdir }}
 
 /usr/bin/mongorestore --host {{ settings.host }} \
                       --port {{ settings.get('port', 27017) }} \
                       --password {{ settings.password }} --username admin \
                       --authenticationDatabase admin \
-                      {{ backupdir }}
+                      --db {{ settings.database }} \
+                      --drop {{ backupdir }}/{{ settings.database }}
```
blarghmatey: To support mapping between source/destination databases for restore we will need to iterate over a mapping for the `mongorestore` piece, but this is good as a first step to bring everything in line. We can do that as a second PR or second commit.

blarghmatey:
:+1:

PR #351, opened 2017-06-30
Updated cloud profiles to pull AMI id from SDB

blarghmatey:
**Updated templates to pull AMI id from SDB**

**Updated orchestrate scripts to properly render profile templates**


shaidar:
:thumbsup:

PR #360, opened 2017-07-07
Updated to set file type and add dummy snapshot id for root volumes

blarghmatey:
When launching new instances with the Debian 9 AMI they were
complaining about duplicate device names. Updated the snapshot ID in
the profiles to point to a dummy snapshot as a means of avoiding that
issue.

Also updated the file metadata to properly identify the fact that the
profiles are YAML


shaidar:
Is that something with the AMI that's causing to complain about duplicate device names? Apart from that it's :thumbsup:

PR #361, opened 2017-07-07
Updated to allow for specifying the root volume via SDB

blarghmatey:


salt/orchestrate/aws/cloud_profiles/backup_host.conf
```diff
@@ -14,7 +14,7 @@ backup_host:
     roles:
       - backups
   block_device_mappings:
-    - DeviceName: /dev/xvda
+    - DeviceName: {{ salt.sdb.get('sdb://consul/debian_root_device')|default('/dev/xvda') }}
```
shaidar: Isn't the ubuntu root device the same name?

blarghmatey: Actually, the newer ubuntu devices use the /dev/sda1 device name...

blarghmatey: Also, the backup host uses Debian and not Ubuntu

shaidar:
:thumbsup:

PR #365, opened 2017-07-12
Modified restore state due to db naming mismatch

shaidar:
Current dogwood environment that we're upgrading uses a non-standard naming convention for mongodb and mysql. This change (including PR for pillar data) maps old name to new name.

### DIFF
--- a/salt/backups/templates/restore_mongodb.sh
+++ b/salt/backups/templates/restore_mongodb.sh
@@ -15,5 +15,5 @@ PASSPHRASE={{ settings.duplicity_passphrase }} /usr/bin/duplicity restore\
                       --port {{ settings.get('port', 27017) }} \
                       --password {{ settings.password }} --username admin \
                       --authenticationDatabase admin \
-                      --db {{ settings.database }} \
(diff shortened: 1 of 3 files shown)


salt/backups/templates/restore_mongodb.sh
```diff
@@ -15,5 +15,5 @@ PASSPHRASE={{ settings.duplicity_passphrase }} /usr/bin/duplicity restore\
                       --port {{ settings.get('port', 27017) }} \
                       --password {{ settings.password }} --username admin \
                       --authenticationDatabase admin \
-                      --db {{ settings.database }} \
-                      --drop {{ backupdir }}/{{ settings.database }}
+                      --db {{ settings.restore_to }} \
+                      --drop {{ backupdir }}/{{ settings.restore_from }}
```
blarghmatey: For mongo this won't quite work because all of the databases will be downloaded from S3 at once, so it would be best to  add a loop in the script to run the mapping all at once.

blarghmatey:
:+1:

PR #367, opened 2017-07-13
Added log output redirection for gitreload

blarghmatey:
The gitreload process was only writing to the systemd journal and not
the target log file. Updated to redirect stderr and stdout to the
target log file so they will get shipped to Elasticsearch


shaidar:
:thumbsup:

PR #368, opened 2017-07-17
Added states needed to run to build devstack instance using top file

shaidar:
Devstack runs all services on one host and we're planning on using our existing formulas and configs to build out the devstack instance. Made the following changes:
- In order to minimize the changes needed between a production deployment and devstack, we will be using consul and masterless-salt to orchestrate the devstack instance which necessitates the use of the top file.
- Created a pillar file to override some settings in the mysql formula to get the variables from one source of truth and that being the devstack file.

### DIFF
--- a/salt/top.sls
+++ b/salt/top.sls
@@ -102,7 +102,8 @@ base:
     - consul.dns_proxy
     - consul.tests
     - consul.tests.test_dns_setup
-    - mysql.server
+    - mysql
+    - mysql.remove_test_database
     - mongodb
     - mongodb.consul_check
     - rabbitmq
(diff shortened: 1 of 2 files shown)


salt/top.sls
```diff
@@ -96,3 +96,17 @@ base:
   'roles:restores':
     - match: grain
     - backups.restore
+  'G@roles:devstack and P@environment:dev':
+    - match: compound
+    - consul
+    - consul.dns_proxy
+    - consul.tests
+    - consul.tests.test_dns_setup
+    - mysql.server
```
blarghmatey: Are there any other  mysql states necessary?

blarghmatey:
:+1:

PR #370, opened 2017-07-17
Exclude devstack from top file

shaidar:
There's no need to include fluentd in devstack but even if we do want to include it, we'd also have to include the fluentd pillar data. In an effort of trying to keep things simple, it might be best to simply exclude devstack.

### DIFF
--- a/salt/top.sls
+++ b/salt/top.sls
@@ -1,5 +1,5 @@
 base:
-  '* and not G@roles:devstack':
+  'not G@roles:devstack':
     - match: compound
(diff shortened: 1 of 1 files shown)


salt/top.sls
```diff
@@ -1,5 +1,6 @@
 base:
-  '*':
+  '* and not G@roles:devstack':
```
blarghmatey: You don't need the `'* and `, you can just put `not G@roles:devstack`

PR #372, opened 2017-07-18
Overrride default settings in rabbitmq and elasticsearch formulas

shaidar:
Added pillar files to override default settings and added those files to top pillar file.


blarghmatey:
:+1:

PR #373, opened 2017-07-19
Removed s3 multiprocess on restore and added async upload

blarghmatey:
- Removed s3 multiprocess on restore to prevent I/O deadlock
- Added explicit supdirectory for S3 cache for course assets
- Removed command to purge directories on restore
- Added async upload parameter for backups

### DIFF
--- a/salt/backups/templates/restore_course_assets.sh
+++ b/salt/backups/templates/restore_course_assets.sh
@@ -2,14 +2,16 @@
 set -e
 
 {% set backupdir = '{}'.format(settings.get('directory', 'course_assets')) %}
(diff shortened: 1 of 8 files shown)


salt/backups/templates/restore_course_assets.sh
```diff
@@ -2,14 +2,14 @@
 set -e
 
 {% set backupdir = '{}'.format(settings.get('directory', 'course_assets')) %}
+{% set cachedir = '/backups/.cache/{}'.format(settings.get('directory', 'course_assets')) %}
```
shaidar: Shouldn't there be a mkdir -p {{ backupdir }} and {{ cachedir }}

shaidar:
:thumbsup:

PR #376, opened 2017-07-24
Added edx-worker as source group for services in env

blarghmatey:
In order for the edx workers to be able to access some of the other
hosts in the environment we missed adding the edx-worker group as an
allowed source.


shaidar:
:thumbsup:

PR #378, opened 2017-07-24
Additions to configure devstack

shaidar:
Added the following:
- consul config to run as a single server and registered all the needed services
- specified the edx_platform_branch in the devstack vars file
- set mysql_admin_username to 'root' as I was getting errors during config of failing to connect to db.
- added consul to devstack pillar file
- specified version of mariadb-client because the edx ansible run was failing without it complaining about dependencies.


pillar/consul_devstack.sls
```diff
@@ -0,0 +1,40 @@
...
+        check:
+          tcp: 'localhost:5672'
+          interval: 10s
+    elasticsearch_service:
+      service:
+        name: elasticsearch
+        port: 9200
+        check:
+          tcp: 'localhost:9200'
+          interval: 10s
```
blarghmatey: This will work but you can also just define all of the services as list items under a single `services` key in the `defaults` config file.

pillar/devstack.sls
```diff
@@ -300,3 +309,4 @@ edx:
     FORUM_USE_TCP: True
 
     edx_platform_repo: 'https://github.com/mitodl/edx-platform.git'
+    edx_platform_version: {{ edx_platform_branch }}
```
blarghmatey: For the edX codebase we should use our branch that is deployed to production (`mitx/ficus-1`)

blarghmatey:
Other than updating the edX branch :+1:

PR #382, opened 2017-07-31
Updated orchestration to allow for AMI builds

blarghmatey:
Added orchestrate routines to build an AMI from an empty base image
for edx app and edx workers, including assigning to the appropriate ELB.

@shaidar

### DIFF
--- a/docs/runbooks/blue_green_edx_deployment.md
+++ b/docs/runbooks/blue_green_edx_deployment.md
@@ -1,0 +1,31 @@
+# Background
+Blue/Green deployments are a methodology whereby a "blue" set of servers is running in production and a second "green"
+set is provisioned alongside them with the new software. Once the "green" servers are active they are added to the load
+balancer that controls access to the overall set of production hosts. Once the load balancer identifies the "green"
+nodes as healthy then the "blue" nodes are removed from circulation.
+
+# Building A Release Artifact
+There is an orchestrate script (`orchestrate.edx.build_ami`) that will create, configure, and then snapshot an edX
+application and worker node. As part of this process the value stored in the SDB URL
+`sdb://consul/edxapp-release-version` is incremented and included as part of the AMI name as well as being set in the
+`release-version` grain on the deployed instances.
+
+A second orchestrate script (`orchestrate.edx.update_edxapp_ami_sdb`) must then be run to update the AMI Ids stored in
+SDB that get rendered into the edX cloud profile. This updated AMI ID is what will be used for successive deployments of
+MITx production and/or QA instances.
+
+# Deploying
+Once the AMI has been created and the SDB values updated then the `orchestrate.edx` routine can be executed to deploy a
+new set of instances to the target environment. The currently running instances do not get overridden, nor do they
+prevent deployment of new instances, because the `release_version` is incorporated in the instance name/minion id. This
+results in a "blue" set of servers that are identified by having a `release-version` grain and minion id suffix
+containing the previous version. The "green" servers are identifiable by the fact that they have the current release
+version set on them.
+
+Once the "green" instances are deployed and the highstate run completes then you should run a smoke test against them to
+ensure that they can serve traffic and the MITx theme has been properly applied. You can then run the
+'orchestrate.aws.mitx_elb` routine to add them to the load balancer. As soon as they are marked as healthy and serve
+traffic then the "blue" hosts can be removed.
+
+As soon as you feel comfortable that the "green" hosts are able to serve traffic without error and no rollback is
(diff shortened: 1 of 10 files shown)


blarghmatey:
One minor problem with this approach is that it forces us into deploying new worker instances every time we want to redeploy app nodes. That's not a big problem because the overall deploy should be fairly fast (on the order of ~10 minutes as opposed to ~1 hour), but it is still unnecessary work.

salt/orchestrate/edx/build_ami.sls
```diff
@@ -0,0 +1,175 @@
...
+{% endfor %}
+{% endfor %}
+
+load_pillar_data_on_edx_base_nodes:
+  salt.function:
+    - name: saltutil.refresh_pillar
+    - tgt: 'P@roles:(edx-base|edx-base-worker) and G@environment:{{ ENVIRONMENT }}'
+    - tgt_type: compound
+    - require:
+        - salt: deploy_edx_cloud_map
```
shaidar: Unless I'm missing something, but I don't see where this one's defined

salt/orchestrate/edx/build_ami.sls
```diff
@@ -0,0 +1,176 @@
...
+    - kwarg:
+        vm_overrides:
+          tag:
+            business_unit: {{ BUSINESS_UNIT }}
+            environment: {{ ENVIRONMENT }}
+            purpose_prefix: {{ purpose_prefix }}
+          SecurityGroupId:
+            - {{ salt.boto_secgroup.get_group_id(
+            'edx-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
+            - {{ salt.boto_secgroup.get_group_id(
```
shaidar: Are we missing the edx-worker sec group?

salt/orchestrate/edx/build_ami.sls
```diff
@@ -0,0 +1,176 @@
...
+    - tgt: 'roles:master'
+    - tgt_type: grain
+    - name: sdb.set
+    - arg:
+        - {{ release_number }}
+    - require:
+        - boto_ec2: snapshot_edx_app_node
+        - boto_ec2: snapshot_edx_worker_node
+
+destroy_edx_base_instance:
```
shaidar: In addition to the two listed requirements, should we add some tests to verify that all's functional before we go ahead and destroy the old instances? Should we also add some notifications on success and failure?

docs/runbooks/blue_green_edx_deployment.md
```diff
@@ -0,0 +1,31 @@
...
+SDB that get rendered into the edX cloud profile. This updated AMI ID is what will be used for successive deployments of
+MITx production and/or QA instances.
+
+# Deploying
+Once the AMI has been created and the SDB values updated then the `orchestrate.edx` routine can be executed to deploy a
+new set of instances to the target environment. The currently running instances do not get overridden, nor do they
+prevent deployment of new instances, because the `release_version` is incorporated in the instance name/minion id. This
+results in a "blue" set of servers that are identified by having a `release-version` grain and minion id suffix
+containing the previous version. The "green" servers are identifiable by the fact that they have the current release
+version set on them.
```
shaidar: How do we filter them out from analytics since if I understand this correctly, the only difference between green and blue is the release-version, correct?

shaidar:
Apart from a few clarifying questions, I think this looks good :thumbsup:

docs/runbooks/blue_green_edx_deployment.md
blarghmatey: I'm not quite sure I understand what you mean by this. If you're referring to the tracking logs then no, we just want to indiscriminately capture all tracking logs.

salt/orchestrate/edx/build_ami.sls
blarghmatey: Adding notification would be useful, but this will primarily be run interactively so it's not as big of a need. The destruction here is just for the base instances that are for snapshotting, so it's desirable to destroy them each time. As far as tests, if the unit tests fail then the overall highstate will return False, thereby halting to execution.

docs/runbooks/blue_green_edx_deployment.md
shaidar: Yeah I was referring to tracking logs. I get shipping them indiscriminately to the logging cluster, but don't we need to filter them out from the s3 buckets given that they're not prod logs?

blarghmatey: No, we don't want to filter out any logs based on release version. The tracking events only occur if someone is actually using the platform so if there are logs to be had then we want to ship them.

PR #387, opened 2017-07-31
Added condition to edx.prod to check for devstack

shaidar:
Some of the functionality in edx.prod only applies to aws instances, so added a condition to check whether the instance has the devstack role and skip some of the calls if that's the case.
Removed edx.run_ansible call since it's redundant given that edx.prod includes it and added a variable to pillar devstack in order not to install the aws role.


blarghmatey:
:+1:

PR #388, opened 2017-08-02
Added states to run asset compilation on theme changes

blarghmatey:
If the `git.latest` state pulls any changes for the MITx theme then
the assets will be recompiled.


shaidar:
:thumbsup:

PR #389, opened 2017-08-02
Moved Nginx updates to separate sls file

blarghmatey:
The Nginx patches are not applicable to worker nodes which is causing
errors during their highstate. Moved to a separate SLS file and
updated the top file accordingly.


shaidar:
How come patch_nginx sls file isn't part of this PR?

blarghmatey:
What do you mean?... :innocent:

shaidar:
:thumbsup:

PR #390, opened 2017-08-03
Added explicit passing of flags to ansible runs

blarghmatey:
In order to make it easier to only update configuration values during blue/green deploys I added the ability to pass
flags to ansible playbook runs via pillar data and then exposed that as an env var for orchestrate routines. This lets
us specify `ANSIBLE_FLAGS=--tags install:configure` when deploying.


shaidar:
:thumbsup:

PR #391, opened 2017-08-03
Moved install_os_packages list to pillar data

shaidar:
Moved install_os_packages list to pillar data so that we can override them when/if we need to and but kept the ones we currently need available as default.
Currently have a need to override the mariadb-client install when building devstack and as it stands there's no easy way to do that. This PR would allow us to override the os_packages to be installed.

### DIFF
--- a/salt/edx/prod.sls
+++ b/salt/edx/prod.sls
@@ -44,7 +44,7 @@ configure_python_ppa_for_edx:
         - pkg: install_os_packages
 {% endif %}
 
-install_{{ os_packages }}:
+install_os_packages:
   pkg.installed:
     - pkgs: {{ os_packages }}
     - refresh: True
--- a/salt/edx/prod.sls
+++ b/salt/edx/prod.sls
@@ -13,6 +13,19 @@
 {% set theme_name = salt.pillar.get('edx:edxapp:THEME_NAME', None) -%}
 {% set theme_branch = salt.pillar.get('edx:edxapp:custom_theme:branch', 'mitx') -%}
 {% set theme_dir = salt.pillar.get('edx:edxapp:EDXAPP_COMPREHENSIVE_THEME_DIR', '/edx/app/edxapp/themes') -%}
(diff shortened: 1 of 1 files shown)


salt/edx/prod.sls
```diff
@@ -31,26 +44,16 @@ configure_python_ppa_for_edx:
         - pkg: install_os_packages
 {% endif %}
 
-install_os_packages:
+{% for os_pkg in os_packages %}
+install_{{ os_pkg }}:
```
blarghmatey: There's no need to turn this into a for loop, you just replace the list parameter of the `pkgs` attribute with the jinja variable.
```
install_os_package:
  pkg.installed:
    - pkgs: {{ os_packages }}
```

salt/edx/prod.sls
```diff
@@ -13,6 +13,19 @@
 {% set theme_name = salt.pillar.get('edx:edxapp:THEME_NAME', None) -%}
 {% set theme_branch = salt.pillar.get('edx:edxapp:custom_theme:branch', 'mitx') -%}
 {% set theme_dir = salt.pillar.get('edx:edxapp:EDXAPP_COMPREHENSIVE_THEME_DIR', '/edx/app/edxapp/themes') -%}
+{% set os_packages = salt.pillar.get('edx:edxapp:os_packages:os_pkg',
```
blarghmatey: I would skip the last `os_pkg` in the key space unless you have a specific reason for it. You could probably even get away with `edx:os_packages` or `edx:dependencies:os_packages`

salt/edx/prod.sls
```diff
@@ -31,21 +44,9 @@ configure_python_ppa_for_edx:
         - pkg: install_os_packages
 {% endif %}
 
-install_os_packages:
+install_{{ os_packages }}:
```
blarghmatey: You forgot to remove the `{{...}}` :smile:

shaidar: That was intentional but I can surely change that :-)

blarghmatey:
:+1:

PR #392, opened 2017-08-03
Updated mysql scripts to use mydumper/myloader

blarghmatey:
Switched to the mydumper/myloader utility to improve backup/restore speeds.


shaidar:
Are the binaries getting installed as part of the orchestrate state for that instance?

shaidar:
:thumbsup:

PR #393, opened 2017-08-08
Added explicit symlink override for edx DATA_DIR

blarghmatey:
The DATA_DIR value is hardcoded to the /edx/app/edxapp/data directory so the configuration repository currently relies on creating a symlink to the `edxapp_course_data_dir` location which we are overriding. Unfortunately, Ansible doesn't actually bother updating the symlink if it is already present on disk so we need to enforce the correct value at deploy time since we are pre-baking the AMIs to increase deployment speed.

### DIFF
--- a/salt/edx/prod.sls
+++ b/salt/edx/prod.sls
@@ -101,6 +101,17 @@ create_course_asset_symlink:
     - user: edxapp
     - group: www-data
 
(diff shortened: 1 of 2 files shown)


shaidar:
:thumbsup:

PR #394, opened 2017-08-16
State to run lms and cms migrations

shaidar:
Run lms and cms migrations as the edxapp user and added it to the top file for devstack as a last step to run.


salt/edx/migration.sls
```diff
@@ -0,0 +1,8 @@
+{% set edxapp_bin = '/edx/app/edxapp/venvs/edxapp/bin/' %}
+{% set migrations = ['lms', 'cms'] %}
+
+run_edxapp_{{  }}_migrations:
```
blarghmatey: Looks like you started writing a for loop but didn't quite get it finished.

blarghmatey:
:+1:

PR #395, opened 2017-08-25
Changes to complete devstack setup

shaidar:
- Added configuration value to rabbitmq to set the disk limit to override the existing calculation as the service was failing to start
- Creating a staff django user account so that someone using devstack already has an account that can be used
- Added new django_user state to top file and also reordered list so that rabbitmq.config runs after edxapp gets installed so that the disk space calculation is accurate
- Changed some values in devstack pillar


blarghmatey:
:+1:

PR #396, opened 2017-08-25
Added quotes around 'env' as it was failing

shaidar:
Added quotes around 'env' as it was failing. Renamed django devstack pillar to match with what's being created.


blarghmatey:
:+1:

PR #397, opened 2017-08-28
Created state file to install ruby

shaidar:
Created state file to install a ruby version that matches the Gemfile and added that to top file.


shaidar:
Ended up specifying version in pillar devstack instead.

PR #402, opened 2017-09-06
Added additional data to edx hosts for targeting

blarghmatey:
Added a `launch-date` grain so that we can add that to the target scoping to prevent us from accidentally running states
on existing instances when we are scaling up. This will let us increase the instance count in the
environment_settings.yml and run the orchestrate state to bring in new instances that are of the same release version.


shaidar:
:thumbsup:

PR #412, opened 2017-09-14
Updated minor issues with orchestration

blarghmatey:
- Added pre-commit hook to top file
- Removed requirements from update_sdb for destroying base instances for AMI
- Added step to restart supervisor processes after deploy


shaidar:
:thumbsup:

PR #414, opened 2017-09-15
Added transcripts s3 bucket to odl-video list based on Issue#413

shaidar:
Added transcripts s3 bucket to odl-video list based on [Issue#413](Added transcripts s3 bucket to odl-video list based on Issue#413)


blarghmatey:
:+1:

PR #418, opened 2017-09-21
Relocated files and fixed a few bugs

shaidar:
- Moved files under the orchestrate folder since it makes more sense given that we're configuring cloudfront and s3 buckets
- Fixed typo and variable name
- Added cloudfront distribution name to boto_cloudfront.get_distribution and moved it inside one of the for loops as the name is a mandatory field


blarghmatey:
:+1:

PR #419, opened 2017-09-21
[WIP] Adding orchestration of app environments

blarghmatey:
Adding orchestrate states for being able to build environments to deploy web apps

### DIFF
--- a/salt/orchestrate/aws/create_vpc.sls
+++ b/salt/orchestrate/aws/create_vpc.sls
@@ -23,6 +23,9 @@ create_{{ ENVIRONMENT }}_vpc:
     - tags:
         Name: {{ VPC_NAME }}
         business_unit: {{ BUSINESS_UNIT }}
+        Department: {{ BUSINESS_UNIT }}
+        OU: {{ BUSINESS_UNIT }}
+        Environment: {{ ENVIRONMENT }}
 
 create_{{ ENVIRONMENT }}_internet_gateway:
   boto_vpc.internet_gateway_present:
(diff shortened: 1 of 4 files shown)


salt/orchestrate/aws/create_vpc.sls
```diff
@@ -23,6 +23,9 @@ create_{{ ENVIRONMENT }}_vpc:
     - tags:
         Name: {{ VPC_NAME }}
         business_unit: {{ BUSINESS_UNIT }}
+        Department: {{ BUSINESS_UNIT }}
+        OU: {{ BUSINESS_UNIT }}
+        Environment: {{ ENVIRONMENT }}
```
shaidar: If we're now adding all those tags to a resource, is there some ways to bundle it up into a list and add it to env_settings and just have one statement to iterate through it?

salt/orchestrate/aws/elasticache.sls
```diff
@@ -0,0 +1,102 @@
+{% set ENVIRONMENT = salt.environ.get('ENVIRONMENT') %}
```
shaidar: No default qa environment?

salt/orchestrate/aws/elasticache.sls
```diff
@@ -0,0 +1,102 @@
...
+{% set default_port = {
+    'redis': 6379,
+    'memcached': 11211
+} %}
+
+create_{{ ENVIRONMENT }}_elasticache_security_group:
+  boto_secgroup.present:
+    - name: elasticache-{{ VPC_RESOURCE_SUFFIX }}
+    - vpc_name: {{ VPC_NAME }}
+    - description: ACL for PostGreSQL RDS servers
```
shaidar: Is this accurate description?

salt/orchestrate/aws/create_vpc.sls
blarghmatey: I think that it would make sense for us to encapsulate the tagging of resources into a reusable component (jinja variable, macro, python lib, etc.) but that is something for us to think about as a separate task.

shaidar:
:thumbsup:

PR #436, opened 2017-09-28
Added a state file to configure reddit ini file

shaidar:
Added a state file to configure reddit ini file based on a jinja template and restart the service


blarghmatey:
:+1:

PR #437, opened 2017-09-29
Added state to orchestrate reddit build and run the highstate

shaidar:


salt/orchestrate/reddit/init.sls
```diff
@@ -0,0 +1,109 @@
...
+    - name: /etc/salt/cloud.maps.d/{{ VPC_RESOURCE_SUFFIX }}_{{ app_name }}_map.yml
+    - source: salt://orchestrate/aws/map_templates/instance_map.yml
+    - template: jinja
+    - makedirs: True
+    - context:
+        business_unit: {{ BUSINESS_UNIT }}
+        environment_name: {{ ENVIRONMENT }}
+        purpose_prefix: {{ PURPOSE_PREFIX }}
+        securitygroupids:
+          webapp: {{ salt.boto_secgroup.get_group_id(
```
blarghmatey: This should be a list, not a dictionary

salt/top.sls
```diff
@@ -138,3 +138,6 @@ base:
     - match: grain
     - scylladb.configure
     - scylladb.tests
+  'G@roles:reddit and P@environment:(rc|production)-apps':
```
blarghmatey: This doesn't need to be predicated on environment. We should just match on the `roles:reddit` grain with a match type of grain.

salt/orchestrate/reddit/init.sls
```diff
@@ -0,0 +1,109 @@
...
+
+build_{{ app_name }}_nodes:
+  salt.state:
+    - tgt: 'P@roles:{{ app_name }} and G@environment:{{ ENVIRONMENT }}'
+    - tgt_type: compound
+    - highstate: True
+    - require:
+        - salt: deploy_consul_agent_to_{{ app_name }}_nodes
+        - salt: restart_consul_service_to_load_updated_configs
+
```
blarghmatey: This is just missing execution of the post_deploy state from the reddit formula. You'll just want to run that using the `subset: 1` parameter so that we don't run into race conditions with multiple hosts trying to provision the same values on the databases.

PR #447, opened 2017-10-04
Stripped the hyphen at the end of the string

shaidar:
Stripped the hyphen at the end of the string since it's not an allowed terminating character.


blarghmatey:
:+1:

PR #453, opened 2017-10-16
Add entry to /etc/hosts to point the lms DNS name

shaidar:
Add entry to /etc/hosts to point the lms DNS name to localhost to avoid routing traffic intended to the localhost all the way back out to ELB


blarghmatey:
:+1:

blarghmatey:
If this works the way that we want it to then we should add that state to the top file for the edx nodes as well so that it gets run on highstate.

PR #456, opened 2017-10-20
Removed xsiftx state file and templates

shaidar:
Removed xsiftx state file and templates as this service is no longer being installed


blarghmatey:
Gotta love the red diffs :) :+1:

PR #458, opened 2017-10-27
Added conditional to only apply sticky sessions to draft instances

shaidar:
Added conditional to only apply sticky sessions to draft instances and skip live ones since we're now using a memcached cluster


blarghmatey:
:+1:

PR #460, opened 2017-11-06
Changed values to point to ginkgo release

shaidar:


blarghmatey:
We have a ginkgo branch in our edX fork. Also, I would make the branches of the config repo and the edX repo use environment variables, or have a set of values for each version and use an if block to choose between them based on environment variables.

blarghmatey:
:+1:

PR #461, opened 2017-11-07
State file to backup tracking logs from retired mitx instances

shaidar:
State file to backup tracking logs from retired mitx instances to S3 buckets

### DIFF
--- a/salt/backups/backup_tracking.sls
+++ b/salt/backups/backup_tracking.sls
@@ -1,0 +1,18 @@
+{% set edx_tracking_bucket = 'odl-residential-tracking-data' %}
+{% set edx_tracking_bucket_creds = salt.vault.read('aws-mitx/creds/read-write-{bucket}'.format(bucket=edx_tracking_bucket)) %}
+{% set edx_tracking_local_folder = '/edx/var/log/tracking' %}
+{% set instance_id = salt.environ.get('MINION_ID') %}
+
+tar_tracking_data:
+  cmd.run:
+    - name: 'tar -czf edx_tracking_{{ instance_id }}.tgz *'
+    - cwd: {{ edx_tracking_local_folder }}
+
+upload_tar_to_s3:
+  module.run:
+    - name: s3.put
+    - bucket: {{ edx_tracking_bucket }}
+    - path: 'retired-instance-logs'
+    - keyid: {{ edx_tracking_bucket_creds.data.access_key }}
+    - key: {{ edx_tracking_bucket_creds.data.secret_key }}
+    - local_file: {{ edx_tracking_local_folder }}/edx_tracking_{{ instance_id }}.tgz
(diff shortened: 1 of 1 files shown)


salt/backups/backup_tracking.sls
```diff
@@ -0,0 +1,18 @@
+{% set edx_tracking_bucket = 'odl-residential-tracking-data' %}
+{% set edx_tracking_bucket_creds = salt.vault.read('aws-mitx/creds/read-write-{bucket}'.format(bucket=edx_tracking_bucket)) %}
```
blarghmatey: This isn't going to work because this state file will be compiled on the minion which doesn't have permissions to read from Vault. We will need to pass this value via pillar data, or modify or Vault config to create a policy which permits access to that route and associates the policy with the minion via EC2 auth or approle or something like that.

salt/backups/backup_tracking.sls
```diff
@@ -0,0 +1,18 @@
...
+{% set edx_tracking_local_folder = '/edx/var/log/tracking' %}
+{% set instance_id = salt.environ.get('MINION_ID') %}
```
blarghmatey: You can get the minion id from the `id` grain. `salt.grains.get('id')`

salt/backups/backup_tracking.sls
```diff
@@ -0,0 +1,18 @@
...
+
+tar_tracking_data:
+  cmd.run:
+    - name: 'tar -czf edx_tracking_{{ instance_id }}.tgz *'
+    - cwd: {{ edx_tracking_local_folder }}
```
blarghmatey: You can use the `module.run` state to execute the `archive.tar` execution module as an alternative to using `cmd.run`. If you do stick with `cmd.run` then you should add the `creates` parameter so that if it gets run multiple times it will short circuit on recreating the archive.

salt/backups/backup_tracking.sls
```diff
@@ -1,18 +1,20 @@
...
+{% set edx_tracking_bucket_backup_keyid = salt.pillar.get('backups:mitx_residential_tracking:aws_keyid') %}
 {% set edx_tracking_local_folder = '/edx/var/log/tracking' %}
-{% set instance_id = salt.environ.get('MINION_ID') %}
+{% set instance_id = salt.grains.get('id') %}
 
 tar_tracking_data:
   cmd.run:
     - name: 'tar -czf edx_tracking_{{ instance_id }}.tgz *'
     - cwd: {{ edx_tracking_local_folder }}
+    - creates: edx_tracking_{{ instance_id }}.tgz
```
blarghmatey: I think this might need to be an absolute path to the file

blarghmatey:
:+1:

salt/backups/backup_tracking.sls
shaidar: Spent a good amount of time yesterday afternoon looking at docs and trying to get `module.run` to execute `archive.tar` but couldn't get it to work for me and that's when I switched to using `cmd.run`

PR #465, opened 2017-11-08
Added a function to set the edx_codename in sdb so that it's easier t…

shaidar:
Added a function to set the edx_codename in sdb so that it's easier to read the value in update_edxapp_ami_sdb state file. The ami_id in set based on an ec2 query that is not updated with the codename.


blarghmatey:
:+1:

PR #467, opened 2017-11-08
Changed CacheClusterId to only be purpose name

shaidar:
Changed CacheClusterId to only be purpose name to better match with the edx build ami elasticache query. The 20 character restriction on the name is causing some complications where if purpose name is too long, we would have a mistmatch between the two and the build would fail.


blarghmatey:
:+1:

PR #473, opened 2017-11-13
Changed sdb.set to edx_worker_ginkgo...

shaidar:
sdb.get is looking for edx_worker_ginkgo... while the set command was setting it as edx_ginkgo_worker. Changed it to be edx_worker_ginkgo...


blarghmatey:
:+1:

PR #474, opened 2017-11-13
Use edx-instance-role to upload old tracking logs to an S3 bucket

shaidar:
Changed the following:
- Use edx-instance-role to upload old tracking logs to an S3 bucket.
- Added a check to verify that the S3 bucket exists.
- Removed AWS key and keyid


salt/backups/backup_tracking.sls
```diff
@@ -1,9 +1,28 @@
...
+
+ensure_instance_profile_exists_for_tracking:
+  boto_iam_role.present:
+    - name: edx-instance-role
+    - delete_policies: False
+    - policies:
+        edx-old-tracking-logs-policy:
+          Statement:
+            - Action:
+                - s3:*
```
blarghmatey: This should be scoped to only those capabilities necessary to upload the files (you can refer to the vault role settings for that)

blarghmatey: https://github.mit.edu/mitx-devops/salt-pillar/blob/master/vault/init.sls#L124

blarghmatey:
:+1:

PR #492, opened 2017-12-08
Split fluentd-servers from the logging orchestrate state

shaidar:
Split fluentd-servers from the logging orchestrate state which includes elasticsearch and kibana into its own state in order to be able to deploy fluentd-server instances on their own.

### DIFF
--- a/salt/orchestrate/aws/security_groups.sls
+++ b/salt/orchestrate/aws/security_groups.sls
@@ -269,24 +269,3 @@ create_webapp_security_group:
         Department: {{ BUSINESS_UNIT }}
         OU: {{ BUSINESS_UNIT }}
         Environment: {{ ENVIRONMENT }}
-
-create_fluentd_aggregator_security_group:
-  boto_secgroup.present:
-    - name: fluentd-{{ VPC_RESOURCE_SUFFIX }}
-    - vpc_name: {{ VPC_NAME }}
-    - description: ACL for Fluentd aggretators
-    - rules:
-        {% for portnum in [443, 5001] %}
-        - ip_protocol: tcp
-          from_port: {{ portnum }}
-          to_port: {{ portnum }}
-          cidr_ip:
-            - 0.0.0.0/0
-            - '::/0'
-        {% endfor %}
-    - tags:
-        Name: fluentd-{{ VPC_RESOURCE_SUFFIX }}
-        business_unit: {{ BUSINESS_UNIT }}
-        Department: {{ BUSINESS_UNIT }}
-        OU: {{ BUSINESS_UNIT }}
-        Environment: {{ ENVIRONMENT }}
--- a/salt/orchestrate/aws/security_groups.sls
+++ b/salt/orchestrate/aws/security_groups.sls
@@ -269,3 +269,24 @@ create_webapp_security_group:
         Department: {{ BUSINESS_UNIT }}
         OU: {{ BUSINESS_UNIT }}
         Environment: {{ ENVIRONMENT }}
+
+create_fluentd_aggregator_security_group:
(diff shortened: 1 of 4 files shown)


salt/orchestrate/operations/services/fluentd.sls
```diff
@@ -0,0 +1,102 @@
...
+    - template: jinja
+    - makedirs: True
+    - context:
+        environment_name: {{ ENVIRONMENT }}
+        num_instances: {{ INSTANCE_COUNT }}
+        service_name: {{ app_name }}
+        roles:
+          - fluentd-server
+          - log-aggregator
+          - {{ app_name }}
```
blarghmatey: These roles are already defined in the cloud profile so it is not necessary to duplicate them here.

salt/orchestrate/operations/services/fluentd.sls
```diff
@@ -0,0 +1,102 @@
...
+    - require:
+        - salt: load_pillar_data_on_{{ app_name }}_nodes
+
+deploy_consul_agent_to_{{ app_name }}_nodes:
+  salt.state:
+    - tgt: 'G@roles:{{ app_name }} and G@environment:{{ ENVIRONMENT }}'
+    - tgt_type: compound
+    - sls:
+        - consul
+        - consul.dns_proxy
```
blarghmatey: This is largely a relic of the edX build process and is included in the highstate for the nodes, so it shouldn't be necessary to include as a distinct step in the orchestrate.

salt/orchestrate/operations/services/fluentd.sls
```diff
@@ -0,0 +1,102 @@
...
+
+restart_{{ app_name }}_service:
+  salt.function:
+    - tgt: 'G@roles:{{ app_name }} and G@environment:{{ ENVIRONMENT }}'
+    - tgt_type: compound
+    - name: service.restart
+    - arg:
+        - {{ app_name }}
+    - require:
+        - salt: build_{{ app_name }}_nodes
```
blarghmatey: The states for FluentD already include restarts based on config changes, so this shouldn't be necessary as a deploy step.

salt/orchestrate/aws/security_groups.sls
```diff
@@ -269,3 +269,24 @@ create_webapp_security_group:
...
+          cidr_ip:
+            - 0.0.0.0/0
+            - '::/0'
+        {% endfor %}
+    - tags:
+        Name: fluentd-{{ VPC_RESOURCE_SUFFIX }}
+        business_unit: {{ BUSINESS_UNIT }}
+        Department: {{ BUSINESS_UNIT }}
+        OU: {{ BUSINESS_UNIT }}
+        Environment: {{ ENVIRONMENT }}
```
blarghmatey: Maybe move this security group definition into the orchestrate step, or as a separate state file specific to the operations network since this configuration isn't applicable to other environments (yet).

salt/orchestrate/operations/services/fluentd.sls
```diff
@@ -24,6 +24,20 @@ create_fluentd_aggregator_security_group:
...
+    ).items() %}
+{% do hosts.append(grains['external_ip']) %}
+{% endfor %}
+register_log_aggregator_dns:
+  boto_route53.present:
+    - name: log-input.odl.mit.edu
+    - value: {{ hosts }}
+    - zone: odl.mit.edu.
+    - record_type: A
+
```
blarghmatey: This will need to be the last step run after the instances are deployed and provisioned. It will also need an explicit call to refresh the mine data before this step gets executed.

blarghmatey:
:+1:

PR #493, opened 2017-12-08
Updated the ELB orchestration to directly use the DNS entries

blarghmatey:
Now that we are exposing the studio service on the LMS servers we need to ensure that the DNS entries are applied
properly, which means that we no longer need to special case the logic when provisioning the ELBs.


blarghmatey:
Don't merge this until we release Ginkgo to production

shaidar:
👍

PR #495, opened 2017-12-11
Separated kibana orchestrate state into its own state file

shaidar:
Separated kibana orchestrate state into its own state file and removed logging related files since we now have split all the roles out into individual ones.


salt/orchestrate/aws/cloud_profiles/kibana.conf
```diff
@@ -5,12 +5,7 @@ kibana:
   image: {{ salt.sdb.get('sdb://consul/debian_ami_id')|default('ami-b14ba7a7', True) }}
   ssh_username: admin
   iam_profile: kibana-instance-role
-  grains:
-    roles:
-      - kibana
   ssh_interface: private_ips
-  tag:
-    role: kibana
```
blarghmatey: Was there a particular reason for removing the `role` tag and grain?

shaidar: Yeah, I figured that we now have that standard instance_map template which is populated by the state file with the appropriate tags/roles, so why include those in the cloud profile. Unless I'm missing something?

blarghmatey:
:+1:

PR #496, opened 2017-12-12
Added launch_date tag

shaidar:
Added launch_date tag to use when filtering so that we don't apply the state to existing instances'


shaidar:
Already spoke to Tobias about this and he mentioned adding this tag so that we when running the orchestrate we don't impact the existing fluentd instances.

PR #498, opened 2017-12-13
Added CORS rules for S3 buckets for edX

blarghmatey:


salt/orchestrate/edx/init.sls
```diff
@@ -106,6 +106,16 @@ create_edx_s3_bucket_{{ bucket }}_{{ PURPOSE_PREFIX }}-{{ type }}_{{ ENVIRONMENT
     - region: us-east-1
     - Versioning:
        Status: "Enabled"
+    - CORSRules:
```
shaidar: I would be in favor of adding an `if` statement to filter out the other s3 buckets that might not necessarily need to have this CORS rule.

shaidar:
👍

PR #499, opened 2017-12-19
Files and folder don't appear to be in use anymore

shaidar:


blarghmatey:
:+1:  Yay for red diffs!

PR #501, opened 2017-12-19
ETL state file to run nightly mitx queries

shaidar:
ETL state file that installs python3, clones mitx_etl repo and schedules script to run as a nightly cron task.

### DIFF
--- a/salt/edx/mitx_etl.sls
+++ b/salt/edx/mitx_etl.sls
@@ -31,7 +31,7 @@ install_mitx_residential_etl_requirements:
       - git: clone_mitx_etl_repo
       - pkg: install_etl_os_depencies
 
-{% set settings = salt.pillar.get('edx:mitx_etl:mitx_residential_etl:settings', {}).items() %}
+{% set settings = salt.pillar.get('edx:mitx_etl:mitx_residential_etl:settings', {}) %}
 mitx_residential_etl_config:
   file.managed:
     - name: /mitx/settings.json
@@ -49,7 +49,7 @@ ensure_upload_bucket_exists:
 
 add_task_to_cron:
   cron.present:
-    - name: 'source /mitx/mitx_etl/bin/python3 /mitx/mitx_residential_etl.py'
+    - name: '/mitx/mitx_etl/bin/python3 /mitx/mitx_residential_etl.py'
     - comment: mitx_residential_etl_script
     - special: '@daily'
     - require:
--- a/salt/edx/mitx_etl.sls
+++ b/salt/edx/mitx_etl.sls
@@ -1,0 +1,57 @@
+install_python3_pip:
+  pkg.installed:
+    - pkgs:
+        - python3
+        - python3-pip
+    - refresh: True
+    - require:
+      - pkg: install_python_3
+
+create_mitx_directory:
+  file.directory:
+    - name: /mitx
+
+clone_mitx_etl_repo:
+  git.latest:
+    - name: https://github.com/mitodl/odl-etl
+    - target: /mitx
+    - force_clone: True
+    - require:
+      - pkg: install_python_3
+
+install_mitx_residential_etl_requirements:
+  virtualenv.managed:
+    - name: /mitx/mitx_etl
+    - system_site_packages: False
+    - requirements: /mitx/requirements.txt
+    - env_vars:
+        PATH_VAR: '/usr/local/bin/pip3'
+    - require:
+      - git: clone_mitx_etl_repo
+      - pkg: install_python3_pip
+
+{% for key, value in salt.pillar.get('edx:mitx_etl:mitx_residential_etl:settings', {}).items() %}
+mitx_residential_etl_config:
+  file.managed:
+    - name: /mitx/settings.json
+    - contents: |
+        settings: {{ value|json(indent=2, sort_keys=True) |indent(8) }}
+    - require:
+      - git: clone_mitx_etl_repo
+{% endfor %}
+
+ensure_upload_bucket_exists:
+  boto_s3_bucket.present:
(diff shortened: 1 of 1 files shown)


salt/edx/mitx_etl.sls
```diff
@@ -0,0 +1,57 @@
...
+    - system_site_packages: False
+    - requirements: /mitx/requirements.txt
+    - env_vars:
+        PATH_VAR: '/usr/local/bin/pip3'
+    - require:
+      - git: clone_mitx_etl_repo
+      - pkg: install_python3_pip
+
+{% for key, value in salt.pillar.get('edx:mitx_etl:mitx_residential_etl:settings', {}).items() %}
+mitx_residential_etl_config:
```
blarghmatey: There's no need to use a for loop here. You can just pass the entire settings object through the JSON filter to render it out as a file. This will end up overwriting the file multiple times with each value.

salt/edx/mitx_etl.sls
```diff
@@ -0,0 +1,57 @@
...
+ensure_upload_bucket_exists:
+  boto_s3_bucket.present:
+    - Bucket: {{ settings.S3Bucket.bucket }}
+    - Versioning:
+        Status: Enabled
+    - region: us-east-1
+
+add_task_to_cron:
+  cron.present:
+    - name: 'source /mitx/mitx_etl/bin/activate && python3 /mitx/mitx_residential_etl.py'
```
blarghmatey: Rather than sourcing the activate script (which can sometimes behave oddly in cron tasks) you can just use the full path to the python3 binary (e.g. `/mitx/mitx_etl/bin/python3 ...`)

salt/edx/mitx_etl.sls
```diff
@@ -0,0 +1,57 @@
...
+    - comment: mitx_residential_etl_script
+    - special: '@daily'
+    - require:
+      - file: mitx_residential_etl_config
+      - boto_s3_buclet: ensure_upload_bucket_exists
```
blarghmatey: typo on `bucket`

salt/edx/mitx_etl.sls
```diff
@@ -0,0 +1,57 @@
...
+    - force_clone: True
+    - require:
+      - pkg: install_python_3
+
+install_mitx_residential_etl_requirements:
+  virtualenv.managed:
+    - name: /mitx/mitx_etl
+    - system_site_packages: False
+    - requirements: /mitx/requirements.txt
+    - env_vars:
```
blarghmatey: This is missing a specification of the python version to use to build the virtualenv, so it will likely default to Python2 which is the system default (I believe)

salt/edx/mitx_etl.sls
```diff
@@ -0,0 +1,57 @@
...
+  file.directory:
+    - name: /mitx
+
+clone_mitx_etl_repo:
+  git.latest:
+    - name: https://github.com/mitodl/odl-etl
+    - target: /mitx
+    - force_clone: True
+    - require:
+      - pkg: install_python_3
```
blarghmatey: This should require the `create_mitx_directory` step

salt/edx/mitx_etl.sls
```diff
@@ -0,0 +1,57 @@
+install_python3_pip:
+  pkg.installed:
+    - pkgs:
+        - python3
```
blarghmatey: Since we're depending on Git to load the etl repo we should pass it as a required package here and maybe rename the id to something like `install_etl_os_dependencies`

shaidar: Added git pkg and renamed state id

salt/edx/mitx_etl.sls
shaidar: Changed

salt/edx/mitx_etl.sls
shaidar: Added path to python3 executable

salt/edx/mitx_etl.sls
shaidar: Removed for loop

salt/edx/mitx_etl.sls
shaidar: Changed source command with full path to python3

salt/edx/mitx_etl.sls
shaidar: Fixed typo

salt/edx/mitx_etl.sls
```diff
@@ -17,28 +16,29 @@ clone_mitx_etl_repo:
...
+    - python: /usr/bin/python3
     - env_vars:
         PATH_VAR: '/usr/local/bin/pip3'
     - require:
       - git: clone_mitx_etl_repo
-      - pkg: install_python3_pip
+      - pkg: install_etl_os_depencies
 
-{% for key, value in salt.pillar.get('edx:mitx_etl:mitx_residential_etl:settings', {}).items() %}
+{% set settings = salt.pillar.get('edx:mitx_etl:mitx_residential_etl:settings', {}).items() %}
```
blarghmatey: You don't want to have the `.items()` call at the end, it should just be the `pillar.get` call.

salt/edx/mitx_etl.sls
```diff
@@ -49,9 +49,9 @@ ensure_upload_bucket_exists:
 
 add_task_to_cron:
   cron.present:
-    - name: 'source /mitx/mitx_etl/bin/activate && python3 /mitx/mitx_residential_etl.py'
+    - name: 'source /mitx/mitx_etl/bin/python3 /mitx/mitx_residential_etl.py'
```
blarghmatey: You don't need the `source` command at all since you are providing the full path to the python3 binary.

blarghmatey:
:+1:

PR #502, opened 2017-12-20
Updated rabbit and mongo orchestrate to populate vault password

blarghmatey:
Updated the orchestrate states to make them more re-entrant by storing the generated admin password in vault to be retrieved in pillar data for future runs.


shaidar:
:thumbsup:

PR #507, opened 2017-12-22
Moved setting vault password in jinja during orchestrate to yaml func…

blarghmatey:
…tion

### DIFF
--- a/salt/orchestrate/services/rabbitmq.sls
+++ b/salt/orchestrate/services/rabbitmq.sls
@@ -3,7 +3,15 @@
 {% set rabbitmq_admin_password = salt.vault.read('secret-{}/{}/rabbitmq-admin-password'.format(BUSINESS_UNIT, ENVIRONMENT)).data.value %}
 {% if not rabbitmq_admin_password %}
 {% set rabbitmq_admin_password = salt.random.get_str(42) %}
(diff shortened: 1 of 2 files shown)


salt/orchestrate/services/rabbitmq.sls
```diff
@@ -2,8 +2,16 @@
  ENVIRONMENT, BUSINESS_UNIT, subnet_ids with context %}
 {% set rabbit_admin_password = salt.vault.read('secret-{}/{}/rabbitmq-admin-password'.format(BUSINESS_UNIT, ENVIRONMENT)).data.value %}
 {% if not rabbit_admin_password %}
-{% set rabbitmq_admin_password = salt.random.get_str(42) %}
-{% salt.vault.write('secret-{}/{}/rabbitmq-admin-password'.format(BUSINESS_UNIT, ENVIRONMENT), value=rabbitmq_admin_password) %}
+{% set rabbit_admin_password = salt.random.get_str(42) %}
```
shaidar: Should we just use `rabbitmq_admin_password` throughout since that's what we're calling it in vault?

blarghmatey: Probably not a bad idea. I'll update.

shaidar:
👍

Issue #512, opened 2017-12-27
Crawl leases in Vault to notify of imminent expiration


blarghmatey:
In order to get early warning of leases that will be expiring we want to write a task that will crawl the Vault API of all leases to notify of any that are going to be expiring in the near future.

PR #513, opened 2017-12-28
Added a reactor to notify on Slack when a lease will be expiring.

blarghmatey:
#512 

#### What's this PR do?
Adds a reactor to notify in Slack when an event is fired to signify a lease expiring in Vault

There have been a handful of occasions where a set of Vault credentials expired and resulted in systems failing without our knowledge. This will provide advance warning of this type of situation.

### DIFF
--- a/salt/reactors/vault/alert_expiring_leases.sls
+++ b/salt/reactors/vault/alert_expiring_leases.sls
@@ -1,0 +1,12 @@
+alert_on_lease_near_expiration:
+  local.slack.post_message:
+    - tgt: 'roles:master'
+    - expr_form: grain
+    - kwarg:
+        channel: "#devops"
+        message: |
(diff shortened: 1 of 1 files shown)


salt/reactors/vault/alert_expiring_leases.sls
```diff
@@ -0,0 +1,12 @@
+alert_on_lease_near_expiration:
+  local.slack.post_message:
+    - tgt: 'roles:master'
+    - expr_form: grain
+    - kwarg:
+        channel: "#devops"
+        message: |
+          @channel The Vault lease `{{ data['id'] }}` will be expiring at `{{ data['expire_time'] }}`.
```
shaidar: No quotes needed here?

blarghmatey: The quotes aren't needed because it is being rendered as multi-line YAML

shaidar:
👍

PR #514, opened 2018-01-09
Added config for additional load balancer for studio live

blarghmatey:
The studio application requires sticky sessions to function properly so we are adding a third load balancer that will manage the studio domain for live application servers.

### DIFF
--- a/salt/orchestrate/aws/mitx_elb.sls
+++ b/salt/orchestrate/aws/mitx_elb.sls
@@ -43,9 +43,11 @@ create_elb_for_edx_{{ purpose_name }}:
           timeout: 300
     - cnames:
         {% for domain_key, domain in purpose.domains.items()  %}
+        {% if not (edx_type == 'live' and domain_key == 'cms') %}
         - name: {{ domain }}.
           zone: mitx.mit.edu.
           ttl: 60
+        {% endif %}
         {% endfor %}
     - health_check:
         target: 'HTTPS:443/heartbeat'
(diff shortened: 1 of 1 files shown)


salt/orchestrate/aws/mitx_elb.sls
```diff
@@ -43,9 +43,11 @@ create_elb_for_edx_{{ purpose_name }}:
           timeout: 300
     - cnames:
         {% for domain_key, domain in purpose.domains.items()  %}
+        {% if not (edx_type == 'live' and domain_key == 'studio') %}
```
shaidar: Isn't `domain_key` in this case `cms` and not `studio`?

salt/orchestrate/aws/mitx_elb.sls
```diff
@@ -71,4 +73,64 @@ register_edx_{{ purpose_name }}_nodes_with_elb:
         {% endfor %}
     - require:
         - boto_elb: create_elb_for_edx_{{ purpose_name }}
+
+{% if edx_type == 'live' %}
+{% set elb_name = 'edx-studio-live-{env}'.format(
+   purpose=purpose_name, env=ENVIRONMENT)[:32].strip('-') %}
+create_elb_for_edx_studio_live:
```
shaidar: Wouldn't it be better if this was templatized similar to the previous call?

salt/orchestrate/aws/mitx_elb.sls
```diff
@@ -71,4 +73,64 @@ register_edx_{{ purpose_name }}_nodes_with_elb:
         {% endfor %}
     - require:
         - boto_elb: create_elb_for_edx_{{ purpose_name }}
+
+{% if edx_type == 'live' %}
+{% set elb_name = 'edx-studio-live-{env}'.format(
```
shaidar: Wouldn't it be better if this was templatized similar to the previous elb_name?

salt/orchestrate/aws/mitx_elb.sls
```diff
@@ -71,4 +73,64 @@ register_edx_{{ purpose_name }}_nodes_with_elb:
...
+   purpose=purpose_name, env=ENVIRONMENT)[:32].strip('-') %}
```
shaidar: Doesn't appear that `purpose` is needed if this stays the same

salt/orchestrate/aws/mitx_elb.sls
blarghmatey: We can't use the same template value because it would conflict with the other instance. In this case the purpose is studio-live but we don't want to introduce that concept everywhere.

salt/orchestrate/aws/mitx_elb.sls
blarghmatey: Yes

salt/orchestrate/aws/mitx_elb.sls
blarghmatey: Same as above

shaidar:
👍

PR #515, opened 2018-01-09
Added templating of elb name and ids

blarghmatey:
In order to support multiple purposes per env I added templating of the elb name and state ids to use the purpose prefix for the studio ELB


shaidar:
👍

PR #528, opened 2018-01-22
Reorganized devstack pillar files

shaidar:
Reorganized devstack pillar files now that all the other pillar data has been moved in to this repo.


pillar/top.sls
```diff
@@ -149,8 +149,8 @@ base:
...
     - match: compound
     - devstack
-    - consul_devstack
-    - mysql_devstack
-    - rabbitmq_devstack
-    - elasticsearch_devstack
-    - mongodb_devstack
+    - consul.devstack
+    - db.mongodb_devstack
+    - db.mysql_devstack
```
blarghmatey: I would just consider moving the mongo and mysql pillar files under directories for their respective services and factoring out any common values for Mongo into an `init.sls` file.

pillar/top.sls
```diff
@@ -156,8 +156,8 @@ base:
...
```
blarghmatey: The references for mongo and mysql need to be updated to point to their new paths.

blarghmatey:
:+1:

PR #537, opened 2018-02-01
Migrated edx configs to default pillar

blarghmatey:
There were some remaining settings in `next-residential` which had not been moved to the default ansible vars config.


shaidar:
👍

PR #546, opened 2018-02-13
Changed query folder to reside on different volume

shaidar:
Changed query folder to reside on different volume to avoid disk space issues.


blarghmatey:
:-1:

PR #551, opened 2018-02-27
Updated reddit ELB to be generic across apps

blarghmatey:


salt/environment_settings.yml
```diff
@@ -261,7 +261,9 @@ environments:
         domain: video.odl.mit.edu
         business_unit: odl-video-service
       reddit:
+        domain: discussions-reddit-rc-apps.odl.mit.edu
```
shaidar: Shouldn't this be `prod` instead of `rc`?

shaidar:
👍

PR #552, opened 2018-02-28
Added Gunicorn config to work around memory leak and promoted unicode usernames

blarghmatey:
The LMS app is running into issues with a memory leak that can be worked around by causing Gunicorn to restart its
workers after every N requests. To start with I am setting N=1000 and we can tune from there.

Moved the unicode usernames setting from the QA config to be universal across environments.


pillar/edx/ansible_vars.sls
```diff
@@ -370,6 +371,7 @@ edx:
     EDXAPP_GRADE_STORAGE_TYPE: S3
     EDXAPP_JWT_SECRET_KEY: {{ salt.vault.read('secret-{business_unit}/{env}/edxapp-jwt-secret-key'.format(env=environment, business_unit=business_unit)).data.value }}
     EDXAPP_LMS_BASE: "{{ LMS_DOMAIN }}"
+    EDXAPP_LMS_MAX_REQ: 1000
```
pdpinch: What does this do exactly? Restart gunicorn after 1000 requests?

pillar/edx/ansible_vars.sls
```diff
@@ -442,6 +444,7 @@ edx:
         ENABLE_S3_GRADE_DOWNLOADS: true
         ENABLE_SHOPPING_CART: true
         ENABLE_SYSADMIN_DASHBOARD: true
+        ENABLE_UNICODE_USERNAME: true
```
pdpinch: :+1:

pillar/edx/ansible_vars.sls
blarghmatey: Yes, it causes the workers to restart after 1000 requests so that they can free up the memory that they are consuming, thereby working around the memory leak. Because it's the workers restarting and not the entire process it avoids the issue of a user hitting a 502 error from not having any backend listeners available.

PR #556, opened 2018-03-06
Changed celery worker concurrency numbers

shaidar:
Changed celery worker concurrency numbers since after looking at the load on the worker instances, it appears that we can run more tasks on them and thus speed things up a bit

#### What's this PR do?
It should help speed things up as far as task processing since the workers would be able to handle more tasks concurrently.


blarghmatey:
:+1:

PR #557, opened 2018-03-08
Updated pillar data to match current bucket naming

blarghmatey:
The current production buckets don't have any suffix so I updated the rendering to match that for the production environment.


shaidar:
👍

PR #559, opened 2018-03-09
Added the _search route to the allowed paths

blarghmatey:
N/A

#### What's this PR do?
Fixes the `scroll` api calls in micromasters

Deploy it to the ES cluster and verify micromasters functionality

Since the `scroll` API requires access to the `_search` path I added it to the location regex for proxying requests to ES from Nginx.

### DIFF
--- a/pillar/nginx/micromasters_es.sls
+++ b/pillar/nginx/micromasters_es.sls
@@ -29,6 +29,10 @@ nginx:
                     - proxy_pass: http://127.0.0.1:9200$request_uri
                     - proxy_set_header: 'X-Forwarded-For $proxy_add_x_forwarded_for'
                     - proxy_pass_header: 'X-Api-Key'
+                - location /_search/scroll:
+                    - proxy_pass: http://127.0.0.1:9200$request_uri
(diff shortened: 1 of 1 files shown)


pillar/nginx/micromasters_es.sls
```diff
@@ -25,7 +25,7 @@ nginx:
                 - listen:
                     - '[::]:443'
                     - ssl
-                - location ~ ^/(_alias|_aliases|micromasters|_refresh|_mapping):
+                - location ~ ^/(_alias|_aliases|micromasters|_refresh|_mapping|_search):
```
noisecapella: Can we limit this to the scroll API, at `_search/scroll`?

blarghmatey: My understanding from the docs is that the scroll API uses a query parameter and not an actual route, but I could very well be wrong about that.

noisecapella: I mean, instead of `_search` we should do `_search/scroll` since all other `_search` calls use the index in the URL

noisecapella:
I also noticed that the error occurred because the request was `DELETE /_search/scroll`. Do we need to do anything to allow `DELETE` methods?

PR #563, opened 2018-03-27
Moved etl pillar to analytics instance

shaidar:
#### What's this PR do?
Moved etl pillar data to analytics instance since that is where it needs to be deployed to


blarghmatey:
:+1:

PR #565, opened 2018-03-27
Moved notification settings to init.sls pillar

blarghmatey:
In order to prevent duplication and ensure that all notifications go to slack I created an init.sls with the slack notification settings.

### DIFF
--- a/pillar/monit/init.sls
+++ b/pillar/monit/init.sls
@@ -1,0 +1,5 @@
+{% set slack_webhook_url_devops = salt.vault.read('secret-operations/global/slack/slack_webhook_url').data.value %}
+
+monit_app:
+  notification: 'slack'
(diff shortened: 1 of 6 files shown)


shaidar:
I would assume that this would cause a problem [here](https://github.com/mitodl/monit-formula/blob/master/monit/config.sls#L32). The reason I had each alert specifying the type of notification is to keep the formula generic and allow someone to add whatever alert might be needed to that specific rule.

blarghmatey:
For the cases where we don't want to use slack then we would add a `notification` attribute to the particular pillar file where it needs to be overridden. Otherwise, if that needs to be a per-module attribute then we should update the pillar files and states accordingly.

shaidar:
👍

PR #566, opened 2018-03-27
Added techtv domain to server name list

blarghmatey:
Adding the techtv domain to the nginx config allows the site to respond to requests that are using that domain name.

### DIFF
--- a/pillar/nginx/odlvideo.sls
+++ b/pillar/nginx/odlvideo.sls
@@ -36,7 +36,9 @@ nginx:
           enabled: True
           config:
             - server:
-                - server_name: {{ server_domain_name }}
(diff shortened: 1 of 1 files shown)


shaidar:
Should we not add this to [environment_settings.yml](https://github.com/mitodl/salt-ops/blob/master/salt/environment_settings.yml) and iterate over that list?

blarghmatey:
I started to work on that, but the issue comes up with handling the cnames in the ELB since we don't control the zone that the subdomain is attached to (namely `mit.edu`). In the interest of not blocking testing I think we should move forward with hardcoding it in the nginx pillar for now.

shaidar:
👍

PR #571, opened 2018-04-02
DRYing up AWS service discovery in Consul configs

blarghmatey:
To simplify management of AWS services in Consul configs I moved the settings to the `server.sls` pillar file.


pillar/consul/mitx.sls
```diff
@@ -19,22 +19,4 @@ consul:
           check:
             tcp: '{{ mysql_endpoint }}'
             interval: 10s
-        {% for cache_config in env_data.get('backends', {}).get('elasticache', []) %}
```
shaidar: Why not remove previous lines starting from line13 where the if statement is?

blarghmatey: Because we currently rely on the DNS name being `mysql.service.consul` and if I remove that then it will break that resolution.

shaidar:
👍

Issue #527, opened 2018-01-22
Migrate the micromasters edX sandbox to use RDS and run in the QA environment


blarghmatey:
Dump the current contents of the micromasters sandbox database and load it into the QA RDS instance, then deploy a new edX host running the current master branch for use as the new MM sandbox.

shaidar:
Sandbox instance based on hawthorn beta has been deployed and needed information communicated to the team.

PR #574, opened 2018-04-04
Build an edx sandbox environment used for testing and demos

shaidar:
[Issue#527](https://github.com/mitodl/salt-ops/issues/527)

#### What's this PR do?
- Sets up two AWS instances, edxapp and edx-worker, running off of the edx master branch.
- Uses local accounts for authentication as opposed to relying on CAS.
- Creates a few local accounts.

### DIFF
--- a/pillar/edx/ansible_vars.sls
+++ b/pillar/edx/ansible_vars.sls
@@ -73,23 +73,17 @@
         business_unit=business_unit,
         env=environment,
         purpose=purpose)) %}
-{% set remote_gradebook = salt.vault.read(
-    'secret-{business_unit}/{env}/remote_gradebook'.format(
-        business_unit=business_unit, env=environment)) %}
 {# END VAULT DATA LOOKUPS #}
 
 {# Begin Duplicated Variables #}
 {# multivariate #}
 {% set CMS_DOMAIN = purpose_data.domains.cms %}
-{% set EDXAPP_CMS_ISSUER = "https://{}/oauth2".format(CMS_DOMAIN) %}
+
 {% set COMMENTS_SERVICE_KEY = salt.vault.read('secret-residential/global/forum-api-key').data.value %} # TODO: randomly generate? (tmacey 2017/03/16)
-{% set DEFAULT_FEEDBACK_EMAIL = 'mitx-support@mit.edu' %}
-{% set DEFAULT_FROM_EMAIL = 'mitx-support@mit.edu' %}
 {# multivariate, needs to be different for Professional Education, sandbox, etc #}
 {% set GIT_REPO_DIR = edx.edxapp_git_repo_dir %}
 {# multivariate #}
 {% set LMS_DOMAIN = purpose_data.domains.lms %}
-{% set EDXAPP_LMS_ISSUER = "https://{}/oauth2".format(LMS_DOMAIN) %}
 {% set MONGODB_HOST = 'mongodb-master.service.consul' %}
 {% set MONGODB_MODULESTORE_ENGINE =  'xmodule.modulestore.mongo.MongoModuleStore' %}
 {% set MONGODB_REPLICASET = salt.pillar.get('mongodb:replset_name', 'rs0') %}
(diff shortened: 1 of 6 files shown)


pdpinch:
Will this close #527 when it merges?

pillar/edx/ansible_vars.sls
```diff
@@ -73,23 +73,17 @@
...
         purpose=purpose)) %}
-{% set remote_gradebook = salt.vault.read(
-    'secret-{business_unit}/{env}/remote_gradebook'.format(
-        business_unit=business_unit, env=environment)) %}
 {# END VAULT DATA LOOKUPS #}
 
 {# Begin Duplicated Variables #}
 {# multivariate #}
 {% set CMS_DOMAIN = purpose_data.domains.cms %}
-{% set EDXAPP_CMS_ISSUER = "https://{}/oauth2".format(CMS_DOMAIN) %}
```
blarghmatey: This should stay in the base pillar config since it is actually more useful for sandboxes than residential deployments.

pillar/edx/ansible_vars.sls
```diff
@@ -73,23 +73,17 @@
...
+
 {% set COMMENTS_SERVICE_KEY = salt.vault.read('secret-residential/global/forum-api-key').data.value %} # TODO: randomly generate? (tmacey 2017/03/16)
-{% set DEFAULT_FEEDBACK_EMAIL = 'mitx-support@mit.edu' %}
-{% set DEFAULT_FROM_EMAIL = 'mitx-support@mit.edu' %}
 {# multivariate, needs to be different for Professional Education, sandbox, etc #}
 {% set GIT_REPO_DIR = edx.edxapp_git_repo_dir %}
 {# multivariate #}
 {% set LMS_DOMAIN = purpose_data.domains.lms %}
-{% set EDXAPP_LMS_ISSUER = "https://{}/oauth2".format(LMS_DOMAIN) %}
```
blarghmatey: This should stay in the base pillar config since it is actually more useful for sandboxes than residential deployments.

pillar/edx/ansible_vars.sls
```diff
@@ -453,11 +335,6 @@ edx:
         - courseware.student_field_overrides.IndividualStudentOverrideProvider
       GIT_IMPORT_STATIC: false
       LOGGING_ENV: lms-{{ edx.edxapp_log_env_suffix}}
-      OAUTH_OIDC_ISSUER: "{{ EDXAPP_LMS_ISSUER }}"
```
blarghmatey: This should be kept in the base pillar so that OAUTH can be used for sandbox deployments

pillar/edx/ansible_vars.sls
```diff
@@ -472,9 +349,7 @@ edx:
         ENABLE_GIT_AUTO_EXPORT: true
         ENABLE_SQL_TRACKING_LOGS: true
         SEGMENT_IO: false
-        STAFF_EMAIL: mitx-support@mit.edu
       LOGGING_ENV: cms-{{ edx.edxapp_log_env_suffix }}
-      OAUTH_OIDC_ISSUER: "{{ EDXAPP_CMS_ISSUER }}"
```
blarghmatey: This should remain in the base config so that sandbox deployments can use OAUTH

salt/environment_settings.yml
```diff
@@ -141,19 +149,22 @@ environments:
           edx-worker:
             number: 1
             type: t2.medium
-      sandbox:
+      sandbox-residential-live:
```
blarghmatey: What's the reason for changing the purpose name?

salt/environment_settings.yml
```diff
@@ -141,19 +149,22 @@ environments:
...
         business_unit: operations
         domains:
-          cms: studio-micromasters.d.mitx.mit.edu
-          lms: micromasters.d.mitx.mit.edu
-          preview: preview-micromasters.d.mitx.mit.edu
+          cms: studio-mitx-qa-sandbox.mitx.mit.edu
+          lms: mitx-qa-sandbox.mitx.mit.edu
+          preview: preview-mitx-qa-sandbox.mitx.mit.edu
+          gitreload: gr-qa-sandbox.mitx.mit.edu
```
blarghmatey: There's no need for a gitreload URL on the sandbox box because we won't be running that service there.

salt/environment_settings.yml
```diff
@@ -141,19 +149,22 @@ environments:
...
         versions:
-          edxapp: master
-          forum: master
+          <<: *sandbox_residential_versions
         instances:
           edx:
             number: 1
             type: t2.medium
+          edx-worker:
```
blarghmatey: There's no need to specify the edx worker instance count because the worker will be colocated with the edxapp on the same box.

salt/top.sls
```diff
@@ -157,6 +154,9 @@ base:
     - fluentd
     - fluentd.plugins
     - fluentd.config
+  'purpose:sandbox':
+    - match: grain
+    - edx.django_user
```
blarghmatey: This will need additional states defined to ensure that the ansible playbooks get deployed, Consul gets set up, etc.

shaidar: I would assume consul is getting triggered by the environment grain and does not need to be setup separately in this case, no?

salt/environment_settings.yml
shaidar: Figured it would be consistent with the overall naming of residential instances and thought that we have cases were we are filtering on `type` and needed to have that specified otherwise it would mean more code refactoring.

salt/top.sls
```diff
@@ -156,6 +156,12 @@ base:
     - fluentd.config
   'purpose:sandbox':
     - match: grain
+    - edx.prod
+    - edx.run_ansible
```
blarghmatey: edx.prod automatically runs edx.run_ansible

salt/top.sls
```diff
@@ -156,6 +156,12 @@ base:
...
+    - edx.patch_nginx
+    - edx.edxapp_global_pre_commit
```
blarghmatey: We don't need this for sandbox as it is only used for managing a hack for git exports of courses.

salt/top.sls
```diff
@@ -156,6 +156,12 @@ base:
...
+    - edx.etc_hosts
```
blarghmatey: We don't need this for sandbox as it is a hack for fixing issues with xqueue latency

salt/top.sls
blarghmatey: Most likely

salt/environment_settings.yml
blarghmatey: I think that most of the cases around type are for things like the orchestration for deploy and ELB. For the sandbox I think we're going to just use the apps.deploy orchestrate command.

shaidar: Wouldn't we need to add an entry in pillar for sandbox if we revert the name back to sandbox from `sandbox-residential-live`?
https://github.com/mitodl/salt-ops/blob/master/pillar/top.sls#L159

PR #576, opened 2018-04-05
Added edx codename as grain and conditional for static assets

blarghmatey:
N/A

#### What's this PR do?
The path for the script(s) to recompile static assets on edX has changed in master/Hawthorn so I added the codename grain to deployed instances for future reference, and a conditional for specifying which command to use for building assets.


shaidar:
👍

PR #577, opened 2018-04-06
Added Nathan to Kibana &moved mailgun alerts to their own channel

shaidar:
#### What's this PR do?
Added Nathan to Kibana &moved mailgun alerts to their own channel. Mailgun alerts can get a bit noisy and render the slack channel somewhat useless and tough to communicate in. This uses a dedicated slack channel for mailgun elastalerts and folks interested in viewing those can subscribe to that channel.


blarghmatey:
:+1:

PR #578, opened 2018-04-06
Added event triggers to backup/restore states and slack messages

blarghmatey:
In order to get more visibility into the overall execution of our backup/restore jobs I added event triggers to the state runs for each script and set up a slack message to post the results of those events


shaidar:
👍

PR #585, opened 2018-04-10
Added missing cloud profile for sandbox instance

shaidar:


salt/orchestrate/aws/cloud_profiles/sandbox.conf
```diff
@@ -0,0 +1,21 @@
+# -*- mode: yaml; coding: utf-8; -*-
+sandbox:
+  provider: mitx
+  size: t2.medium
+  image: {{ salt.sdb.get('sdb://consul/xenial_ami_id') }}
+  ssh_username: ubuntu
+  ssh_interface: private_ips
+  block_device_mappings:
+    - DeviceName: /dev/sda1
+      Ebs.VolumeSize: 8
```
blarghmatey: Might want to increase the disk size for this instance since it will be containing some course data

shaidar:
Changed disk size to 40 instead of 20 just in case they decide to add multiple courses and we flip to DEBUG for log files.

blarghmatey:
:+1:

PR #586, opened 2018-04-11
Removed references to shared edx pillar

shaidar:
Removed references to shared pillar as we were receiving maximum recursion rendering errors and refactored code to remove references to those shared files.


blarghmatey:
:+1:

PR #588, opened 2018-04-12
Added newly generated OVS cert that includes ci/rc and techtv as SANs

shaidar:
Created cert file that includes interm cert and also verified that returned cert has the SAN entries listed for the DNS entires we requested. Added key and cert to vault and verified that they are formatted correctly.

### DIFF
--- a/pillar/nginx/odlvideo.sls
+++ b/pillar/nginx/odlvideo.sls
@@ -3,7 +3,7 @@
 {% set ENVIRONMENT = salt.grains.get('environment', 'rc-apps') %}
 {% set env_data = env_settings.environments[ENVIRONMENT] %}
 {% set server_domain_name = env_data.purposes['odl-video-service'].domain %}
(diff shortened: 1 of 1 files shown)


pillar/nginx/odlvideo.sls
```diff
@@ -3,7 +3,7 @@
 {% set ENVIRONMENT = salt.grains.get('environment', 'rc-apps') %}
 {% set env_data = env_settings.environments[ENVIRONMENT] %}
 {% set server_domain_name = env_data.purposes['odl-video-service'].domain %}
-{% set odl_wildcard = salt.vault.read('secret-operations/global/odl_wildcard_cert') %}
+{% set odl_wildcard = salt.vault.read('secret-operations/global/ovs_techtv_cert') %}
```
blarghmatey: Should we change the variable name?

shaidar: It's pretty descriptive, no?

blarghmatey:
:+1:

PR #591, opened 2018-04-13
Moved source repos from ansible_vars files into environment_settings

shaidar:
Git branches and source repos were interspersed between environment_settings and ansible_vars files which can lead to some confusion as to what repo is being deployed along with the related branch. I moved all the source repos into environment_settings since that is where we define the git branches.
I also removed settings from the generic ansible_vars (such as xqueue and forum) to the residential_ansible_vars.


blarghmatey:
:+1:

PR #593, opened 2018-04-23
Moved memcached configuration to residential_ansible_vars

shaidar:
Moved memcached configuration to residential_ansible_vars and added a localhost config to sandbox.


blarghmatey:
:+1:

PR #594, opened 2018-04-24
Make migrations if rapid-response xblock is installed

shaidar:


salt/edx/migration.sls
```diff
@@ -8,3 +8,11 @@ run_edxapp_{{ migration }}_migrations:
...
     - runas: edxapp
 {% endfor %}
+
+{% if salt['file.directory_exists']('/edx/app/edxapp/venvs/edxapp/rapid_response') %}
+run_rapid_response_xblock_migration:
+  cmd.run:
+    - name: '{{ edxapp_bin }}python manage.py lms makemigrations rapid_response_xblock --settings=aws'
+    - cwd: /edx/app/edxapp/edx-platform
+    - runas: edxapp
+{% endif %}
```
blarghmatey: This would need to be executed before the migrations get run, otherwise nothing will happen.

blarghmatey:
:+1:

PR #599, opened 2018-05-01
Made changes to include multiple OVS domain names & changed ELB state

shaidar:
Made changes to include multiple OVS domain names & changed ELB state to only create an ELB if the domain name matches a zone that we manage.
Also modified vault path to OVS web cert since we now have separate certs for `ci/rc` and `prod` that should also include all the different domain names.

### DIFF
--- a/pillar/nginx/odlvideo.sls
+++ b/pillar/nginx/odlvideo.sls
@@ -2,7 +2,7 @@
 {% set env_settings = salt.cp.get_file_str("salt://environment_settings.yml")|load_yaml %}
 {% set ENVIRONMENT = salt.grains.get('environment', 'rc-apps') %}
 {% set env_data = env_settings.environments[ENVIRONMENT] %}
-{% set server_domain_names = env_data.purposes['odl-video-service'].domain %}
+{% set server_domain_names = env_data.purposes['odl-video-service'].domains %}
 {% set ovs_web_cert = salt.vault.read('secret-odl-video/{env}/ovs_web_cert'.format(env=ENVIRONMENT)) %}
 {% set ovs_login_path = 'collections' %}
 
--- a/pillar/nginx/odlvideo.sls
+++ b/pillar/nginx/odlvideo.sls
@@ -2,7 +2,7 @@
 {% set env_settings = salt.cp.get_file_str("salt://environment_settings.yml")|load_yaml %}
 {% set ENVIRONMENT = salt.grains.get('environment', 'rc-apps') %}
 {% set env_data = env_settings.environments[ENVIRONMENT] %}
-{% set server_domain_name = env_data.purposes['odl-video-service'].domain %}
+{% set server_domain_names = env_data.purposes['odl-video-service'].domain %}
 {% set ovs_web_cert = salt.vault.read('secret-odl-video/{env}/ovs_web_cert'.format(env=ENVIRONMENT)) %}
 {% set ovs_login_path = 'collections' %}
 
@@ -36,8 +36,7 @@ nginx:
           enabled: True
           config:
             - server:
-                - server_name:
-                    {{ server_domain_name }}
+                - server_name: {{ server_domain_names }}
                 - listen:
                     - 80
                 - listen:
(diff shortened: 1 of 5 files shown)


salt/environment_settings.yml
```diff
@@ -314,7 +314,10 @@ environments:
     purposes:
       odl-video-service:
         num_instances: 3
-        domain: video.odl.mit.edu
+        domain:
```
blarghmatey: We should change this to be plurarl

pillar/nginx/odlvideo.sls
```diff
@@ -37,8 +37,7 @@ nginx:
           config:
             - server:
                 - server_name:
-                    - {{ server_domain_name }}
-                    - techtv.mit.edu
+                    {{ server_domain_name }}
```
blarghmatey: Pluralize this as well. Also, since it is now a list it should be moved up a line
```
config:
  - server:
    - server_name: {{ server_domain_names }}
```

pillar/nginx/odlvideo.sls
```diff
@@ -47,8 +46,7 @@ nginx:
                     - return: 301 https://$host$request_uri
             - server:
                 - server_name:
-                    - {{ server_domain_name }}
-                    - techtv.mit.edu
+                    {{ server_domain_name }}
```
blarghmatey: Same as above

salt/orchestrate/aws/app_elb.sls
```diff
@@ -20,6 +22,8 @@
 {% do instance_ids.append(grains['instance-id']) %}
 {% endfor %}
 
+{% for zone_name, domain in zip(zone_names, domains) %}
+{% if zone_name.strip('.') in domain %}
```
blarghmatey: This loop should be under `cnames`, not wrapping the entire ELB block

salt/environment_settings.yml
shaidar: I had that as plural initially, but changed it back since it means we would need to change all occurrences of it in that file so that when we call it from elsewhere, it would be one or the other. Right?

salt/orchestrate/aws/app_elb.sls
```diff
@@ -40,18 +38,20 @@ create_elb_for_{{ app_name }}_{{ ENVIRONMENT }}:
           enabled: True
           timeout: 300
     - cnames:
+    {% for zone_name, domain in zip(zone_names, domains) %}
+    {% if zone_name.strip('.') in domain %}
         - name: {{ purpose.domain }}
```
blarghmatey: This should be using the `domain` variable from the `zip` function call, rather than `purpose.domain`

salt/orchestrate/aws/app_elb.sls
```diff
@@ -40,18 +38,20 @@ create_elb_for_{{ app_name }}_{{ ENVIRONMENT }}:
           enabled: True
           timeout: 300
     - cnames:
+    {% for zone_name, domain in zip(zone_names, domains) %}
```
blarghmatey: I think we might need to do this as a nested loop since zip will stop as soon as one of the iterables is exhausted, and it won't exhaustively exercise the possible combinations.

salt/environment_settings.yml
blarghmatey: I think that it makes sense to just commit to it always being pluarl

salt/orchestrate/aws/app_elb.sls
```diff
@@ -10,7 +10,7 @@
         name=env_data.vpc_name).vpcs[0].id
     ).subnets|map(attribute='id')|list %}
 {% set zone_names = salt.boto_route53.describe_hosted_zones|map(attribute='Name')|list %}
-{% set domains = [purpose.domain] %}
+{% set domains = [purpose.domains] %}
```
blarghmatey: The value of `purpose.domains` is already a list so wrapping it in `[]` will cause it to be a nested list.

blarghmatey:
:+1:

PR #602, opened 2018-05-04
Redirect rule to throw a 404 when collection is not available

shaidar:
[Issue#604](https://github.com/mitodl/odl-video-service/issues/604)

#### What's this PR do?
Added redirect rule to throw a 404 error when collection is not available

Applied and tested it on `RC`


shaidar:
We ended up going another route and setting up a `login` endpoint. Closing as this is no longer necessary.

PR #606, opened 2018-05-18
Increased number of xqwatchers to build an new instance and provide a…

shaidar:
Increased number of xqwatchers to build an new instance and provide access to 686 course team.


shaidar:
Took another route by refactoring xqwatcher code to split up the graders into separate state files and deploying the 686 graders to its own instance.

PR #607, opened 2018-05-18
Split xqwatcher graders into separate files for granular minion targe…

shaidar:
#### What's this PR do?
Split xqwatcher graders into separate files for granular minion targeting since we would like to run the 686 grader on its own minion.

### DIFF
--- a/pillar/edx/xqwatcher.sls
+++ b/pillar/edx/xqwatcher.sls
@@ -1,37 +1,11 @@
-#!jinja|yaml|gpg
+#!jinja|yaml
+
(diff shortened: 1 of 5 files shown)


pillar/edx/xqwatcher.sls
```diff
@@ -1,38 +1,12 @@
-#!jinja|yaml|gpg
+#!jinja|yaml
```
blarghmatey: We can just delete this line

blarghmatey:
:+1:

PR #608, opened 2018-05-18
Added entries to top files to target lightsail instance for xqwatcher

shaidar:
Added entries to top files to target lightsail instance for xqwatcher 686 grader


blarghmatey:
:+1:

PR #615, opened 2018-05-23
Moved elastalert configs to separate file, fixed beacon syntax for kibana

blarghmatey:
#### What's this PR do?
Moves elastalert configs to separate file and updates syntax for service beacon


shaidar:
👍

PR #617, opened 2018-05-24
Updated orchestrate states under operations to use new var pattern

blarghmatey:
#### What's this PR do?
Updates the pattern for orchestrate states in the operations directory to use the environments.yml file to pull in various facts about the network environment, rather than relying on specifying multiple environment variables at the command line.


shaidar:
👍

PR #618, opened 2018-05-24
Refactored to use cached_read in pillar for connection strings

blarghmatey:
#### What's this PR do?
Updated pillar rendering to use new `cached_read` execution module function to leverage the lease caching mechanism and elminate churn in dynamic credentials


shaidar:
👍

PR #619, opened 2018-05-31
Moved ansible vars files to edx subdirectory

blarghmatey:
In order to make it easier to identify which files are primarily for setting ansible vars I moved them to a subdirectory of the edx pillar.

This is just an idea and I welcome pushback if it doesn't make sense.


shaidar:
No complaints from me about adding more structure 👍

blarghmatey:
One slight problem that I have with this as it stands currently is that there are still some ansible vars values in other files in the edx pillar directory, so having this folder present would suggest that those should get moved out of their current files and into new files under this folder...

PR #621, opened 2018-06-07
Updated rabbit orchestration to remove aws macro

blarghmatey:
This updates the rabbit orchestration state to match our more recent pattern for specifying instance count at the command line and only modifying newly deployed instances as part of the run.


shaidar:
`VPC_RESOURCE_SUFFIX` is used on line 37, however it is no longer defined

shaidar:
👍

PR #622, opened 2018-06-07
Modified alert so that we also get notified on failures

shaidar:
Modified alert so that we also get notified on failures to login to rabbitmq from datadog

### DIFF
--- a/pillar/elastalert.sls
+++ b/pillar/elastalert.sls
@@ -164,7 +164,7 @@ elasticsearch:
             minutes: 5
           alert:
             - slack
-          alert_text: "<!subteam^S9PK3B39V|devopseng> Rabbitmq AMQPLAIN login refused due to expired vault credentials"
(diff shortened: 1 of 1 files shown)


pillar/elastalert.sls
```diff
@@ -173,7 +173,7 @@ elasticsearch:
             - bool:
                 must:
                   - match:
-                      message: AMQPLAIN login refused
+                      message: invalid credentials
```
blarghmatey: Will this match all cases?

shaidar: The two cases I have comes across in the logs are the following:
`2018-06-07 15:55:46.001 [warning] <0.11298.0> HTTP access denied: user <username>  - invalid credentials`

`AMQPLAIN login refused: user <username>  - invalid credentials`

Based on that, I believe it should match both cases.

blarghmatey:
:+1:

Issue #511, opened 2017-12-26
Hawthorn: missing 503 error page


pdpinch:
I'm not sure if this is a theme issue, or what. But when there's a 503 error, the response is empty.

pdpinch:
@gsidebo do you think this is addressed by https://github.com/mitodl/edx-platform/pull/67 ?

gsidebo:
I'd say most likely not. I'm pretty sure my fix works for 404s and 500s, not 503s. I don't think we even have a custom error page for 503s. I would need some way to reproduce this in order to say for sure

pdpinch:
We can live with this for now, but I'd like to make another attempt for Hawthorn.

pdpinch:
Depending on the nature of the 503 error, it may originate in the load balancer. In that case, edX doesn't even have a chance to provide an error page. 

Unfortunately, it my experience the load balancer returns an empty page (with a 503 header) under these circumstances. It's difficult for a naive user to know what to do.

pdpinch:
People have been asking for Amazon a customizable 503 message (or any message) since 2011. 

https://forums.aws.amazon.com/thread.jspa?threadID=72363&start=100&tstart=0

Instead of customizing the message, can we get monit to set an urgent alert if any of the QA or production stacks are returning 503?

blarghmatey:
@shaidar can you add a check for Monit to run against our LMS instances to alert on 503 errors. Probably deploy the agent with those checks to something in the ops VPC

shaidar:
Added a [PR](https://github.com/mitodl/salt-ops/pull/626/files) to post message to slack when 503 encountered

blarghmatey:
We may finally have an answer for this https://aws.amazon.com/about-aws/whats-new/2018/07/elastic-load-balancing-announces-support-for-redirects-and-fixed-responses-for-application-load-balancer/

PR #626, opened 2018-06-08
Added monit alert to notify slack when LMS page is returning a 503 error

shaidar:
[Issue#511](https://github.com/mitodl/salt-ops/issues/511)

#### What's this PR do?
Monit alert to notify slack when LMS page is returning a 503 error

### DIFF
--- a/pillar/monit/lms_503.sls
+++ b/pillar/monit/lms_503.sls
@@ -1,0 +1,16 @@
+{% set env_settings = salt.cp.get_file_str("salt://environment_settings.yml")|load_yaml %}
+{% set purpose = salt.grains.get('purpose', 'current-residential-live') %}
+{% set environment = salt.grains.get('environment', 'mitx-qa') %}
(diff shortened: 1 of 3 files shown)


blarghmatey:
You'll need to add this to the pillar top file for targeting. Also, is it possible to make this flag the `devopseng` user in the slack message so that we don't miss it?

blarghmatey:
Also, this won't work if we're running it on the edX servers since we're overriding the DNS lookup with the /etc/hosts hack that we set up for xqueue submissions.

shaidar:
I'll change it to have a static value to the LMS DNS entry then so that it runs on an instance in operations like you originally requested. 

On the matter of flagging `devopseng`, I think we need to have a conversation about alerts in general as it feels like we're heading down the dangerous path of too much noise. You had previously mentioned possibly tagging alert by level of severity which sounds like an interesting idea.

blarghmatey:
I think that is a valuable discussion to have. Let's chat about it today after standup.

blarghmatey:
:-1:

PR #629, opened 2018-06-12
Added updated revocation statements for postgres roles

blarghmatey:
We had a number of expired leases that were not cleaned up due to faulty revocation statements. Updating our pillar data to use a known-good approach.

#512 

#### What's this PR do?
Fixes the revocation statements for postgres roles

Modify an existing role with the updated revocation statements and revoke an expired lease


shaidar:
:thumbsup:

PR #630, opened 2018-06-13
Fixed issues with failing pillar refresh on xqwatcher

shaidar:
#### What's this PR do?
The xqwatcher pillars were trying to load values from purpose data that doesn't apply to them so I created a separate xqwatcher purpose to contain their version information.


shaidar:
👍

PR #631, opened 2018-06-15
Redirect residential registration page to login page

shaidar:


salt/edx/files/nginx_static_assets.j2
```diff
@@ -21,6 +21,11 @@
         try_files /staticfiles/$collected /course_static/$collected =404;
     }
 
+    # Redirect registration page to authentication page
+    location ~ ^/register/(.*)$ {
```
blarghmatey: We can just make this `location /register`. That will by default assume that it is the beginning of the route and anything under that will be handled by this location block. Avoiding the regex match will prevent additional compute cycles

salt/edx/files/nginx_static_assets.j2
```diff
@@ -22,7 +22,7 @@
     }
 
     # Redirect registration page to authentication page
-    location ~ ^/register/(.*)$ {
+    location /register/ {
```
blarghmatey: Remove the trailing slash and then :+1:

PR #632, opened 2018-06-15
Added additional port range for internode communication

blarghmatey:
RabbitMQ 3.7.x adds another set of ports for inter-node communication that gets used by commands such as list-queues. Added these ports to the rabbit security groups to resolve timeouts when executing queries.

Also updated the mitx vpc orchestration to use env settings data in the new pattern.

N/A

#### What's this PR do?
Adds additional ports to the rabbitmq security groups for inter-node communication in 3.7

Apply the security group changes to an existing security group and verify proper operation of the `rabbitmq.list_queues_vhost` command.

### DIFF
--- a/pillar/rabbitmq/init.sls
+++ b/pillar/rabbitmq/init.sls
@@ -7,9 +7,6 @@ rabbitmq:
   overrides:
     version: '3.7.4-1'
     erlang_version: '1:20.1'
-  environment:
-    RABBITMQ_CTL_DIST_PORT_MIN: 35672
-    RABBITMQ_CTL_DIST_PORT_MAX: 35672
   configuration:
     cluster_partition_handling: autoheal
     cluster_partition_handling.pause_if_all_down.recover: autoheal
--- a/salt/orchestrate/aws/mitx.sls
+++ b/salt/orchestrate/aws/mitx.sls
@@ -3,16 +3,12 @@
 # Make sure that instance profiles exist for node types so that
 # they can be granted access via permissions attached to those
 # profiles because it's easier than managing IAM keys
-{% set VPC_NAME = salt.environ.get('VPC_NAME', 'MITx QA') %}
-{% set VPC_RESOURCE_SUFFIX = salt.environ.get(
-    'VPC_RESOURCE_SUFFIX',
-    VPC_NAME.lower().replace(' ', '-')) %}
-{% set ENVIRONMENT = salt.environ.get(
-    'ENVIRONMENT',
-    VPC_NAME.lower().replace(' ', '-')) %}
-{% set BUSINESS_UNIT = salt.environ.get('BUSINESS_UNIT', 'residential') %}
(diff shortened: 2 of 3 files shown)


salt/orchestrate/aws/mitx.sls
```diff
@@ -40,88 +34,88 @@ create_{{ VPC_RESOURCE_SUFFIX_UNDERSCORE }}_vpc:
...
+    - conn_name: {{ ENVIRONMENT }}-operations-peer
 
-create_{{ VPC_RESOURCE_SUFFIX_UNDERSCORE }}_routing_table:
+create_{{ ENVIRONMENT }}_routing_table:
   boto_vpc.route_table_present:
-    - name: {{ VPC_RESOURCE_SUFFIX }}-route_table
+    - name: {{ ENVIRONMENT }}-route_table
     - vpc_name: {{ VPC_NAME }}
     - subnet_names:
         - public1-{{ VPC_RESOURCE_SUFFIX }}
```
shaidar: Should this be `ENVIRONMENT`?

salt/orchestrate/aws/security_groups.sls
```diff
@@ -163,6 +163,10 @@ create_rabbitmq_security_group:
           from_port: 25672
           to_port: 25672
           source_group_name: rabbitmq-{{ ENVIRONMENT }}
+        - ip_protocol: tcp
+          from_port: 35672
+          to_port: 35682
```
shaidar: Shouldn't this be `35672` since we only made the cli use a single ports as opposed to the range?

blarghmatey: No, because by default it uses a range and I removed the min/max from the env file. I had been experimenting with that before I discovered that it was a network issue.

shaidar: But in `pillar/rabbitmq/init.sls` used the port_min and port_max as 35672. And sounds like in their docs, u can restrict it to one port if u don’t have mutliple cli’s making queries, no?

blarghmatey: You can, but it's not required as long as the internode communication is working. Also, we often run the same command against all of the instances in a cluster so restricting the port range at the config and network layer may cause contention in those situations.

pillar/rabbitmq/init.sls
```diff
@@ -7,9 +7,6 @@ rabbitmq:
   overrides:
     version: '3.7.4-1'
     erlang_version: '1:20.1'
-  environment:
-    RABBITMQ_CTL_DIST_PORT_MIN: 35672
-    RABBITMQ_CTL_DIST_PORT_MAX: 35672
```
shaidar: So should this be `35682`?

blarghmatey: This is removing those lines so that it will just use the defaults.

PR #634, opened 2018-06-18
Added micromasters to list of envs for datadog

blarghmatey:
The MM ES cluster didn't have the datadog API key applied due to an oversight in top file targeting.


shaidar:
👍

PR #641, opened 2018-06-28
Fixed celery command formatting to avoid uwsgi comma parsing

blarghmatey:
The scheduled queries were not being executed because of the fact that uWSGI was parsing the comma as part of the queues list for the celery command. Combined both celery statements onto one line and added escaping of commas so that all queues will be processed properly.

closes odl-analytics/bi.odl.mit.edu#11

#### What's this PR do?
Fixes the scheduled query setup for Celery in Redash

Apply the changes in the uwsgi config and ensure scheduled queries are being run

### DIFF
--- a/pillar/apps/redash.sls
+++ b/pillar/apps/redash.sls
@@ -51,6 +51,8 @@ django:
     REDASH_EVENT_REPORTING_WEBHOOKS: https://log-input.odl.mit.edu/redash-webhook/redash/events?token={{ redash_fluentd_webhook_token }}
     REDASH_HOST: https://bi.odl.mit.edu
     REDASH_LOG_LEVEL: INFO
+    REDASH_LOG_PREFIX: REDASH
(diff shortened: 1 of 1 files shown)


pillar/apps/redash.sls
```diff
@@ -111,12 +113,7 @@ uwsgi:
         - env: '%(_)'
         - endfor: ''
         - attach-daemon2: >-
-            cmd=/usr/local/pyenv/versions/{{ python_version }}/bin/celery worker -A redash.worker -B -c2 -Qqueries,celery --pidfile /opt/{{ app_name }}/celery.pid -Ofair,
```
shaidar: Why are we getting rid of this?

blarghmatey: Because it's unnecessary. I'm just combining all of the queues into the one command.

shaidar:
👍

PR #642, opened 2018-07-02
Added hearbeat endpoint to check for status

shaidar:
#### What's this PR do?
Added hearbeat endpoint to check for status so that monit logs don't appear in the edx tracking logs file.

### DIFF
--- a/pillar/monit/lms_503.sls
+++ b/pillar/monit/lms_503.sls
@@ -7,5 +7,5 @@ monit_app:
         with:
           address: lms.mitx.mit.edu
         if:
-          failed: port 443 protocol https status = 503
+          failed: port 443 protocol https request "/heartbeat" status = 200
           action: exec "/bin/sh -c /usr/local/bin/slack.sh"


pillar/monit/lms_503.sls
```diff
@@ -7,5 +7,5 @@ monit_app:
         with:
           address: lms.mitx.mit.edu
         if:
-          failed: port 443 protocol https status = 503
+          failed: port 443 protocol https request "/heartbeat" status = 200
```
blarghmatey: Does this mean that it will fail if the heartbeat endpoint returns a success?

shaidar: No my understanding of the docs is that if it doesn't return a 200, that would be considered as a failure and would trigger the alert. From the docs:
```
For example to test that a page does not exist (the HTTP server should return 404 in this case):

  if failed
     port 80
     protocol http
     request "/non/existent.php"
     status = 404
  then alert
```

blarghmatey:
:+1:

PR #643, opened 2018-07-02
Added post-mortem doc for outage due to Vault lease expirations

blarghmatey:


shaidar:
Looks good and not a fun read having to relive that day 👍

PR #658, opened 2018-07-13
Updated to delegate security groups to env yaml

blarghmatey:
Some apps will need different security group names so delegating that to env yaml seems easiest.


shaidar:
👍

PR #660, opened 2018-07-17
Moved ga settings to use jinja templating for jsonifying

blarghmatey:
Using the `__vault__` syntax for rendering through vault precludes our ability to modify the resulting value. Moved the retrieval of the data from Vault to use Jinja so that we can pass it through the JSON filter which will fix the syntax issue.


shaidar:
👍

Issue #215, opened 2017-01-18
Review, reconfigure, and document notifications


shaidar:
Once we've centralized all the server logs, we should review and reconfigure some of the alerts and notifications. In some instances, we need to add some alerts to certain services and in others we need to tweak the current settings. 
In addition, we need to ensure that notifications are all funneled through the same system and then distributed to the appropriate channel (Email, Slack, SMS ...).
Last but not least, we need to document the process and possibly review Salt states to ensure that new builds are configured properly going forward.

blarghmatey:
Can this be closed, or do we have remaining work for this ticket?

shaidar:
Some changes were merged and a first pass at creating a [doc](https://github.com/mitodl/salt-ops/blob/master/docs/infrastructure/notifications.md) took place. Closing this issue for now.

PR #661, opened 2018-07-17
Changed alerts/notifications to go to OpsGenie

shaidar:
Part of [Issue#215](https://github.com/mitodl/salt-ops/issues/215)

#### What's this PR do?
- Added SaltStack integration to OpsGenie
- Created state files for alerts and notifications
- Removed posting to slack from reddit_service_low_memory state and just added a notification to reactor

### Notes
I have added policies to OpsGenie to do the following:
- Auto-close anything that is a P4 or lower
- Set the priority on an alert to `P5` if the name contains notification so that anything generated by the saltstack reactor and uses the `post_notification` state would go to the appropriate slack channel - `devops-notifications`


salt/reactors/opsgenie/post_alert.sls
```diff
@@ -0,0 +1,9 @@
+post_alert_to_opsgenie:
+  local.opsgenie.post_data:
+    - tgt: 'roles:master'
+    - tgt_type: grain
+    - kwarg:
+        name: salt.reactor.alert
+        api_key: __vault__::secret-operations/global/opsgenie/opsgenie_saltstack_api>data>value
+        reason: {{ data['data']['reason'] }}
```
blarghmatey: Is it possible to send the entire data object as an additional field?

shaidar: Don't believe so. Think the only think we can maybe do is send it in the `reason` key. The salt stack integration appears to be fairly limited - https://docs.opsgenie.com/docs/saltstack-integration

blarghmatey:
:+1:

PR #662, opened 2018-07-19
Added webapp security group that only allows access from ODL & VPN

shaidar:
#### What's this PR do?
Added webapp security group that only allows access from ODL & VPN. Also assigned this new secgroup to BI.


blarghmatey:
:+1:

Issue #20, opened 2016-03-29
Create PKI/CA infrastructure


blarghmatey:
Having a CA/PKI setup will be useful for various infrastructure projects. Current examples include:
- Salt master key signing for HA and disposable salt master
- SSL key authentication between FluentD forward/aggregate nodes
- SSL key authentication for Kibana (?)

Potentially leverage https://github.com/cloudflare/cfssl

blarghmatey:
This can be achieved more easily once we have a functioning installation of Hashicorp Vault as described in its documentation here: https://www.vaultproject.io/docs/secrets/pki/index.html

pdpinch:
What's the status of this?

blarghmatey:
This will be done after the dogwood deploy.


Regards,

Tobias Macey

Senior DevOps Engineer

MIT Office of Digital Learning

http://odl.mit.edu

On Monday, July 18, 2016 8:16:58 AM EDT Peter Pinch wrote:

What's the status of this?

pdpinch:
Moving this to "New Issues" because I don't think we're ready to work on it?

pdpinch:
I think this is done.

shaidar:
Working on setting up Vault to have a PKI mount and issue signed certs for our infrastructure. I did test run on my local vagrant environment which was successful and then worked on getting the necessary certs to use on the production Vault. As part of that work, ended up upgrading our vault cluster to the latest version.

PR #663, opened 2018-07-19
Setup vault pki backends and roles to use those backends

shaidar:
[Issue#20](https://github.com/mitodl/salt-ops/issues/20)

#### What's this PR do?
Setup vault pki backends and roles to use those backends. Also added new variables to env_settings that specifies what services each environment needs. The roles file needs extra scrutiny as I'm a bit unsure about how the server/client flags will work.

### DIFF
--- a/pillar/vault/roles/pki.sls
+++ b/pillar/vault/roles/pki.sls
@@ -1,5 +1,5 @@
 {% set env_settings = salt.cp.get_file_str("salt://environment_settings.yml")|load_yaml %}
-{% set ttl = '8760h' %}
+{% set ttl = '8760h' %} # ONE_YEAR
 {% set ou = 'Open Learning' %}
 {% set org = 'Massachusetts Institute of Technology' %}
 {% set country = 'US' %}
@@ -12,7 +12,7 @@ vault:
     {% for env in env_settings.environments %}
     {% for app in env.backends.pki %}
     {{ app }}-{{ env }}-pki:
-      backend: pki_int_{{ env }}
+      backend: pki-int-{{ env }}
       {% for type in ['client', 'server'] %}
       name: pki-{{ env }}-{{ app }}-{{ type }}
       {% endfor %}
@@ -20,7 +20,7 @@ vault:
       max_ttl: {{ ttl }}
       allowed_domains:
         - {{ app }}.service.consul
-        - nearest-{{ app }}.service.consul
+        - nearest-{{ app }}.query.consul
         {% if app == 'mongodb' %}
         - {{ app }}-master.service.consul
         {% endif %}
--- a/pillar/vault/roles/pki.sls
+++ b/pillar/vault/roles/pki.sls
@@ -1,0 +1,46 @@
+{% set env_settings = salt.cp.get_file_str("salt://environment_settings.yml")|load_yaml %}
+{% set ttl = '8760h' %}
+{% set ou = 'Open Learning' %}
+{% set org = 'Massachusetts Institute of Technology' %}
+{% set country = 'US' %}
+{% set locality = 'Cambridge' %}
+{% set street_address = '77 Massachusetts Ave' %}
+{% set postal_code = '02139' %}
+
+vault:
+  roles:
+    {% for env in env_settings.environments %}
+    {% for app in env.backends.pki %}
+    {{ app }}-{{ env }}-pki:
+      backend: pki_int_{{ env }}
+      {% for type in ['client', 'server'] %}
+      name: pki-{{ env }}-{{ app }}-{{ type }}
+      {% endfor %}
+      ttl: {{ ttl }}
+      max_ttl: {{ ttl }}
+      allowed_domains:
+        - {{ app }}.service.consul
+        - nearest-{{ app }}.service.consul
+        {% if app == 'mongodb' %}
+        - {{ app }}-master.service.consul
+        {% endif %}
+      {% if type == 'server' %}
+      server_flag: true
+      {% else %}
+      client_flag: true
+      {% endif %}
+      key_type: rsa
+      key_bits: 4096
+      key_usage:
+        - DigitalSignature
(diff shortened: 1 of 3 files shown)


pillar/vault/roles/pki.sls
```diff
@@ -0,0 +1,46 @@
+{% set env_settings = salt.cp.get_file_str("salt://environment_settings.yml")|load_yaml %}
+{% set ttl = '8760h' %}
```
blarghmatey: Rename this variable to be a descriptive constant (e.g. `THIRTY_DAYS, or whatever it actually is)

pillar/vault/roles/pki.sls
```diff
@@ -0,0 +1,46 @@
...
+{% set locality = 'Cambridge' %}
+{% set street_address = '77 Massachusetts Ave' %}
+{% set postal_code = '02139' %}
+
+vault:
+  roles:
+    {% for env in env_settings.environments %}
+    {% for app in env.backends.pki %}
+    {{ app }}-{{ env }}-pki:
+      backend: pki_int_{{ env }}
```
blarghmatey: We are using `-` separated names for our other backends

pillar/vault/roles/pki.sls
```diff
@@ -0,0 +1,46 @@
...
+      {% for type in ['client', 'server'] %}
+      name: pki-{{ env }}-{{ app }}-{{ type }}
+      {% endfor %}
+      ttl: {{ ttl }}
+      max_ttl: {{ ttl }}
+      allowed_domains:
+        - {{ app }}.service.consul
+        - nearest-{{ app }}.service.consul
```
blarghmatey: For the `nearest` services it's a stored query so the domain ends up being `.query.consul` rather than `.service.consul`

salt/vault/secret_backends.sls
```diff
@@ -1,4 +1,6 @@
 {% set SIX_MONTHS = '4368h' %}
+{% set pki_ttl = '8760h' %}
```
blarghmatey: Same as above, prefer a descriptive constant describing the time frame

salt/vault/secret_backends.sls
```diff
@@ -15,6 +17,28 @@ enable_mitx_aws_secret_backend:
     - lease_max: {{ SIX_MONTHS }}
     - lease_default: {{ SIX_MONTHS }}
 
+enable_pki_intermediate_backend:
+  vault.secret_backend_enabled:
+    - backend_type: pki
+    - mount_point: pki_int
```
blarghmatey: We are using `-` separators for our other backends so we should stick with that convention. Also, a more descriptive mount point would be helpful (e.g. `pki-intermediate-ca`)

salt/vault/secret_backends.sls
```diff
@@ -15,6 +17,28 @@ enable_mitx_aws_secret_backend:
...
+    - ttl_default: {{ pki_ttl }}
+    - connection_config:
+        issuing_certificates: 'https://vault.service.consul:8200/vi/pki_int/ca'
+        crl_distribution_points: 'https://vault.service.consul:8200/v1/pki_int/crl'
+
+{% for environment in env_settings.environments %}
+enable_pki_intermediate_{{ environment }}_backend:
+  vault.secret_backend_enabled:
+    - backend_type: pki
+    - mount_point: pki_int_{{ environment }}
```
blarghmatey: Same as above with using `-`, also, maybe use a name like `pki-{{ environment }}-ca`? I'm happy to stick with `pki-{{environment }}-intermediate-ca` as well

salt/vault/secret_backends.sls
blarghmatey: Might also be more descriptive (though not _technically_ accurate) to refer to this as the 'root' CA

salt/vault/secret_backends.sls
```diff
@@ -15,6 +17,28 @@ enable_mitx_aws_secret_backend:
...
+    - description: Backend to create self-signed certificates for {{ environment }}
```
blarghmatey: Clarification: They're not technically self-signed since they've got a trust chain, it's just not a CA that's widely acknowledged and part of the default trust store.

blarghmatey:
The next step after this will be to write a utility state that populates the trust store for the servers with the certificate chain for the given environment

salt/vault/secret_backends.sls
```diff
@@ -15,6 +17,28 @@ enable_mitx_aws_secret_backend:
...
+    - ttl_default: {{ pki_ttl }}
+    - connection_config:
+        issuing_certificates: 'https://vault.service.consul:8200/vi/pki_int_{{ environment }}/ca'
+        crl_distribution_points: 'https://vault.service.consul:8200/v1/pki_int_{{ environment }}/crl'
```
blarghmatey: Don't we need to specify information about how this intermediate CA gets generated? (e.g. submitting a CSR to the "root" CA and then populating the mount point with that signed cert)

pillar/vault/roles/pki.sls
shaidar: Doesn't that defeat the purpose since then if/when you want to change the variable you'd have to change it everywhere you've used it?

salt/vault/secret_backends.sls
shaidar: The intermediate CA needs to have a signed cert, but that would need to happen manually given that the "root" CA is offline.

salt/vault/secret_backends.sls
```diff
@@ -20,23 +20,23 @@ enable_mitx_aws_secret_backend:
 enable_pki_intermediate_backend:
   vault.secret_backend_enabled:
     - backend_type: pki
-    - mount_point: pki_int
```
blarghmatey: I think changing this to something like `pki-intermediate-ca` might be less confusing in the long run

PR #664, opened 2018-07-19
Table formatted data showing source of alerts and notifications

shaidar:
#### What's this PR do?
Table formatted data showing source of alerts and notifications, how they're triggered, and what slack channel is notified.


docs/infrastructure/notifications.md
```diff
@@ -0,0 +1,37 @@
+| Source                          | Environment        | Trigger                                   | Slack                  | Schedule |
```
blarghmatey: Maybe change the `Slack` heading to be `destination` and specify which alerts are being routed through OpsGenie

blarghmatey:
:+1:

PR #666, opened 2018-07-26
Began adding configuration for metric beat

blarghmatey:
#### What's this PR do?
Adds configuration for installing and setting up metricbeat for our servers

### DIFF
--- a/pillar/elastic_stack/beats/init.sls
+++ b/pillar/elastic_stack/beats/init.sls
@@ -1,0 +1,37 @@
+elastic_stack:
+  beats:
+    metricbeat:
+      config:
+        metricbeat.config.modules:
+          path: /etc/metricbeat/modules.d/*.yml
+          reload.enabled: true
+          reload.period: 30s
+        name: {{ grains['id'] }}
+        tags:
+          - {{ grains['id'] }}
+          - {{ grains.get('environment') }}
+          - {{ grains.get('roles') }}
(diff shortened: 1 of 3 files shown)


pillar/elastic_stack/beats/init.sls
```diff
@@ -0,0 +1,37 @@
...
+              - cpu
+              - filesystem
+              - load
+              - memory
+              - network
+              - process
+              - process_summary
+              - uptime
+            enabled: 'true'
+            period: 1s
```
shaidar: Isn't this too low? Would for example `5s` or `10s` still give us a good picture?

shaidar:
We probably need to evaluate the need for increased disk space on the ES logging cluster once we start deploying this

pillar/elastic_stack/beats/init.sls
blarghmatey: In order for it to be actually useful it's best to go as low as possible. If we can't handle the volume of writes at 1s granularity then we can try going coarser, but this is a good place to start.

blarghmatey:
I'm sure we'll need to scale up the cluster regardless as we add more logs and do more work on adding visibility to our full infrastructure. We will also be wanting to add packetbeat and auditbeat to the majority of our instanaces, and in some places we'll want to run heartbeat.

shaidar:
👍

PR #667, opened 2018-07-26
Logrotate template and state file to handle log file rotation

shaidar:
#### What's this PR do?
Logrotate template and state file to handle log file rotation defined in pillar

### DIFF
--- a/salt/utils/logrotate.sls
+++ b/salt/utils/logrotate.sls
@@ -1,0 +1,12 @@
+{% set logrotate_files = salt.pillar.get('logrotate') %}
+{% for name, settings in logrotate_files.items() %}
+
+generate_{{ name }}_logrotate:
+  file.managed:
+    - name: /etc/logrotate.d/{{ name }}
+    - source: salt://utils/logrotate/conf.jinja
+    - template: jinja
+    - mode: '0644'
+    - context:
+        logrotate_files: {{ logrotate_files }}
+{% endfor %}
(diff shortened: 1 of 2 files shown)


salt/utils/logrotate.sls
```diff
@@ -0,0 +1,12 @@
...
+{% for name, settings in logrotate_files.items() %}
+
+generate_{{ name }}_logrotate:
+  file.managed:
+    - name: /etc/logrotate.d/{{ name }}
+    - source: salt://utils/logrotate/conf.jinja
+    - template: jinja
+    - mode: '0644'
+    - context:
+        logrotate_files: {{ logrotate_files }}
```
blarghmatey: I'm assuming this should be the `settings` variable from the `.items()` iteration

salt/utils/logrotate/conf.jinja
```diff
@@ -0,0 +1,16 @@
+{%- for name, settings in logrotate_files.items() -%}
```
blarghmatey: I'm assuming this is intended to be `settings` instead of `logrotate_files`

shaidar: No, believe it is correct. A sample pillar file looks like the following:
```logrotate:
  test0:
    conf_file: /etc/logrotate.d/test0
    name: /var/log/test0.log
    options:
      - rotate 2
      - size 5k
      - missingok
     - postrotate systemctl restart nginx
```

blarghmatey:
:+1: as long as the intent is to have one block per file

shaidar:
Yeah think it might be best to have each service have its separate config file. That way, we can use same configs across multiple services.

PR #674, opened 2018-08-06
Removing override of static url to eliminate cloudfront

blarghmatey:
The `static_url` attribute gets compiled into the static assets for loading font files. This means that when we build our AMIs for edX (which happens in QA) the cloudfront domain gets set at that point. When we deploy to production, the production instances then try to resolve font files from QA hosts and the only option to override the value is to use `sed` or recompile the assets. The alternative is to remove cloudfront from the equation and just load the assets from disk per instance. This commit implements that strategy to remove complexity from our deployments.

Build a new set of AMIs and verify that they aren't attempting to load assets from cloudfront


shaidar:
👍

blarghmatey:
We are going to update the deploy process instead to build an AMI for production that has the assets recompiled for the production CDN.

PR #678, opened 2018-08-13
Added config to split mailgun records by event type

blarghmatey:
mitodl/open-discussions#663

#### What's this PR do?
Splits the mailgun event records by the trailing tag element so that the events will be uploaded to S3 as separate files to simplify processing in AWS Glue and Athena

### DIFF
--- a/pillar/fluentd/server.sls
+++ b/pillar/fluentd/server.sls
@@ -199,11 +199,18 @@ fluentd:
                     - aws_sec_key: __vault__:cache:aws-mitx/creds/read-write-{{ data_lake_bucket }}>data>secret_key
                     - s3_bucket: {{ data_lake_bucket }}
                     - s3_region: us-east-1
-                    - path: mailgun
-                    - buffer_path: {{ fluentd_directories.data_lake }}
+                    - path: mailgun/${tag.split('.').last'}/
+                    - nested_directives:
+                        - directive: buffer
+                          directive_arg: tag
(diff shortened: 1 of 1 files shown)


pillar/fluentd/server.sls
```diff
@@ -199,11 +199,18 @@ fluentd:
                     - aws_sec_key: __vault__:cache:aws-mitx/creds/read-write-{{ data_lake_bucket }}>data>secret_key
                     - s3_bucket: {{ data_lake_bucket }}
                     - s3_region: us-east-1
-                    - path: mailgun
-                    - buffer_path: {{ fluentd_directories.data_lake }}
```
shaidar: I believe `buffer_path` is [required](https://docs.fluentd.org/v0.12/articles/out_s3#buffer_path-(required))

blarghmatey: That link is for the v0.12 version of FluentD. You can refer to https://github.com/fluent/fluent-plugin-s3 for the S3 output and https://docs.fluentd.org/v1.0/articles/out_s3 in the fluentD docs for the current version.

shaidar: But we're running v0.14 and not 1.0. At least the aggregators are

blarghmatey: v0.14 is what turned into v1.0 so the syntax should be the same.

shaidar:
👍

PR #680, opened 2018-08-16
Removed datadog rabbit integration from all environments except prod

shaidar:


pillar/top.sls
```diff
@@ -210,12 +210,17 @@ base:
     - mysql.devstack
     - elasticsearch.devstack
     - rabbitmq.devstack
-  'roles:rabbitmq':
+  'G@roles:rabbitmq and P@environment:(mitx-production|production-apps)':
```
blarghmatey: You can just move this line directly above the datadig integration line and remove the duplicated block

pillar/top.sls
```diff
@@ -215,6 +215,8 @@ base:
     - rabbitmq
     - fluentd.rabbitmq
     - consul.rabbitmq
+  'G@roles:rabbitmq and P@environment:(mitx-production|production-apps)':
+    - match: grain
```
blarghmatey: This should be `- match: compound`

Issue #679, opened 2018-08-13
Refactor edX build/deploy for recompiling assets


blarghmatey:
The static assets are being compiled into the AMI during our build process which causes font paths to be pointed at the QA cloudfront distribution. In order to resolve that situation we need to recompile the assets as part of the production deployment to pick up the production cloudfront domain.

PR #683, opened 2018-08-24
Added recompile assets function

shaidar:
[Issue#679](https://github.com/mitodl/salt-ops/issues/679)

#### What's this PR do?
Recompiles assets if the instances being deployed are production instances in order to use the prod cloudfront distribution

### DIFF
--- a/salt/orchestrate/edx/deploy.sls
+++ b/salt/orchestrate/edx/deploy.sls
@@ -142,7 +142,7 @@ deploy_edx_cloud_map:
 
 {% for purpose in PURPOSES %}
 {% set codename = defined_purposes[purpose].versions.codename %}
-{% set release_version = salt.sdb.get('sdb://consul/edxapp-{}-release-version'.format(codename)) %}
+{% set release_version = salt.sdb.get('sdb://consul/edxapp-{}-{}-release-version'.format(ENVIRONMENT, codename)) %}
 sync_external_modules_for_{{ purpose }}_{{ codename }}_edx_nodes:
   salt.function:
     - name: saltutil.sync_all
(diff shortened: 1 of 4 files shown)


blarghmatey:
We don't want to take this approach because that will recompile the assets for each instance separately and they will have a different content hash. What we need to do is build a new AMI in the production environment (or with assets compiled for production) that will be deployed into the production environment.

blarghmatey:
The worker and app AMI ID lookups need to be updated in the `edxapp_update_ami_id` script to use the environment attribute.

blarghmatey:
We also need to update the map template to pull from the proper SDB URI (e.g. https://github.com/mitodl/salt-ops/blob/master/salt/orchestrate/aws/map_templates/edx.yml#L13)

salt/orchestrate/edx/deploy.sls
```diff
@@ -184,6 +184,15 @@ build_{{ purpose }}_{{ codename }}_edx_nodes:
...
     {% endif %}
 
+{% if ENVIRONMENT == 'mitx-production' %}
+{# Recompile assets to pickup the production cloudfront domain #}
+compile_assets_for_edx_{{ purpose }}:
+  cmd.run:
+    - name: /edx/bin/edxapp-update-assets
+    - require:
+        - salt: build_{{ purpose }}_{{ codename }}_edx_nodes
+{% endif %}
```
blarghmatey: This should be in the `build_ami` state, not the deploy state. If we run it here then all of the instances will have a different content hash and the CDN won't work...

blarghmatey:
:+1:

PR #684, opened 2018-08-28
Increased OVS default disk size to 20GB

shaidar:
Increased OVS default disk size to 20GB since the instance performs some local processing and requires a larger disk.


blarghmatey:
:+1:

Issue #204, opened 2017-01-10
Add logging configuration for edX Nginx servers


blarghmatey:
This should be added to the `fluentd/mitx.sls` pillar configuration. This is also a good file to use for reference about how to configure the log inputs.

shaidar:
We looked at this and it appears that it might be best to tweak nginx log settings first. So, need to take a look at the nginx conf and see how to tweak settings to get relevant logs and then configure fluentd.

PR #688, opened 2018-08-31
Set mitx nginx logging format to match our other applications

shaidar:
[Issue#204](https://github.com/mitodl/salt-ops/issues/204)

#### What's this PR do?
Sets mitx nginx logging format to match the our other applications


shaidar:
@blarghmatey Doesn't the edxapp role already call the nginx role and install that as part of the deployment?

blarghmatey:
It calls the Ansible Nginx role, but not the Salt one

blarghmatey:
Honestly, it might be more straightforward to just do a file.managed directly as part of the edx.prod state

blarghmatey:
That looks good, but we'll still want to add the log format name to the ansible vars as well to be sure that it gets picked up.

blarghmatey:
:+1:

PR #690, opened 2018-09-05
Added a healtchecks check to curl when the cron job completes

shaidar:


blarghmatey:
:+1:

blarghmatey:
We should use a specific Vault path for that healthchecks URL since we can have multiples for different scripts

PR #694, opened 2018-09-17
Added monit init to pillar top file

shaidar:
Added monit init to pillar top file since that adds the slack shell script which is used for notifications


blarghmatey:
:+1:

PR #696, opened 2018-09-28
Added salt reactor to notify slack when there's a cached read miss event

shaidar:
Added salt reactor to notify slack when there's a cached read miss event.


blarghmatey:
:+1:

PR #697, opened 2018-10-04
Added schedule to refresh credentials on regular basis

shaidar:
#### What's this PR do?
In order to refresh vault credentials automatically prior to expiration, we added a scheduler to run every so often and update the configs.

### DIFF
--- a/pillar/apps/odlvideo.sls
+++ b/pillar/apps/odlvideo.sls
@@ -141,6 +141,7 @@ django:
   states:
     setup:
       - apps.odlvideo.install
+    config:
       - apps.odlvideo.configure
     post_install:
       - apps.odlvideo.post_deploy
--- a/pillar/apps/odlvideo.sls
+++ b/pillar/apps/odlvideo.sls
@@ -141,6 +141,7 @@ django:
   states:
     setup:
       - apps.odlvideo.install
+      - apps.odlvideo.configure
     post_install:
       - apps.odlvideo.post_deploy
 
--- a/pillar/apps/odlvideo.sls
+++ b/pillar/apps/odlvideo.sls
@@ -50,6 +50,13 @@
 {% set rabbit_creds = salt.vault.cached_read("rabbitmq-{env}/creds/odlvideo".format(env=ENVIRONMENT), cache_prefix=minion_id) %}
 {% set ga_json = salt.vault.read('secret-odl-video/' ~ ENVIRONMENT ~ '/ga-keyfile-json').data.value|json %}
 
+schedule:
+  refresh_{{ app_name }}_credentials:
+    days: 14
+    function: state.sls
+    args:
+      - django.config
+
 python:
   versions:
     - number: {{ python_version }}
--- a/pillar/backups/mitx.sls
+++ b/pillar/backups/mitx.sls
@@ -10,6 +10,13 @@
 {% set efs_id = 'fs-1f27ae56' %}
 {% endif %}
 
+schedule:
+  refresh_{{ app_name }}_configs:
+    days: 21
+    function: state.sls
+    args:
+      - backups.backup
+
 backups:
   enabled:
   {% for purpose, purpose_data in env_data.purposes.items() %}
--- a/pillar/backups/mitx.sls
+++ b/pillar/backups/mitx.sls
@@ -10,13 +10,6 @@
 {% set efs_id = 'fs-1f27ae56' %}
 {% endif %}
 
-schedule:
-  refresh_{{ app_name }}_configs:
-    days: 21
-    function: state.sls
-    args:
-      - backups.backup
-
 backups:
   enabled:
   {% for purpose, purpose_data in env_data.purposes.items() %}
--- a/pillar/backups/restore.sls
+++ b/pillar/backups/restore.sls
@@ -17,13 +17,6 @@
     'forum',
 ] %}
 
-schedule:
-  refresh_{{ app_name }}_configs:
-    days: 21
-    function: state.sls
-    args:
-      - backups.restore
-
 restores:
 {% for purpose, purpose_data in env_data.purposes.items() %}
 {% if 'live' in purpose or 'draft' in purpose %}
--- a/pillar/backups/restore.sls
+++ b/pillar/backups/restore.sls
@@ -16,6 +16,14 @@
     'gitlog',
     'forum',
 ] %}
+
+schedule:
+  refresh_{{ app_name }}_configs:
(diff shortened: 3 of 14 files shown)


pillar/apps/odlvideo.sls
```diff
@@ -50,6 +50,13 @@
 {% set rabbit_creds = salt.vault.cached_read("rabbitmq-{env}/creds/odlvideo".format(env=ENVIRONMENT), cache_prefix=minion_id) %}
 {% set ga_json = salt.vault.read('secret-odl-video/' ~ ENVIRONMENT ~ '/ga-keyfile-json').data.value|json %}
 
+schedule:
+  refresh_{{ app_name }}_credentials:
+    days: 14
+    function: state.sls
+    args:
+      - django.config
```
blarghmatey: For OVS, we'll need to move the .env file creation into a separate state file and then update the pillar to call that as part of the django.config state. As it stands right now it's part of the `django.install` step.

pillar/backups/mitx.sls
```diff
@@ -10,6 +10,13 @@
 {% set efs_id = 'fs-1f27ae56' %}
 {% endif %}
 
+schedule:
```
blarghmatey: This one isn't necessary since the credentials would get updated as part of the backup orchestrate state and we want it to run nightly.

pillar/backups/restore.sls
```diff
@@ -16,6 +16,14 @@
     'gitlog',
     'forum',
 ] %}
+
+schedule:
```
blarghmatey: Same as above, not necessary since this should run weekly and will pull new credentials as needed.

pillar/edx/init.sls
```diff
@@ -24,3 +24,10 @@ edx:
     TLS_LOCATION: '/etc/pki/tls/certs'
     TLS_KEY_NAME: 'edx-ssl-cert'
     max_upload_size: 20
+
+schedule:
+  refresh_mitx-{{ environment }}_configs:
+    days: 21
+    function: state.sls
+    args:
+      - edx.run_ansible pillar="{'edx': {'ansible_flags':  '--tags install:configuration'}}"
```
blarghmatey: The pillar attributes should get passed as kwargs (I think)

pillar/mongodb/init.sls
```diff
@@ -31,3 +31,11 @@ mongodb:
...
   replset_config: {{ replset_config }}
   cluster_key: __vault__::secret-{{ business_unit }}/{{ environment }}/mongodb-cluster-key>data>value
+
+{% if 'production' in environment %}
+schedule:
+refresh_datadog_mongodb-{{ ENVIRONMENT }}_credentials:
+  days: 21
+  function: state.sls
+  args:
+    - datadog.plugins
```
blarghmatey: Missing the `endif` tag

pillar/rabbitmq/init.sls
```diff
@@ -30,3 +30,11 @@ rabbitmq:
...
         password: __vault__:gen_if_missing:secret-{{ BUSINESS_UNIT }}/{{ ENVIRONMENT }}/rabbitmq-admin-password>data>value
   erlang_cookie: __vault__:gen_if_missing:secret-{{ BUSINESS_UNIT }}/{{ ENVIRONMENT }}/erlang_cookie>data>value
+
+{% if 'production' in ENVIRONMENT %}
+schedule:
+refresh_datadog_rabbitmq-{{ ENVIRONMENT }}_credentials:
+  days: 21
+  function: state.sls
+  args:
+    - datadog.plugins
```
blarghmatey: MIssing the `endif` tag. Also needs to run the `rabbitmq.set_datadog_permissions` state which can get added as another element in the args list.

pillar/reddit.sls
```diff
@@ -24,8 +24,8 @@
 {% endfor %}
 
 schedule:
-  refresh_reddit_configs:
-    days: 25
+  refresh_{{ app_name }}_configs:
+    days: 14
     function: state.sls
     args:
       - reddit.config
```
blarghmatey: This also requires the `pgbouncer` state to be executed.

pillar/apps/odlvideo.sls
```diff
@@ -141,6 +141,7 @@ django:
   states:
     setup:
       - apps.odlvideo.install
+      - apps.odlvideo.configure
```
blarghmatey: This state should go under the path `states:config` so that it gets run when the `django.config` state gets executed. You can refer to the mitx-cas or starcellbio pillar for an example.

pillar/edx/init.sls
```diff
@@ -30,4 +30,6 @@ schedule:
     days: 21
     function: state.sls
     args:
-      - edx.run_ansible pillar="{'edx': {'ansible_flags':  '--tags install:configuration'}}"
+      - edx.run_ansible
+      kwargs:
```
blarghmatey: Indentation. Also, I"m not positive that this is the correct syntax so it's worth double checking.

salt/apps/odlvideo/configure.sls
```diff
@@ -0,0 +1,14 @@
...
+
+create_env_file_for_odlvideo:
+  file.managed:
+    - name: /opt/{{ salt.pillar.get('django:app_name') }}/.env
+    - contents: |
+        {%- for var, val in salt.pillar.get('django:environment').items() %}
+        {{ var }}={{ val }}
+        {%- endfor %}
+    - require:
+        - deploy_application_source_to_destination
```
blarghmatey: This require statement will probably fail when executed in isolation as `django.config` but is probably not needed now that it is getting run at a different step in the execution.

pillar/edx/init.sls
```diff
@@ -31,5 +31,5 @@ schedule:
     function: state.sls
     args:
       - edx.run_ansible
-      kwargs:
-        pillar: "{'edx': {'ansible_flags':  '--tags install:configuration'}}"
+    kwargs:
+      pillar: "{'edx': {'ansible_flags':  '--tags install:configuration'}}"
```
blarghmatey: The actual pillar data should be structured as a YAML dict under the pillar key.

blarghmatey: This still needs to be modified to be a yaml structure rather than a string.
```suggestion
      pillar:
        edx:
          ansible_flags: '--tags install:configuration'
```

blarghmatey:
:+1: for now but we still need to change the timeframes so that they fall within the expiry period of the cache.

Issue #653, opened 2018-07-11
Create salt states for building staging and production origin servers


blarghmatey:
The aka-ocw and ocw2 servers (possibly ocw-mirror as well) are just Nginx servers with a certain directory structure that has SSH keys for known users that cms2 publishes to. Building them with salt should be easy and quick to implement so we should automate these first.

PR #698, opened 2018-10-17
State to build ocw-aka instance

shaidar:
Part of [Issue#653](https://github.com/mitodl/salt-ops/issues/653)

#### What's this PR do?
Create an instance of aka-ocw using Nginx and letsencrypt. In addition to creating the folder structure and some of the required users.

A few things to point out:
- Created users are missing ssh keys but wasn't exactly sure how we wanted to handle that
- Used a different domain name from the existing instance since I'm assuming they might be running in parallel initially
- The state file under `apps.ocw.ocw-aka.install` might raise questions since it appears that structure is used by the uwsgi apps, but this kinda felt that it belonged here, however I can easily move to wherever makes more sense.


pillar/nginx/ocw_aka.sls
```diff
@@ -0,0 +1,53 @@
...
+          enabled: True
+          config:
+            - server:
+                - server_name: {{ server_domain_names }}
+                - listen:
+                    - 80
+                - listen:
+                    - '[::]:80'
+                - location /:
+                    - return: 301 https://$host$request_uri
```
blarghmatey: I'm not sure if we want/need this to force redirect to HTTPS.

pillar/nginx/ocw_aka.sls
```diff
@@ -0,0 +1,53 @@
...
+            - server:
+                - server_name: {{ server_domain_names }}
+                - listen:
+                    - 443
+                    - ssl
+                    - default
+                - listen:
+                    - '[::]:443'
+                    - ssl
+                - root: /ocwdata/apache-tomcat/webapps/
```
blarghmatey: I don't think we should cargo-cult the tomcat, etc. directories. We should just make it reasonable and semantically relevant for our own sanity. There will be an initial bit of work to update configs, but worth it in the end (I hope...)

salt/apps/ocw/ocw_aka/install.sls
```diff
@@ -0,0 +1,35 @@
+{% set users = [fsuser, ocwuser, sshacs] %}
+{% set root_directory = '/ocwdata/apache-tomcat/webapps/' %}
```
blarghmatey: Same as above, we should make this reasonable. Possibly just put it to `/var/www/ocw`

salt/apps/ocw/ocw_aka/install.sls
```diff
@@ -0,0 +1,35 @@
...
+{% set directories = ['about', 'course', 'donate', 'educator', 'faculty', 'give',
+                     'help', 'high-school', 'icons', 'images', 'jsp', 'jw-player-free',
+                     'jwplayer', 'OcwWeb', 'ocw-labs', 'resources', 'rss', 'scp48112',
+                     'scripts', 'search', 'styles', 'subscribe', 'support', 'terms',
+                     'webfonts', 'xml', 'xsd'] %}
+
+{% for user in users %}
+create_{{ user }}_user:
+  - name: {{ user }}
+  - shell: /bin/bash
```
blarghmatey: We should probably populate the users with appropriate ssh_authorized_keys as well.

salt/apps/ocw/ocw_aka/install.sls
```diff
@@ -0,0 +1,35 @@
...
+
+create_{{ root_directory }}_directory:
+  file.directory:
+    - name: {{ root_directory }}
+    - user: fsuser
+    - group: www-data
+    - makedirs: True
+    - require:
+      - user: create_{{ user }}_user
+      - file: create_{{ root_directory }}_directory
```
blarghmatey: This is requiring itself...

blarghmatey: This is also not strictly necessary since the `makedirs` setting for the following states is set to `True`

salt/apps/ocw/ocw_aka/install.sls
```diff
@@ -0,0 +1,35 @@
...
+
+{% for dir in directories %}
+create_{{ dir }}_directory:
+  file.directory:
+    - name: {{ dir }}
```
blarghmatey: This should be
```suggestion
    - name: {{ root_directory }}/{{ dir }}
```

salt/environment_settings.yml
```diff
@@ -574,6 +574,14 @@ environments:
     business_unit: open-courseware
     network_prefix: '10.100'
     vpc_name: OCW
+    purposes:
+      ocw-production-aka:
+        business_unit: open-courseware
+        domains:
+          - ocw-aka.mit.edu
```
blarghmatey: See above comment about domain

salt/orchestrate/aws/cloud_profiles/ocw-production-aka.conf
```diff
@@ -0,0 +1,24 @@
+# -*- mode: yaml; coding: utf-8; -*-
+ocw-production-aka:
+  provider: mitx
+  size: t3.small
+  image: {{ salt.sdb.get('sdb://consul/debian_ami_id')|default('ami-03006931f694ea7eb') }}
+  ssh_username: admin
+  ssh_interface: private_ips
+  script_args: -Z -A salt.private.odl.mit.edu
+  iam_profile: ocw-production-instance-role
+  ebs_optimized: True
```
blarghmatey: T3 instance types don't have this enabled so we should remove this line.

salt/orchestrate/aws/cloud_profiles/ocw-production-aka.conf
```diff
@@ -0,0 +1,24 @@
...
+  tag:
+    role: ocw-aka
```
blarghmatey: We should probably update the naming to be more obvious since `aka` is arbitrary. Possibly something like `origin-server`

salt/apps/ocw/ocw_aka/install.sls
shaidar: @blarghmatey  This doesn't appear to be an option in the state. Did you mean just using some kinda `file.manage` to add the keys after creating the users?

blarghmatey: There's a module `ssh.set_auth_key` that can be wrapped in a state

blarghmatey:
:+1:

PR #701, opened 2018-10-23
Testing http_status beacon on ovs-rc

shaidar:


pillar/beacons/http_status_odl_video_service.sls
```diff
@@ -0,0 +1,10 @@
+beacons:
+  http_status:
+    - sites:
+        odl-video-rc-apps:
+          url: "https://video-rc.odl.mit.edu/status?token=rc-apps"
+          json_response:
+            - path: 'certificate:status'
+              value: up
```
blarghmatey: Is this missing the comparison or does it default to `==` if there is no setting?

blarghmatey:
Do we need to set up a reactor?

pillar/beacons/http_status_odl_video_service.sls
shaidar: Just added that. Forgot about it.

shaidar:
Wanted to initially just make sure the extension works as expected and was then going to add all the pillar and reactor stuff.

blarghmatey:
:+1:

PR #702, opened 2018-10-30
Added ovs prod beacon and interval

shaidar:
#### What's this PR do?
Uses the new status beacon that we wrote and compares the response from the django-status package to the expected values specified in pillar. This PR adds pillar data to check on OVS RC and prod for valid certificates and whether the overall `status_all` is up. The check runs every 10 min. 

I tested this on my local vagrant master and minion instances and saw the beacons raising an alert.


blarghmatey:
:+1:

Issue #230, opened 2017-02-10
Create and schedule integration test for lmod proxy


blarghmatey:
In order to ensure that lmod proxy is properly operational we need to write an integration test that gets run on a daily schedule and triggers an alert on failure.

blarghmatey:
@shaidar this would be a good use case for monit

pdpinch:
What's the status of this issue?

shaidar:
The PR has been merged to have a status endpoint on lmod_proxy but I have yet to deploy it and test it on QA and prod. On my radar for this week.

pdpinch:
Checking in on this again.

pdpinch:
Any news on this?

PR #704, opened 2018-11-07
Beacon to check on lmodproxy status

shaidar:
[Issue#230](https://github.com/mitodl/salt-ops/issues/230)

#### What's this PR do?
Uses `http_status` beacon to check lmodproxy status endpoint is returning a 200 and also verifies that the app_cert is still valid

Tested locally on vagrant instances and appears to be functional.


blarghmatey:
:+1:

PR #705, opened 2018-11-13
Added nginx block to allow curling across nodes from within cluster

blarghmatey:


pillar/nginx/micromasters_es.sls
```diff
@@ -19,6 +19,11 @@ nginx:
                 - listen:
                     - '[::]:443'
                     - ssl
+                - location /:
```
shaidar: I know it's within the VPC, but do we want to restrict it a bit further for example `/_cluster`?

shaidar:
👍

PR #707, opened 2018-11-13
Added dnsutils so that dig is available on instances

shaidar:
#### What's this PR do?
Dig is usually helpful to have on instances especially when troubleshooting. It's part of dnsuitls (on debian and bind-utils on Redhat). Added it to the list of installed packages on all the instances. Additionally, renamed the state from `install_pip` to `install_libs` as that state files installs more than pip and its dependencies now and it made sense to rename it.

### DIFF
--- a/salt/top.sls
+++ b/salt/top.sls
@@ -1,6 +1,6 @@
 base:
   '*':
-    - utils.install_pip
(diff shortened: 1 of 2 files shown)


blarghmatey:
:+1:

PR #712, opened 2018-11-15
Updated vault refresh schedules to be more frequent

blarghmatey:
In order to ensure that the scheduled refresh will always fall within the expiration window of the Vault cache I updated the schedule to be fewer than the 7 day window. The current schedules couldd easily fall outside of that 7 day window and end up resulting in invalid credentials for several days.


shaidar:
👍

PR #718, opened 2018-12-04
Overriding the DATA_DIR to fix course imports

blarghmatey:
With the Hawthorn release the DATA_DIR setting is distinct from the COURSE_DATA_DIR which results in errors when trying tom import course archives. In order to resolve this issue we need to set the DATA_DIR back to the course_data_dir/github_repo_dir

https://github.mit.edu/mitx/Spring-2019/issues/6

#### What's this PR do?
Fixes course imports


shaidar:
👍

PR #724, opened 2018-12-19
Added new email flag and removed unused smtp config

shaidar:
#### What's this PR do?
Leverages the new functionality added by [PR#113](https://github.com/mitodl/edx-platform/pull/113) by setting a key in ansible_vars.
Also removed the `edx:smtp` section as we don't appear to be using the edx smtp state and deleted some of the templates that the state calls.


blarghmatey:
:+1:

PR #725, opened 2018-12-19
Changed orchestrate state to use env settings

shaidar:
#### What's this PR do?
Refactored code to be in line with the `deploy` state and use all the keys set in env settings.


blarghmatey:
:+1:

PR #728, opened 2018-12-20
Removed salt-minion args since those are defined in provider

shaidar:


blarghmatey:
This is mostly fine, but we either need to add the `-A salt.private.odl.mit.edu` in the cloud provider config, or we need to keep the section that reads
```
minion:
  master:
    - salt.private.odl.mit.edu
```

blarghmatey:
:+1:

PR #729, opened 2018-12-26
Updated to run via cron

blarghmatey:
Updated to execute via cron tasks instead of relying on salt to stay connected for the duration of execution

### DIFF
--- a/salt/backups/templates/backup_consul.sh
+++ b/salt/backups/templates/backup_consul.sh
@@ -10,3 +10,7 @@ SNAPFILE={{ backupdir }}/consul_snapshot-`date +%Y-%m-%d_%H-%M-%S`.snap
 aws s3 cp $SNAPFILE s3://odl-operations-backups/{{ settings.get('directory', 'consul') }}/
 
 rm $SNAPFILE
+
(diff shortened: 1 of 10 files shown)


salt/backups/templates/backup_consul.sh
```diff
@@ -10,3 +10,5 @@ SNAPFILE={{ backupdir }}/consul_snapshot-`date +%Y-%m-%d_%H-%M-%S`.snap
 aws s3 cp $SNAPFILE s3://odl-operations-backups/{{ settings.get('directory', 'consul') }}/
 
 rm $SNAPFILE
+
+salt-call event.fire_master '{"data": "Completed backup of Consul"}' backup/{{ ENVIRONMENT }}/{{ settings.service.title }}/completed
```
shaidar: We sure that `settings.service.title` is correct?

shaidar:
@blarghmatey Should this not also include a healthcheck ping to ensure that it's not silently failing?

blarghmatey:
@shaidar the scripts now have healtcheck pings included

shaidar:
👍

PR #730, opened 2019-01-02
Create secrets backends matching ones in pillar vault pki

shaidar:
#### What's this PR do?
Create secrets backend matching ones in [pillar vault pki](https://github.com/mitodl/salt-ops/blob/master/pillar/vault/roles/pki.sls).


blarghmatey:
This looks like it would create a mount point for every app in every env. Shouldn't each env just have an intermediate CA and then set up the various roles for the apps across those CAs per env?

blarghmatey:
Small nitpick, but for whatever reason `int` doesn't feel right to me, partly because of the fact that it is a type name. Can we change it to just say `intermediate`?

blarghmatey:
:+1:

PR #731, opened 2019-01-04
Added minion_id to datadog

shaidar:
#### What's this PR do?
When Datadog alerts are triggered, the `minion_id` is not part of the alert message which causes us to have to use the provided `instance id` or IP address to figure out which minion triggered the alert. This PR adds that information to the message.


blarghmatey:
:+1:

PR #733, opened 2019-01-08
Added missing port setting for edxapp outgoing mail

blarghmatey:
The outgoing SMTP connection times out due to the fact that it is defaulting to port 25, whereas outgoing.mit.edu is expecting connections on port 587


shaidar:
👍

PR #735, opened 2019-01-11
Added S3 bucket definition for parsed course data from OCW

blarghmatey:


salt/orchestrate/aws/s3_buckets/operations.sls
```diff
@@ -0,0 +1,11 @@
+open-learning-course-data:
+  boto_s3_bucket.present:
+    - Bucket: open-learning-course-data
+    - Versioning:
+        Status: Enabled
+    - region: us-east-1
+    - Tagging:
+        OU: operations
+        business_unit: operations
```
shaidar: Should we have an `open` or `discussions` business_unit and use here?

blarghmatey: I was wondering about that too and it's probably worth getting @pdpinch's input on the matter.

shaidar:
It was decided to add a business unit for `open` to keep better track of those resources

shaidar:
👍

PR #736, opened 2019-01-16
Added allow_bare_domain param and additional subdomains

shaidar:
#### What's this PR do?
Allows us to generate certs using a common_name specified in allowed_subdomains by adding the `allow_bare_domains` param. Additionally, `consul` and `fluentd` services cross-communicate and thus added a few entries for them. Lastly, changed the path name as it sounded redundant given the name of the backend that already contains the env_name

### DIFF
--- a/pillar/vault/roles/pki.sls
+++ b/pillar/vault/roles/pki.sls
@@ -14,7 +14,7 @@ vault:
     {% for type in ['client', 'server'] %}
     {{ app }}-{{ env_name }}-{{ type }}-pki:
       backend: pki-intermediate-{{ env_name }}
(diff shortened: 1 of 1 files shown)


pillar/vault/roles/pki.sls
```diff
@@ -27,6 +27,10 @@ vault:
           {{ app }}.service.consul
           nearest-{{ app }}.query.consul
           {{ app }}-master.service.consul
+          {% for ['consul', 'fluentd'] in app %}
```
blarghmatey: Shouldn't this be an if clause instead of a for loop?

blarghmatey:
:+1:

PR #737, opened 2019-01-18
Added hacks state to manage platform workarounds

blarghmatey:
In this instance there are underscore template files that are being loaded with a `.js` extension in the UI, despite not existing on disk at that location. The easiest fix (other than fixing the webpack/requirejs build configs) is to copy those files to the expected location.

Zendesk #24351

#### What's this PR do?
Fixes the video editor UI in studio that is broken by the platform trying to load underscore templates with a `.js` extension


shaidar:
👍

PR #740, opened 2019-01-22
Added states and config for deploying mit open etl script

blarghmatey:
#### What's this PR do?
Configures deployment of the MIT Open ETL script to the redash server

In order to allow for mapping data between mailgun events and MIT Open data we need to execute a script to generate the obscured data and populate a table in S3 so that we can join against that in Athena.


pillar/top.sls
```diff
@@ -91,6 +91,7 @@ base:
     - shibboleth.redash
     - apps.redash
     - apps.redash_data_sources
+    - data.mit_open
```
shaidar: Should this be `data.mit_open_etl`?

shaidar:
👍

PR #741, opened 2019-01-28
Deploy salt-proxy to master

shaidar:
#### What's this PR do?
Configures heroku proxy to run on salt master and created a state that can be called to update heroku config for a defined app.

### DIFF
--- a/salt/heroku/proxy_config.sls
+++ b/salt/heroku/proxy_config.sls
@@ -1,4 +1,8 @@
-configure_heroku_proxy:
+{% for env in ['ci', 'rc', 'prod'] %}
+{% for app in ['bootcamp-ecommerce', 'micromasters', 'odl-open-discussions', 'odl-video'] %}
+configure_heroku_proxy_for_{{ app }}-{{ env }}:
   salt_proxy.configure_proxy:
-    - proxyname: heroku_proxy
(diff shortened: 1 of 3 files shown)


salt/heroku/proxy_config.sls
```diff
@@ -0,0 +1,4 @@
+configure_heroku_proxy:
+  salt_proxy.configure_proxy:
+    - proxyname: heroku_proxy
```
blarghmatey: Don't we need to have a different proxy config for each app?

shaidar: Yeah u're right. Was hoping that one would be able to do it, but after u mentioned it, one surely wouldn't make sense in this case.

salt/heroku/proxy_config.sls
```diff
@@ -1,4 +1,8 @@
-configure_heroku_proxy:
+{% for env in ['ci', 'rc', 'prod'] %}
+{% for app in ['bootcamp-ecommerce', 'micromasters', 'odl-open-discussions', 'odl-video'] %}
```
blarghmatey: One issue here is that odl-video only has a CI instance on Heroku... Should still work fine as long as we don't try running a state against it, but could lead to some confusion.

PR #742, opened 2019-01-29
Updated service scripts to use env yaml instead of env_map

blarghmatey:


salt/orchestrate/services/scylladb.sls
```diff
@@ -53,7 +59,7 @@ generate_cloud_map_file:
 deploy_scylladb_nodes:
   salt.runner:
     - name: cloud.map_run
-    - path: /etc/salt/cloud.maps.d/{{ VPC_RESOURCE_SUFFIX }}_scylladb_map.yml
+    - path: /etc/salt/cloud.maps.d/{{ ENVIRONMENT }}_scylladb_map.yml
     - kwargs:
         parallel: True
     - require:
```
shaidar: Is there a reason why lines 150-191 are commented out and not removed?

blarghmatey: I put them there as an initial pass for adding auth but removed them for testing purposes. At this point we don't even have any Scylla servers deployed so :man_shrugging:

PR #749, opened 2019-02-14
Updated settings for building bucket names to work with xpro

blarghmatey:
The settings and names that we were using for S3 in the edX configs had a lot of assumptions about the residential environment built into them. I updated the settings to be a bit more granular in how they get composed so that we can more easily identify which buckets go with which environments and application stacks.

One question that came out of this is whether we should split out the AWS Vault backend a bit more, similar to how we have the other backends configured?

### DIFF
--- a/pillar/edx/ansible_vars/init.sls
+++ b/pillar/edx/ansible_vars/init.sls
@@ -8,7 +8,9 @@
 {% set purpose = salt.grains.get('purpose', 'current-residential-live') %}
 {% set purpose_suffix = purpose.replace('-', '_') %}
 {% set environment = salt.grains.get('environment', 'mitx-qa') %}
-{% set purpose_data = env_settings.environments[environment].purposes[purpose] %}
+{% set env_data = env_settings.environments[environment] %}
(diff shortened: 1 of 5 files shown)


pillar/edx/ansible_vars/init.sls
```diff
@@ -141,8 +141,8 @@ edx:
     #####################################################################
     ########### Auth Configs ############################################
     #####################################################################
-    EDXAPP_AWS_ACCESS_KEY_ID: __vault__:cache:aws-mitx/creds/mitx-s3-{{ purpose }}-{{ environment }}>data>access_key
-    EDXAPP_AWS_SECRET_ACCESS_KEY: __vault__:cache:aws-mitx/creds/mitx-s3-{{ purpose }}-{{ environment }}>data>secret_key
+    EDXAPP_AWS_ACCESS_KEY_ID: __vault__:cache:aws-mitx/creds/{{ bucket_prefix }}-s3-{{ purpose }}-{{ environment }}>data>access_key
```
shaidar: Is `bucket_prefix` set?

shaidar:
👍

PR #750, opened 2019-02-15
Updated xqwatcher deployment to be per course

blarghmatey:
In order to be able to grant access to the course team for a given grader I split out the xqueue-watcher deployment to be per course instead of shared instances. The deployment now iterates over a mapping of courses to build separate groups of instances which allows for independent scaling needs as well.


shaidar:
👍

PR #757, opened 2019-02-25
Readonlyrest plugin conflicts with xpack.security

shaidar:
(Required)

#### What's this PR do?
This change adds a single line to disable xpack.security, as the ES service starts but does not bind to any port. The readonlyrest plugin offers the same features as xpack.security and thus xpack needs to be disabled for the ES service to bind to a port. Came across the solution [here](https://forum.readonlyrest.com/t/how-to-make-readonlyrest-compatible-with-x-pack-plugin/24)


blarghmatey:
:+1:

PR #759, opened 2019-02-26
Added tojson filters for rendering data structures in salt templates

blarghmatey:
The latest version of Salt includes a breaking change for rendering nested data structures directly in YAML from Jinja. This PR adds the `tojson` filter to any locations where we are trying to directly load data structures (lists, dicts, nested objects, etc.) so that we don't have errors trying to render pillar and state data objects.


shaidar:
👍

PR #765, opened 2019-03-07
Reformatted inotify beacon to fix errors in log

shaidar:
#### What's this PR do?
Continual message in the salt-minion logs caused by an issue with the inotify beacon:
```[salt.beacons     :108 ][INFO    ][19557] Beacon inotify configuration invalid, not running.
Configuration for inotify beacon must be a list.
```
This was introduced in the newer version of salt which we are now running on all the residential minions. This change is based on salt [docs](https://docs.saltstack.com/en/latest/ref/beacons/all/salt.beacons.inotify.html) and should fix the problem.

### DIFF
--- a/pillar/edx/inotify_mitx.sls
+++ b/pillar/edx/inotify_mitx.sls
@@ -1,52 +1,53 @@
 beacons:
   inotify:
-    /lib:
-      recurse: True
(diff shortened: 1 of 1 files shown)


blarghmatey:
:+1:

PR #766, opened 2019-03-11
Added schedule to renew etl creds

shaidar:
#### What's this PR do?
ETL AWS creds on the analytics instance were not schedule to renew automatically. This PR adds a renewal schedule for those credentials to any instance with the `analytics` role.


blarghmatey:
:+1: 
Should we change the role to be `edx-analytics` or something like that to hedge against the case where we have another `analytics` style role?

shaidar:
Maybe even go with `edx-residential-analytics` as `edx` is fairly broad too. What do u think?

blarghmatey:
Sounds good to me

shaidar:
Gonna merge this and create a new PR to change the naming of the analytics role. Just feels like a cleaner approach instead of mixing it up with this.

PR #767, opened 2019-03-11
Renamed analytics role to edx-residential-analytics

shaidar:
#### What's this PR do?
Renames `analytics` role to `edx-residential-analytics` in order to clarify that exact type of analytics instance it is since we might have others in the future.


blarghmatey:
:+1:

PR #770, opened 2019-03-13
xPro heroku config vars

shaidar:
#### What's this PR do?
Configures the Heroku config variables defined [here](https://github.com/mitodl/mitxpro/blob/master/app.json) for the three app instances. We probably will need to add/modify some stuff that only applies to the `production` app, but that can be handled once the app is created.


pillar/heroku/xpro.sls
```diff
@@ -0,0 +1,38 @@
...
+    MITXPRO_FROM_EMAIL: 'MIT xPro <xpro-{{ env }}-support@mit.edu>'
+    MITXPRO_LOG_LEVEL: INFO
+    MITXPRO_SECURE_SSL_REDIRECT: True
+    MITXPRO_SUPPORT_EMAIL: 'xpro-{{ env }}-support@mit.edu'
+    MITXPRO_USE_S3: True
+    NODE_MODULES_CACHE: False
+    PGBOUNCER_DEFAULT_POOL_SIZE: 50
+    PGBOUNCER_MIN_POOL_SIZE: 5
+    SENTRY_DSN: __vault__::secret-operations/global/xpro/sentry-dsn>data>value
+    STATUS_TOKEN: {{ status_token }}
```
blarghmatey: This will be a difderent value each time we render the pillar. It would probably be better to use the `__vault__:gen_if_missing:` construct here

pillar/heroku/xpro.sls
```diff
@@ -0,0 +1,38 @@
...
+    MITXPRO_DB_CONN_MAX_AGE: 0
+    MITXPRO_DB_DISABLE_SSL: True
+    MITXPRO_EMAIL_HOST: __vault__::secret-operations/global/mit-smtp>data>relay_host
+    MITXPRO_EMAIL_PASSWORD: __vault__::secret-operations/global/mit-smtp>data>relay_password
+    MITXPRO_EMAIL_PORT: 587
+    MITXPRO_EMAIL_TLS: True
+    MITXPRO_EMAIL_USER: __vault__::secret-operations/global/mit-smtp>data>relay_username
+    MITXPRO_ENVIRONMENT: ci
+    MITXPRO_FROM_EMAIL: 'MIT xPro <xpro-{{ env }}-support@mit.edu>'
+    MITXPRO_LOG_LEVEL: INFO
```
blarghmatey: This would be a good candidate for an env dictionary

pillar/heroku/xpro.sls
```diff
@@ -0,0 +1,38 @@
...
+    MAILGUN_FROM_EMAIL: 'MIT xPRO <no-reply@xpro-{{ env }}-mail.odl.mit.edu>'
+    MITXPRO_ADMIN_EMAIL: 'cuddle-bunnies@mit.edu'
+    MITXPRO_DB_CONN_MAX_AGE: 0
+    MITXPRO_DB_DISABLE_SSL: True
+    MITXPRO_EMAIL_HOST: __vault__::secret-operations/global/mit-smtp>data>relay_host
+    MITXPRO_EMAIL_PASSWORD: __vault__::secret-operations/global/mit-smtp>data>relay_password
+    MITXPRO_EMAIL_PORT: 587
+    MITXPRO_EMAIL_TLS: True
+    MITXPRO_EMAIL_USER: __vault__::secret-operations/global/mit-smtp>data>relay_username
+    MITXPRO_ENVIRONMENT: ci
```
blarghmatey: Should be templated with the env

pillar/heroku/xpro.sls
```diff
@@ -0,0 +1,38 @@
...
+  config_vars:
+    AWS_ACCESS_KEY_ID:  __vault__:cache:aws-mitx/creds/xpro-{{ env }}>data>access_key
+    AWS_SECRET_ACCESS_KEY: __vault__:cache:aws-mitx/creds/xpro-{{ env }}>data>secret_key
+    AWS_STORAGE_BUCKET_NAME: 'xpro-{{ env }}'
+    MAILGUN_URL: 'https://api.mailgun.net/v3/xpro-{{ env }}-mail.odl.mit.edu'
+    MAILGUN_KEY: __vault__::secret-operations/global/mailgun-api-key>data>value
+    MAILGUN_FROM_EMAIL: 'MIT xPRO <no-reply@xpro-{{ env }}-mail.odl.mit.edu>'
+    MITXPRO_ADMIN_EMAIL: 'cuddle-bunnies@mit.edu'
+    MITXPRO_DB_CONN_MAX_AGE: 0
+    MITXPRO_DB_DISABLE_SSL: True
```
blarghmatey: Probably worth adding a comment to describe the reasoning for this as I described on slack

pillar/heroku/xpro.sls
```diff
@@ -0,0 +1,38 @@
...
+proxy:
+  proxytype: heroku
+
+heroku:
+  {% for env in environments %}
+  app_name: {{ app_name }}-{{ env }}
+  api_key: __vault__::secret-operations/global/heroku-api-key>data>value
+  config_vars:
+    AWS_ACCESS_KEY_ID:  __vault__:cache:aws-mitx/creds/xpro-{{ env }}>data>access_key
+    AWS_SECRET_ACCESS_KEY: __vault__:cache:aws-mitx/creds/xpro-{{ env }}>data>secret_key
```
blarghmatey: Is this intended to be a role for xpro to use multiple services, or just for S3? If it's multiple services we will need to define the role in the vault pillar.

pillar/heroku/xpro.sls
```diff
@@ -1,34 +1,54 @@
...
+      }
+} %}
 
 proxy:
   proxytype: heroku
 
 heroku:
-  {% for env in environments %}
-  app_name: {{ app_name }}-{{ env }}
+  {% for env_data in env_dict %}
```
blarghmatey: should this be an iteration over the environment, or should it pull an environment grain from the proxy minion to populate the appropriate values?

pillar/heroku/xpro.sls
```diff
@@ -1,34 +1,54 @@
 {% set app_name = 'xpro' %}
-{% set environments = ['ci', 'rc', 'production'] %}
-{% set status_token = salt.random.get_str(24) %}
+{% set status_token = __vault__:gen_if_missing:secret-operations/global/xpro/django-status-token>data>value %}
```
blarghmatey: If you're going to set it up here, then it will need to be passed as a string to then get rendered into the YAML, which then gets passed through the Vault renderer. As it is right now it's going to try to execute the `__vault__...` as a function call which will fail.

pillar/vault/roles/xpro.sls
```diff
@@ -0,0 +1,26 @@
+vault:
+  roles:
+    xpro-app:
+      backend: postgresql-xpro
```
blarghmatey: The backend would be named differently, along the lines of `postgres-production-apps-xpro` based on the `orchestrate.aws.rds` script.

pillar/vault/roles/xpro.sls
```diff
@@ -0,0 +1,26 @@
...
+      name: app
+      options:
+        {% raw %}
+        sql: >-
+          CREATE USER "{{name}}" WITH PASSWORD '{{password}}' VALID UNTIL '{{expiration}}';
+          GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO "{{name}}";
+          GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public TO "{{name}}";
+          ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT ALL PRIVILEGES ON TABLES TO "{{name}}";
+          ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT ALL PRIVILEGES ON SEQUENCES TO "{{name}}";
+        {% endraw %}
```
blarghmatey: You should copy the creation and revocation SQL from the `apps.sls` file for how the PostGres data is structured there. You could probably even just add an additional parameter to the loop that is already in there.

blarghmatey:
:+1:

PR #774, opened 2019-03-18
Add condition to edX registration redirect

### LINKED ISSUE #753: Fix bug in CSRF token for registration page on xpro QA environment

markbreedlove:

fixes #753

#### What's this PR do?

In Nginx configuration for edX, only redirect to login page for
registration if we're running the residential version.

Deploy to QA and verify (@shaidar would this be correct?)

### DIFF
--- a/salt/edx/files/nginx_static_assets.j2
+++ b/salt/edx/files/nginx_static_assets.j2
@@ -26,7 +26,9 @@
         expires epoch;
     }
 
+{% if salt.grains.get('business_unit') == 'mitxpro' %}
     # Redirect registration page to authentication page
     location /register {
         return 301 https://$host/login;
     }
+{% endif %}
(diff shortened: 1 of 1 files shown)


salt/edx/files/nginx_static_assets.j2
```diff
@@ -26,7 +26,9 @@
         expires epoch;
     }
 
+{% if salt.grains.get('business_unit') == 'mitxpro' %}
```
shaidar: I believe this should be `residential` as opposed to `mitxpro` as we only want that redirect to apply to those instances.

salt/edx/files/nginx_static_assets.j2
```diff
@@ -26,7 +26,9 @@
...
     # Redirect registration page to authentication page
     location /register {
         return 301 https://$host/login;
     }
+{% endif %}
```
shaidar: Looks like a missing new line

shaidar:
👍

PR #776, opened 2019-03-20
Fixed name of production host and directory

shaidar:
#### What's this PR do?
Came across an issue in the OCW AWS environment where `engines.conf` had a misconfigured entry. Made the change locally and submitting this to reflect that change.


blarghmatey:
:+1:

PR #778, opened 2019-03-21
Changes to add an mitxpro openedx sandbox

shaidar:
Part of mitxpro [Issue#32](https://github.com/mitodl/mitxpro/issues/32)

#### What's this PR do?
- Removed the edx-sandbox-formula which we haven't used in quite some time
- Adds mitxpro-qa environment to the sandbox role in the top file
- Adds necessary definitions to the `environment_settings.yml` for the sandbox under mitxpro-qa

### DIFF
--- a/salt/environment_settings.yml
+++ b/salt/environment_settings.yml
@@ -102,6 +102,14 @@ sandbox_residential_versions: &sandbox_residential_versions
   codename: hawthorn
   ami_id: ami-80861296
 
+sandbox_mitxpro_versions: &sandbox_mitxpro_versions
+  edx_config_repo: https://github.com/edx/configuration.git
+  edx_config_version: oopen-release/ironwood.master
+  edx_platform_repo: 'https://github.com/edx/edx-platform'
+  edxapp: open-release/ironwood.master
+  codename: ironwood
+  ami_id: ami-80861296
+
 environments:
   mitxpro-qa:
     business_unit: mitxpro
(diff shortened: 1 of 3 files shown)


salt/environment_settings.yml
```diff
@@ -197,6 +205,19 @@ environments:
           edx-worker:
             number: 1
             type: t2.large
+      sandbox:
```
blarghmatey: Should this be designated as "sandbox" or as "mitxpro-ci"? Not sure if that impacts how it will function as far as deployment, but semantically this is technically for the CI app so it might be useful and less confusing in the future to make sure the name reflects that.

salt/environment_settings.yml
```diff
@@ -197,6 +205,19 @@ environments:
...
+        app: edx
+        business_unit: mitxpro
+        domains:
+          cms: studio-xpro-sandbox.mitx.mit.edu
+          lms: xpro-qa-sandbox.mitx.mit.edu
```
blarghmatey: Should we set this as `xpro-ci`?

salt/environment_settings.yml
shaidar: I kinda like this mapping:
ci - sandbox
rc - qa
production (or no env) - production (or no env)

Besides we already have the discrepancy between rc/qa.

blarghmatey: Fair enough

blarghmatey:
:+1:

Issue #761, opened 2019-03-01
Upgrade Elasticsearch logging cluster to 6.x release


blarghmatey:
We are currently running on the 5.x release of the Elastic Stack and we would like to upgrade to the latest release in the 6.x series. This includes:

- [x] reformatting our Pillar data to match the structure defined by the new elastic-stack-formula
- [x] upgrading/redeploying our Kibana instance
- [x] Migrate Kibana Nginx configuration to use the nginx formula
- [x] Distribute the logging cluster across availability zones (it is currently a 5 node cluster all in one AZ)

markbreedlove:
The cluster has a large number of closed indices. For example, `logstash-heroku-2017.12.19`. It's not directly related to the upgrade (though maybe if mapping syntax has changed?) Do you think we should drop them to free up space in the process and to reduce complexity?

blarghmatey:
It doesn't hurt, but it's not critical. In the end, those instances will be terminated anyway so the disk space issue isn't directly relevant here. (I'm assuming that the closed instances aren't included in the replication)

markbreedlove:
@blarghmatey It's my understanding that I'll need to create a new pillar for `logging.sls` under `elastic_stack`.  I assume I'll base it on the one in `elasticsearch`. I'm noticing that  https://github.com/mitodl/salt-ops/blob/master/pillar/elasticsearch/logging.sls#L56 matches the EC2 `tag:role` == `elasticsearch`. Is that going to be an issue if there are other ES clusters with that `role` tag (micromasters, MITx, etc)?

markbreedlove:
@blarghmatey Also in the existing `logging.sls` (https://github.com/mitodl/salt-ops/blob/master/pillar/elasticsearch/logging.sls#L49) the top-level property `elasticsearch` exists, but I'm noticing that https://github.com/mitodl/elastic-stack-formula/blob/28652485540ab89f145188ebd4744f71a1236fcd/elastic-stack/elasticsearch/map.jinja#L34 references `elastic_stack:elasticsearch`. Am I correct to assume I should nest `elasticsearch` under `elastic_stack`?

markbreedlove:
@blarghmatey updated https://github.com/mitodl/salt-ops/issues/761#issuecomment-473357636 regarding cluster naming.

blarghmatey:
[...]

PR #779, opened 2019-03-22
Logging elastic stack upgrade

markbreedlove:

#761 

#### What's this PR do?

* Adds pillars that work with new `elastic_stack` formula
* Prepares for the new way in which index templates will be updated with the `elastic_stack` formula, though using the feature is on hold until a bug is addressed.

The `elastic_stack` formula has been tested with pillar data similar to what appears in `salt-ops`. Orchestration may be run with `test=True` as a first pass. This upgrade is not high risk.


markbreedlove:
@blarghmatey Do you think I should remove `pillar/elasticsearch` now that `pillar/elastic_stack` has superceded it?

blarghmatey:
No, because we still have to use the old formula for our edX environments. They are hard locked to v1.x and we can't upgrade in the foreseeable future, so we have to keep that configuration around. It's feasible to delete the logging one though.

Also, before we merge this we'll need to define pillar settings in the elastic_stack format for Kibana and Elastalert.

blarghmatey:
No, because we still have to use the old formula for our edX environments. They are hard locked to v1.x and we can't upgrade in the foreseeable future, so we have to keep that configuration around. It's feasible to delete the logging one though.

Also, before we merge this we'll need to define pillar settings in the elastic_stack format for Kibana and Elastalert.

pillar/top.sls
```diff
@@ -16,7 +16,8 @@ base:
     - datadog.elasticsearch-integration
   'roles:kibana':
     - match: grain
-    - kibana
+    - mitca
+    - elastic_stack.kibana
     - elastalert
```
shaidar: I believe this should also have `nginx.kibana`

pillar/nginx/kibana.sls
```diff
@@ -0,0 +1,77 @@
...
+{% set server_domain_names = env_data.purposes[app_name].domains %}
+
+nginx:
+  ng:
+
+    certificates:
+      odl.mit.edu:
+        public_cert: __vault__::secret-operations/global/odl_wildcard_cert>data>value
+        private_key: __vault__::secret-operations/global/odl_wildcard_cert>data>key
+
```
shaidar: nit-pick: Not sure the blank line here and line 9 serve anything.

salt/top.sls
```diff
@@ -133,8 +134,8 @@ base:
     - datadog.plugins
   'G@roles:elasticsearch and P@environment:(micromasters|rc-apps|production-apps)':
     - match: compound
-    - elasticsearch
-    - elasticsearch.plugins
+    - elastic_stack.elasticsearch
+    - elastic_stack.elasticsearch.plugins
```
blarghmatey: These two lines should remain unmodified since that is for a separate cluster which is currently deployed using the old formula.

pillar/top.sls
```diff
@@ -16,7 +16,9 @@ base:
     - datadog.elasticsearch-integration
   'roles:kibana':
     - match: grain
-    - kibana
+    - mitca
+    - elastic_stack.kibana
+    - nginx.kibana
```
blarghmatey: This should also have the `nginx` pillar assigned to pull in the logging configuration. So it would be:
```
- nginx
- nginx.kibana
```

salt/top.sls
```diff
@@ -134,8 +134,8 @@ base:
     - datadog.plugins
   'G@roles:elasticsearch and P@environment:(micromasters|rc-apps|production-apps)':
     - match: compound
-    - elastic_stack.elasticsearch
-    - elastic_stack.elasticsearch.plugins
+    - elasticsearch.elasticsearch
+    - elasticsearch.elasticsearch.plugins
```
blarghmatey: This should just be:
```
- elasticsearch
- elasticsearch.plugins
```

markbreedlove: Oops. I've pushed a fix.

pillar/elastic_stack/elasticsearch/logging.sls
```diff
@@ -0,0 +1,56 @@
...
+    #       mappings:
+    #         fluentd:
+    #           dynamic_templates:
+    #             - strings:
+    #                 match_mapping_type: string
+    #                 mapping:
+    #                   type: text
+    #                   fields:
+    #                     raw:
+    #                       type: keyword
```
blarghmatey: So, I know that the `not analyzed` setting allows us to have a `.raw` attribute for searching items that aren't tokenized. What does the `keyword` type provide us with?

markbreedlove: `keyword` essentially means not analyzed / tokenized; the whole value of that field is considered as one keyword.

They changed their mapping syntax from version <6 to 6. Ran into this on my last job ...

markbreedlove: I should have mentioned my motivation for changing this was actually loading data into a test ES6 index on my Docker container and running into mapping errors w the previous iteration of the mapping.

blarghmatey:
:+1:

PR #782, opened 2019-03-27
Improve Elasticsearch cloud map and profile generation

markbreedlove:

https://github.com/mitodl/salt-ops/issues/761

#### What's this PR do?

* Create a service-per-environment cloud map file instead of one
  "elasticsearch" cloud map file.

* Remove profile_overrides from orchestrate/services/elasticsearch.sls
  and set environment-specific data from environment_settings.yml

This should fix the issues we were having with trying to create the operations logging Elasticsearch cluster and having settings like instance type and volume size get overridden by the hardcoded `profile_overrides` in `salt/orchestrate/services/elasticsearch.sls`. In addition I'm attempting to do the refactoring that I _think_ we three were talking about yesterday to make the Elasticsearch Salt Cloud files more flexible and easier to override per-environment.

### DIFF
--- a/salt/environment_settings.yml
+++ b/salt/environment_settings.yml
@@ -539,6 +539,10 @@ environments:
         business_unit: operations
         security_groups:
           - webapp-odl-vpn
+      elasticsearch:
+        num_instances: 5
+        size: r4.xlarge
+        data_volume_size: 1536
     backends:
       pki:
         - consul
--- a/salt/orchestrate/services/elasticsearch.sls
+++ b/salt/orchestrate/services/elasticsearch.sls
@@ -1,14 +1,16 @@
 {% set env_settings = salt.cp.get_file_str("salt://environment_settings.yml")|load_yaml %}
 {% set ENVIRONMENT = salt.environ.get('ENVIRONMENT', 'rc-apps') %}
 {% set env_data = env_settings.environments[ENVIRONMENT] %}
+{% set purpose_data = env_data.purposes.elasticsearch|default({}) %}
 {% set VPC_NAME = env_data.vpc_name %}
-{% set INSTANCE_COUNT = salt.environ.get('INSTANCE_COUNT', 3) %}
+{% set INSTANCE_COUNT = salt.environ.get('INSTANCE_COUNT', purpose_data.get(num_instances, 3) %}
 {% set BUSINESS_UNIT = salt.environ.get('BUSINESS_UNIT', env_data.business_unit) %}
 {% set launch_date = salt.status.time(format="%Y-%m-%d") %}
 {% set subnet_ids = salt.boto_vpc.describe_subnets(
     vpc_id=salt.boto_vpc.describe_vpcs(
         name=env_data.vpc_name).vpcs[0].id
     ).subnets|map(attribute='id')|list %}
+
 load_elasticsearch_cloud_profile:
   file.managed:
     - name: /etc/salt/cloud.profiles.d/elasticsearch.conf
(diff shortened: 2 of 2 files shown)


blarghmatey:
So, with the approach that you took here, there's no need to turn the profile into a template, you can just modify the context data that is being passed to the map template and it will have the same effect.

An alternate approach which we don't have to worry about tackling right now is to create a separate cloud profile that inherits from the existing elasticsearch one and specifies the overrides for a given environment or use case and then ensuring that we render that to disk as part of the deployment.

For the purpose of this PR and to get this out the door let's just stick with passing those attributes from the `environment_settings.yml` through the context to the map template.

markbreedlove:
@blarghmatey Thanks for your feedback. I've pushed https://github.com/mitodl/salt-ops/pull/782/commits/ae206dfac911eca578aed29b90df35ad57080698

salt/orchestrate/services/elasticsearch.sls
```diff
@@ -1,14 +1,17 @@
...
 {% set VPC_NAME = env_data.vpc_name %}
-{% set INSTANCE_COUNT = salt.environ.get('INSTANCE_COUNT', 3) %}
+{% set INSTANCE_COUNT = salt.environ.get('INSTANCE_COUNT', purpose_data.num_instances|default(3)) %}
 {% set BUSINESS_UNIT = salt.environ.get('BUSINESS_UNIT', env_data.business_unit) %}
 {% set launch_date = salt.status.time(format="%Y-%m-%d") %}
 {% set subnet_ids = salt.boto_vpc.describe_subnets(
     vpc_id=salt.boto_vpc.describe_vpcs(
         name=env_data.vpc_name).vpcs[0].id
     ).subnets|map(attribute='id')|list %}
+{% set service_name = "{{ ENVIRONMENT }}-elasticsearch" %}
```
markbreedlove: @blarghmatey Is this going to have any negative effect? I believe that the `profile_name` in https://github.com/mitodl/salt-ops/blob/746e137e32b665c860fcb1778c1e5dcb270ffc63/salt/orchestrate/aws/map_templates/instance_map.yml#L6 is not going to match the top-level property in https://github.com/mitodl/salt-ops/blob/746e137e32b665c860fcb1778c1e5dcb270ffc63/salt/orchestrate/aws/cloud_profiles/elasticsearch.conf#L2

blarghmatey: Correct, that won't match. For this PR just go ahead and get rid of the service_name override and just restrict the changes to making the instance size, disk size, etc pull from environment settings for generating the cloud map file. Don't change the map template or the profile config.

markbreedlove: Ok! I've pushed https://github.com/mitodl/salt-ops/pull/782/commits/a432876c6cba1cfe701498955cbfb103ac748750

`instance_map.yml` and `elasticsearch.conf` now have no differences vs `master`.

salt/orchestrate/services/elasticsearch.sls
```diff
@@ -1,14 +1,16 @@
 {% set env_settings = salt.cp.get_file_str("salt://environment_settings.yml")|load_yaml %}
 {% set ENVIRONMENT = salt.environ.get('ENVIRONMENT', 'rc-apps') %}
 {% set env_data = env_settings.environments[ENVIRONMENT] %}
+{% set purpose_data = env_data.purposes.elasticsearch|default({}) %}
 {% set VPC_NAME = env_data.vpc_name %}
-{% set INSTANCE_COUNT = salt.environ.get('INSTANCE_COUNT', 3) %}
+{% set INSTANCE_COUNT = salt.environ.get('INSTANCE_COUNT', purpose_data.num_instances|default(3)) %}
```
blarghmatey: Instead of using the Jinja default filter I would just do `purpose_data.get('num_instances', 3)`

salt/environment_settings.yml
```diff
@@ -539,6 +539,10 @@ environments:
         business_unit: operations
         security_groups:
           - webapp-odl-vpn
+      elasticsearch:
+        num_instances: 5
+        size: r4.xlarge
+        data_volume_size: 800
```
blarghmatey: We should probably set this to 1 or 1.5 TB since we will be starting to throw metric data at this cluster.

salt/orchestrate/services/elasticsearch.sls
markbreedlove: ✅

salt/environment_settings.yml
markbreedlove: ✅

markbreedlove:
I'll go ahead and merge this in a little bit if there aren't any more changes to be made, and then I'll try another orchestration run.

blarghmatey:
:+1:

PR #784, opened 2019-03-28
Added forum configs to sandbox

shaidar:
[Issue#32](https://github.com/mitodl/mitxpro/issues/32)

#### What's this PR do?
Adds ansible_vars for the forum service to run on the sandbox instance

### DIFF
--- a/pillar/edx/sandbox.sls
+++ b/pillar/edx/sandbox.sls
@@ -4,6 +4,7 @@
 {% set purpose_suffix = purpose.replace('-', '_') %}
 {% set purpose_data = env_data.purposes[purpose] %}
 {% set environment = salt.grains.get('environment', 'mitx-qa') %}
+{% set business_unit = purpose_data.business_unit %}
 {% set MYSQL_HOST = 'mysql.service.consul' %}
 {% set MYSQL_PORT = 3306 %}
 {% set MONGODB_HOST = 'mongodb-master.service.consul' %}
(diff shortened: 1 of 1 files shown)


pillar/edx/sandbox.sls
```diff
@@ -28,6 +32,24 @@ edx:
     EDXAPP_MYSQL_CSMH_USER: __vault__:cache:mysql-{{ environment }}/creds/edxapp-csmh-{{ purpose }}>data>username
     EDXAPP_MEMCACHE:
       - 'localhost:11211'
+    ################################################################################
+    #################### Forum Settings ############################################
+    ################################################################################
+    FORUM_API_KEY: __vault__:gen_if_missing:secret-residential/global/forum-api-key>data>value
```
blarghmatey: This technically isn't residential so we should probably use a different secret mount (`secret-mitxpro`). It might also be worth thinking about moving to a per-env key but not necessary at the moment.

As an aside, we should check if we're using the residential mount for other secrets in the xpro env.

shaidar: We also use it in the `cloud_deployment` pillar which applies to the xpro env

blarghmatey: Yeah, I thought that might be the case... We can change that in a separate PR

pillar/edx/sandbox.sls
```diff
@@ -35,7 +35,7 @@ edx:
     ################################################################################
     #################### Forum Settings ############################################
     ################################################################################
-    FORUM_API_KEY: __vault__:gen_if_missing:secret-residential/global/forum-api-key>data>value
+    FORUM_API_KEY: __vault__:gen_if_missing:secret-mitxpro/global/forum-api-key>data>value
```
blarghmatey: We should pull the business unit from the env data for this so that we don't have to hardcode it per env.

blarghmatey:
:+1:

PR #785, opened 2019-04-01
Cleaning up after fixing AMI build errors

blarghmatey:
Removing code that was added while debugging AMI build failures for residential production that was a result of the asset compilation step causing the master to timeout waiting for a response from the minion

### DIFF
--- a/salt/orchestrate/edx/build_ami.sls
+++ b/salt/orchestrate/edx/build_ami.sls
@@ -215,9 +215,6 @@ compile_assets_for_edx_{{ PURPOSE }}:
     - name: cmd.run
     - arg:
         - /edx/bin/edxapp-update-assets
-    - kwarg:
-        use_vt: True
-        timeout: 900
     - require:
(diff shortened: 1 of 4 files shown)


salt/orchestrate/edx/build_ami.sls
```diff
@@ -215,9 +215,6 @@ compile_assets_for_edx_{{ PURPOSE }}:
     - name: cmd.run
     - arg:
         - /edx/bin/edxapp-update-assets
-    - kwarg:
-        use_vt: True
```
shaidar: What does this do?

blarghmatey: That's a flag that is supposed to allow for streaming output back to the master rather than waiting until the end to see output from `cmd.run`. I was testing it to see if it would alleviate the timeouts but it had no effect so there's no reason to keep it.

shaidar:
👍

PR #793, opened 2019-04-04
Restart salt-master serivce when low memory

shaidar:
#### What's this PR do?
Restart `salt-master` service when low memory alert is triggered by adding a beacon to the master. Currently an alert is triggered through Datadog, however we should disable that once this is deployed/

### DIFF
--- a/pillar/beacons/low_memory.sls
+++ b/pillar/beacons/low_memory.sls
@@ -1,0 +1,3 @@
+beacons:
+  memusage:
+    - percent: 95%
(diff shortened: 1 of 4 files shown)


blarghmatey:
I would argue against this approach due to the fact that the low memory conditions often arise as a result of executing a long-running orchestrate routine which would be interrupted if the master were to restart. If we _do_ want to automatically restart in low memory situations then we should consider using the Thorium reactor system and only restart if the low memory situation persists over a certain time range https://docs.saltstack.com/en/latest/topics/thorium/index.html

shaidar:
`Thorium` appears to be an experimental feature and has been since 2016. Just doesn't inspire a lot of confidence. I see your point around causing issues restarting in the middle of a running routine, but feel that most of the alerts generated have typically been during off-hours and I'm assuming caused by a slow memory leak somewhere that eventually triggers that alert. I would think that if an orchestrate routine is causing that alert then maybe we should consider changing the instance type to avert that situation.

PR #794, opened 2019-04-08
Fixes for xpro

shaidar:
#### What's this PR do?
- Adds the `SKIP_EMAIL_VALIDATION` feature given that enabling that on the oauth provider level didn't provide the expected result
- Used the `business_unit` value to replace the hard-coded residential value since those pillar values are applied to xpro as well


blarghmatey:
:+1:

PR #808, opened 2019-04-16
 Add OCW "engines" deployment

markbreedlove:
Add states that deploy the "engine" scripts that are responsible for various job-like functionality in OCW. These are found in [the `publishing` directory of `ocwcms`](https://github.com/mitocw/ocwcms/tree/master/publishing)

* Manage cron jobs
* Finish work begun earlier to templatize `engines.conf`
* Add deployment functionality to origin servers
* Symlink directories that are referred to by scripts. We have to leave the paths unaltered for now in the event that one gets copied back to the "current production" legacy system.

At this time of writing, the scripts that are being deployed are on the `add-missing-files` branch of `ocwcms` and pull request in  progress to merge them in: https://github.com/mitocw/ocwcms/pull/63


pillar/apps/ocw-production.sls
```diff
@@ -0,0 +1,45 @@
...
+      dspace_connection_user: not configured
+      dspace_connection_password: not configured
+      dspace_endpoint_prefix: http://dome-test.mit.edu/sword/deposit/
+      deposit_uri_lookup_file_name: UAT_DepositURILookup.xml
+    # production_dspace_conf is for Production_Dspace_conf in engines.conf.jinja
+    production_dspace_conf:
+      dspace_host_id: dspace.mit.edu
+      dspace_host_port: 80
+      dspace_connection_user: ocw-deposit@mit.edu
+      dspace_connection_password: 0cwr0x0r
```
markbreedlove: TODO: I'll pull this in from Vault...

pillar/apps/ocw-production.sls
```diff
@@ -0,0 +1,45 @@
+ocw:
+  engines_conf:
+    database:
+      host: "10.100.0.7"
```
blarghmatey: It would be good to pull this in dynamically rather than hardcode IP addresses. Most likely using the Mine subsystem. You can refer to https://github.com/mitodl/salt-ops/blob/xpro_production_env/pillar/mongodb/init.sls#L7-L11 for an example of using the salt mine

pillar/apps/ocw-qa.sls
```diff
@@ -0,0 +1,45 @@
+ocw:
+  engines_conf:
+    database:
+      host: "10.100.0.34"
```
blarghmatey: As above, it would be good to avoid hardcoding IP addresses.

salt/apps/ocw/engines.sls
```diff
@@ -0,0 +1,72 @@
...
+  file.managed:
+    - name: {{ engines_basedir }}/engines.conf
+    - template: jinja
+    - source: salt://apps/ocw/templates/engines.conf.jinja
+    - user: ocwuser
+    - group: ocwuser
+    - mode: 0640
+
+manage_generate_ocw_news_feeds_cronjob:
+  file.managed:
```
blarghmatey: We can use the cron state module for these. https://docs.saltstack.com/en/latest/ref/states/all/salt.states.cron.html#module-salt.states.cron

salt/apps/ocw/symlinks_cms.sls
```diff
@@ -0,0 +1,82 @@
...
+    - name: /prod_new
+    - target: /mnt/ocwfileshare
+    - force: True
+    - user: root
+    - group: root
+
+symlink_prod:
+  file.symlink:
+    - name: /prod
+    - target: /prod_new
```
blarghmatey: Why are we symlinking to a symlink?

salt/apps/ocw/symlinks_cms.sls
```diff
@@ -0,0 +1,82 @@
...
+    - target: /mnt/ocwfileshare
+    - force: True
+    - makedirs: True
+    - owner: root
+    - group: root
+
+symlink_ocwuser_ocwengines:
+  file.symlink:
+    - name: /home/ocwuser/OCWEngines
+    - target: /mnt/ocwfileshare/OCWEngines
```
blarghmatey: Is this actually necessary? Are there any cases where that code is being used from the ocwuser home directory? If not I'd prefer to avoid muddying the waters further...

salt/apps/ocw/symlinks_origin.sls
```diff
@@ -0,0 +1,18 @@
+
```
blarghmatey: It's probably worth trying to move the files that we're serving to a more sane directory structure (e.g. /var/www/ocw) and then symlinking that to the weird tomcat directory that we're stuck with so that it's easier to factor that out to the sane target down the road.

salt/apps/ocw/sync_repo.sls
```diff
@@ -37,17 +33,31 @@ git_pull_ocwcms_working_copy:
...
+{% if salt['file.directory_exists']('/usr/local/Plone') %}
 ensure_state_of_src_symlink:
   file.symlink:
     - name: /usr/local/Plone/zeocluster/src
     - target: /var/lib/ocwcms/plone/src
     - force: True
     - backupname: src_old
+{% endif %}
+
+sync_ocwcms_web_directory:
```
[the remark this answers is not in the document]
markbreedlove: Changed in 82e7b22

pillar/apps/ocw-production.sls
markbreedlove: I've attempted to fix this in 8679d5c
Does it look OK?

pillar/apps/ocw-production.sls
```diff
@@ -1,7 +1,19 @@
+
+{% set dbhost_ipaddr_data =
+     salt.saltutil.runner('mine.get',
+                          tgt='G@roles:ocw-db and G@environment:ocw-production',
+                          fun='network.ip_addrs',
+                          tgt_type='compound')
+%}
+{% set dbhost_ipaddr = dbhost_ipaddr_data[1][0] %}
```
blarghmatey: The return value of the `mine.get` is a dictionary of the form:
```
{
    "ocw-production-cms-db1": [
        "10.100.0.12"
    ], 
    "ocw-production-cms-db2": [
        "10.100.0.7"
    ]
}
```
So the first index should be a string to retrieve the proper instance.

salt/apps/ocw/engines.sls
markbreedlove: I've changed this in 0b3b819

salt/apps/ocw/symlinks_cms.sls
markbreedlove: Not that any of this makes a lot of sense, but I was following the library paradigm -- for example in some systems' `/usr/lib` -- where there are layers of symlinks from more general to more specific. But it's not really necessary here, I acknowledge. I'll change it.

salt/apps/ocw/symlinks_origin.sls
markbreedlove: If I understand you correctly, that's what's being done [here](https://github.com/mitodl/salt-ops/blob/ab25652e98005bed2072fdd522fc0b5cf95e4f88/salt/apps/ocw/sync_repo.sls#L46-L59).

These symlinks are only being created to allow this script to function: https://github.com/mitocw/ocwcms/blob/add-missing-files/publishing/sitemap.sh
That script runs on the origin server, and is invoked by https://github.com/mitocw/ocwcms/blob/add-missing-files/publishing/run_aka_scripts.sh on the cms2 engines server. We just want to preserve the crazy filesystem locations for the time being in case files get copied back to the legacy servers.

pillar/apps/ocw-production.sls
markbreedlove: OK! Does commit 88f8b53 look any better?

blarghmatey:
:+1:

PR #813, opened 2019-04-17
Added heroku xpro-production instance

shaidar:
#### What's this PR do?
Added needed keys and values to `environment_settings.yml` for the Heroku RDS xpro db

### DIFF
--- a/salt/environment_settings.yml
+++ b/salt/environment_settings.yml
@@ -716,6 +716,13 @@ environments:
         - fluentd
         - rabbitmq
       rds:
+        - name: mitxpro
+          multi_az: True
+          engine: postgres
+          vault_plugin: postgresql-database-plugin
+          db_instance_class: db.t2.medium
+          allocated_storage: 100
+          public_acces: False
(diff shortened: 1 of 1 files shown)


salt/environment_settings.yml
```diff
@@ -716,6 +716,13 @@ environments:
         - fluentd
         - rabbitmq
       rds:
+        - name: mitxpro
+          multi_az: True
+          engine: postgres
+          vault_plugin: postgresql-database-plugin
+          db_instance_class: db.t2.medium
```
blarghmatey: I would steer clear of any t2 instance class. Probably worth just going with an M5

salt/environment_settings.yml
```diff
@@ -716,6 +716,13 @@ environments:
...
+          allocated_storage: 100
+          public_acces: False
```
blarghmatey: This will require public_access to be True in order for Heroku to communicate with it

blarghmatey:
:+1:

PR #817, opened 2019-04-18
Make OCW origin webserver show 404 error page

markbreedlove:
Fixes https://github.com/mitocw/ocwcms/issues/61

#### What's this PR do?

Make the OCW origin server show the custom 404 error page that exists in
/var/www/ocw/jsp/error.html. This is the HTML that's currently being
delivered for https://ocw.mit.edu/mbtest/not_found on the current
production site.

Has to be tested after deployment. See https://github.com/mitocw/ocwcms/issues/61

### DIFF
--- a/pillar/nginx/ocw_origin.sls
+++ b/pillar/nginx/ocw_origin.sls
@@ -66,3 +66,6 @@ nginx:
                         - index.htm
                         - index.html
                         - =404
+                    - error_page:
+                        - '404'
+                        - jsp/error.html


pillar/nginx/ocw_origin.sls
```diff
@@ -65,4 +65,6 @@ nginx:
                         - $uri/
                         - index.htm
                         - index.html
-                        - =404
```
blarghmatey: This line doesn't need to be removed, that will just direct Nginx to return a 404 response when no files are found. The config that you added is what will handle determining what gets loaded for that response.

blarghmatey: We don't want to remove this line as that is what will return the 404 if the file isn't found in the other locations.

markbreedlove: OK. Commit 0204d85 puts it back.

blarghmatey:
:+1:

PR #819, opened 2019-04-18
Changed instance type from t2 to t3

shaidar:
#### What's this PR do?
Changed instance type from t2 to t3 in orchestrate states and cloud_profiles. Left environment_settings intact in order to reflect current reality.


blarghmatey:
:+1:

PR #820, opened 2019-04-19
Modify environment_settings.yml, correct OCW environments

markbreedlove:
Modify environment_settings.yml to have the correct environment names for OCW. They are currently "ocw-production" and "ocw-qa".

This fits the convention in place already for mitxpro and mitx.

We've discussed possibly having QA and Production share the same environment but having some other grain that distinguishes the two, but haven't done that yet.

I ran into an error from Salt master when trying to run the `nginx.ng` state, per https://github.com/mitocw/ocwcms/issues/61#issuecomment-484937988

`SaltRenderError: Jinja variable 'dict object' has no attribute 'ocw-qa'`

#### What's this PR do?

Fixes the Salt render error by having the environments in `environment_settings.yml` that match the OCW origin servers' `environment` grains.

### DIFF
--- a/salt/environment_settings.yml
+++ b/salt/environment_settings.yml
@@ -879,8 +879,17 @@ environments:
         app: nginx
         business_unit: open-courseware
         domains:
-          - ocw-origin.odl.mit.edu
-        num_instances: 1
+          production:
+            production:
+              - ocw-origin.odl.mit.edu
+            staging:
+              - ocw-production-ocw2.odl.mit.edu
+          qa:
+            production:
+              - ocw-qa-origin.odl.mit.edu
+            staging:
+              - ocw-qa-ocw2.odl.mit.edu
+        num_instances: 4
         security_groups:
           - webapp
           - ssh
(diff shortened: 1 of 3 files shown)


salt/environment_settings.yml
```diff
@@ -884,6 +884,20 @@ environments:
         security_groups:
           - webapp
           - ssh
+  ocw-qa:
```
blarghmatey: We don't want to add a new environment definition here since it's all still in the same VPC. The environments in this file map directly to separate VPCs for network segmentation which doesn't exist for OCW. We can add the additional purpose for passing the domain names to the QA instance, but we should be matching on a different grain, such as `purpose` or adding a new one like `ocw-env` to differentiate between "QA" and "production" in the OCW context. We don't want to split them into different actual envs, we just want to keep the current architecture as is as much as possible until we cut over from production.

markbreedlove: Ok, that sounds good. I will follow up with a change along those lines. I think it will probably be to add `ocw-env`, because the property `purposes` exists (as `environments.ocw.purposes`) in `environment_settings.yml` and aligns with the existing `app_name` in the `.sls` file.

I will also add the new grain to the existing minions. I'm making a separate comment in the main thread of this PR so I don't forget.

markbreedlove:
The changes made here will require updating grains on the affected minions, per https://github.com/mitodl/salt-ops/pull/820#discussion_r277082682

PR #823, opened 2019-04-19
Removed opts arg

shaidar:
#### What's this PR do?
`opts` in mount.mounted throws the following error:
```
Passed invalid arguments to state.single: get_sls_opts() got multiple values for 
keyword argument 'opts'
```
Checking slack, I was told that this is a bug caused by a conflict with the opts keyword that is used with `state.single`. Additionally, [since linux kernel 26.30 the kernel defaults to the `relatime` option behavior](http://man7.org/linux/man-pages/man8/mount.8.html).


blarghmatey:
:+1:

PR #825, opened 2019-04-22
Added mitx to use old elasticsearch formula

shaidar:
#### What's this PR do?
MITx environments were configured to use the new elastic-stack formula which is geared towards >=5.x ES releases. However MITx, requires an older version which forces us to use the old elasticsearch-formula we have in place as opposed to the new elastic-stack one.


blarghmatey:
:+1:

PR #827, opened 2019-04-22
Updated mitx runbook and removed old state

shaidar:
#### What's this PR do?
To create `rds` instances, we now use the `orchestrate.aws.rds` state which pulls the needed information from the `enviornment_settings.yml` file. I updated the runback accordingly and to reduce confusion, deleted the old state that we had been previously using. Additionally, added step to create the Elasticache memcached nodes.


blarghmatey:
:+1:

PR #828, opened 2019-04-23
Added variable to instance role

shaidar:
#### What's this PR do?
- Updated runbook based on new AMI build and deploy
- Added an env variable to the `build_ami` state for the tracking_bucket and the edx-instance-role to avoid any conflicts between residential and xpro.


blarghmatey:
:+1:

PR #829, opened 2019-04-25
In Nginx OCW origin pillar, allow http download of edx_courses.json

markbreedlove:

https://github.com/mitocw/ocwcms/issues/34
https://github.com/mitocw/ocwcms/pull/35

#### What's this PR do?

This allows an account on our legacy "current production" origin server to pull this file on cron every day. It seems like the neatest way of doing it to work around being stuck with TLSv1 on the old server, and to avoid hassles with ssh.

It's been "manually staged" ... try running `curl -I http://ocw-origin.odl.mit.edu/courses/edx_courses.json`

### DIFF
--- a/pillar/nginx/ocw_origin.sls
+++ b/pillar/nginx/ocw_origin.sls
@@ -39,6 +39,9 @@ nginx:
                     - '[::]:80'
                 - location ~ /.well-known:
                     - allow: all
+                - location /courses/edx_courses.json:
+                    - root: /var/www/ocw
+                    - allow: all
                 - location /:
                     - return: 301 https://$host$request_uri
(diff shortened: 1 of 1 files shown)


shaidar:
Are we sure there's nothing in the file deemed somewhat sensitive and should be behind some type of auth?

markbreedlove:
I don't think so, because it's designed to be requested with Ajax by the Javascript that runs in an unauthenticated user's browser.

pdpinch:
Can we set this up on legacy QA / ocw5.mit.edu as well?

markbreedlove:
Sure, I've just set that up.

PR #830, opened 2019-04-25
Change OCW engine cron jobs to output to files

markbreedlove:
This changes the OCW engine cron jobs to send their outputs to files instead of mailing them.

They're going to individual files at the moment and each file gets clobbered. Let me know if you prefer one big file, but the output of the scripts is not formatted thoughtfully so it should be easier to tell which lines are for which job this way. For the same reason, I've spent no energy setting up fluentD, though that would be nice in the future. For now I just need to see some output to troubleshoot the jobs, and figuring out how to parse them might take some time.

In addition, I've changed the condition that determines whether to manage cron jobs to use a grain, instead of checking the existence of a filesystem path. The cron jobs should only run on the "cms-2" server, whereas both of the cms servers have the path that was being evaluated previously.

### DIFF
--- a/salt/apps/ocw/engines.sls
+++ b/salt/apps/ocw/engines.sls
@@ -2,7 +2,14 @@
 
 {% if salt['file.directory_exists'](engines_basedir) %}
 
-{% set mailto = salt.pillar.get('ocw:cron_mailto') %}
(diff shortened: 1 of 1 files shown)


blarghmatey:
:+1:

PR #831, opened 2019-04-26
Add management of Plone config file for OCW CMS

markbreedlove:
See https://github.com/mitocw/ocwcms/pull/72#issuecomment-487089348


blarghmatey:
:+1:

PR #839, opened 2019-04-30
Added social_auth_https_redirect to lms env production

shaidar:
#### What's this PR do?
- Adds `SOCIAL_AUTH_REDIRECT_IS_HTTPS` and sets it to `True` in `/edx/app/edxapp/edx-platform/lms/envs/production.py`. This is required for xpro and setting it in the oauth app config under options, didn't seem to have any result.
- Creates a `private.py` file and appends a script to it to remove the Django-toolbar based on this [link](https://openedx.atlassian.net/wiki/spaces/OXA/pages/159575149/How+can+I+remove+the+Django+Toolbar)


blarghmatey:
:+1:

PR #841, opened 2019-05-01
Added IDA_LOGOUT_URI_LIST to xpro ansible_vars

shaidar:
#### What's this PR do?
- This is a bit of an ugly hack but because of the naming discrepancy between our Heroku envs and AWS, I couldn't think of another way of dealing with it. This add the `IDA_LOGOUT_URI_LIST` value to xpro ansible_vars based on this [doc](https://github.com/mitodl/mitxpro/blob/master/docs/configure_open_edx.md).
- Updated `logout_redirect_url` based on above linked doc
- Renamed a few of the default values to reflect xpro.


blarghmatey:
:+1:

blarghmatey:
:+1:

Issue #610, opened 2018-05-21
Ironwood: Use Shibboleth auth to integrate directly with Touchstone


pdpinch:
The CAS server has been causing us a number of problems ( #408 #409 #410 ), in addition to needing an OS and django upgrade. Since we don't use CAS to authenticate any services besides Touchstone with edX, it would be simpler to integrate directly with Touchstone via Shibboleth instead. 

We should be able to use the (relatively new) 3rd party auth functionality in edX.

blarghmatey:
We are being forced to push this back another release. SAML login is functional but as part of that we have to expose public account creation with no way to restrict the available authentication mechanisms. We will need to address that shortcoming before the next release.

blarghmatey:
This is now functional in QA next, along with restricting authentication and registration via the API/registration form. We are moving forward with releasing to production during the next maintenance window.

PR #842, opened 2019-05-01
Updated settings and hacks to resolve SAML login

blarghmatey:
In order for us to be able to create new accounts for users using SAML they need to be able to self-register as part of the auth flow. This requires us to remove our forced redirect of the `/register` endpoint and set the feature flag to allow account creation. This will likely require further efforts to prevent creation of accounts by any means other than SAML.

#610 

#### What's this PR do?
Fixes account creation for edX residential

### DIFF
--- a/pillar/edx/ansible_vars/next_residential.sls
+++ b/pillar/edx/ansible_vars/next_residential.sls
@@ -9,3 +9,22 @@ edx:
     EDXAPP_SESSION_COOKIE_NAME: {{ environment }}-{{ purpose }}-session
     EDXAPP_CMS_AUTH_EXTRA:
       SECRET_KEY: __vault__:gen_if_missing:secret-residential/global/edxapp-lms-django-secret-key>data>value
+    EDXAPP_REGISTRATION_EXTRA_FIELDS:
+      confirm_email: "hidden"
+      level_of_education: "optional"
+      gender: "optional"
+      year_of_birth: "optional"
+      mailing_address: "hidden"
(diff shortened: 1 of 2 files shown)


pillar/edx/ansible_vars/next_residential.sls
```diff
@@ -9,3 +9,22 @@ edx:
...
     EDXAPP_CMS_AUTH_EXTRA:
       SECRET_KEY: __vault__:gen_if_missing:secret-residential/global/edxapp-lms-django-secret-key>data>value
+    EDXAPP_REGISTRATION_EXTRA_FIELDS:
+      confirm_email: "hidden"
+      level_of_education: "optional"
+      gender: "optional"
+      year_of_birth: "optional"
+      mailing_address: "hidden"
+      goals: "optional"
+      honor_code: "required"
```
shaidar: Is this required or isn't it assumed that if they're students here that they already abide by an `honor_code`?

shaidar:
Is there an issue to track the problem related to account creation prevention by means other than SAML?

pillar/edx/ansible_vars/next_residential.sls
blarghmatey: I was just copying from what we have in xpro so that I can eliminate the need to enter the country for adding users.

blarghmatey:
No, not yet. I'm going to do some digging before I merge this to see if that's even viable. If not then we'll just have to keep using CAS.

PR #843, opened 2019-05-01
Secure MITxPro /register with token

markbreedlove:

https://github.com/mitodl/mitxpro/issues/134

#### What's this PR do?

It adds an Nginx configuration for MITxPro to allow access to `/register` only if the value of the `X-Access-Token` HTTP header is correct.

`extra_locations_lms.j2` is defined in https://github.com/mitodl/configuration/blob/master/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/extra_locations_lms.j2

The Vault value for the access token is referenced in one other place: https://github.com/mitodl/salt-ops/blob/3e63b383f1ffa7b3f26e90b94c02bca807cdf353/pillar/heroku/xpro.sls#L75

### DIFF
--- a/salt/edx/templates/extra_locations_lms.j2
+++ b/salt/edx/templates/extra_locations_lms.j2
@@ -1,0 +1,24 @@
+{% raw %}
+{% if EDXAPP_SCORM_PKG_STORAGE_DIR %}
+    location ~ ^/{{ EDXAPP_MEDIA_URL }}/{{ EDXAPP_SCORM_PKG_STORAGE_DIR }}/(?P<file>.*) {
+        add_header 'Access-Control-Allow-Origin' $cors_origin;
+        add_header 'Access-Control-Allow-Credentials' 'true';
+        add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
+
+        root {{ edxapp_media_dir }}/{{ EDXAPP_SCORM_PKG_STORAGE_DIR }};
+        try_files /$file =404;
+        expires 604800s;
+    }
+{% endif %}
+{% endraw %}
+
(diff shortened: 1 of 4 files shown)


salt/edx/templates/extra_locations_lms.j2
```diff
@@ -0,0 +1,25 @@
...
+
+        root {{ edxapp_media_dir }}/{{ EDXAPP_SCORM_PKG_STORAGE_DIR }};
+        try_files /$file =404;
+        expires 604800s;
+    }
+{% endif %}
+{% endraw %}
+
+{% set environment = salt.grains.get('environment') %}
+{% set xpro_registration_access_token = __vault__:gen_if_missing:secret-{{ business_unit }}/{{ environment }}/xpro-registration-access-token>data>value %}
```
blarghmatey: This won't work. The __vault__ rendering only functions in Pillar data. You will need to set a pillar key with that setting and then pull that pillar value in the template.

Also, business_unit and environment aren't defined in this context so that would fail as well. And when templating a string inside a Jinja block you need to use Python string building or use the `~` operator to concatenate strings, rather than the `{{}}` syntax.

markbreedlove: Thanks, I'll move the vault rendering and change the string rendering.

The `environment` grain is being retrieved like `business_unit` is in `nginx_statis_assets.j2`: https://github.com/mitodl/salt-ops/blob/3e63b383f1ffa7b3f26e90b94c02bca807cdf353/salt/edx/templates/nginx_static_assets.j2#L29 ... should that be changed, too?

blarghmatey:
:+1:

PR #845, opened 2019-05-01
Updated states for libgit and added version to pillar

blarghmatey:


shaidar:
We sure this is compatible with the running version of salt-master? Just remember in my test environments, had a rough time getting `libgit` to work and it was very finicky. Other than that, it's a 👍 from me

blarghmatey:
Yup, we're sure :) That's what I just updated it to on the master.

PR #855, opened 2019-05-07
Added permissions settings for Micromasters indices on apps ES clusters

blarghmatey:
In order for Micromasters to be able to run against the ES clusters that we deployed to the RC and Prod-apps environmnets I added the configs for ReadOnlyRest so that those clusters have the permissions needed.


pillar/elasticsearch/apps.sls
```diff
@@ -33,6 +33,82 @@ elasticsearch:
...
+            type: allow
+            accept_x-forwarded-for_header: 'true'
+            methods:
+              - GET
+              - HEAD
+              - OPTIONS
+              - POST
+            indices:
+              - '_all'
+              - 'micromasters*'
```
shaidar: Should this be the `rc` index?

blarghmatey: I think that was causing issues, because this specific use case needs to be able to scan all of the indices. In any case, that's the setting that is currently used on MM and it's probably easiest to keep it as is so we don't accidentally introduce bugs.

shaidar:
👍

PR #856, opened 2019-05-08
Change redirect for xpro /register

markbreedlove:
Following up on `https://github.com/mitodl/salt-ops/pull/843`, where /register was checked for the `X-Access-Token` header before allowing access, make `/register` redirect to `/login` and apply the access token authentication to the API endpoint, `/user_api/v1/account/registration`.


shaidar:
👍

PR #858, opened 2019-05-08
Added openedx_environment

shaidar:
#### What's this PR do?
There's a mismatch between Heroku pillar and mitxpro as far as the vault registration access key. This should address that and I also moved a few mailgun settings to the env_dict


blarghmatey:
:+1:

PR #871, opened 2019-05-21
Add OCW mirror configuration

markbreedlove:
This adds some minimal configuration to make sure that various directories exist on the mirror server.

### DIFF
--- a/pillar/apps/ocw.sls
+++ b/pillar/apps/ocw.sls
@@ -12,6 +12,37 @@ ocw:
   dspace_prod_connection_password: __vault__::secret-open-courseware/{{ ENVIRONMENT }}/dspace/prod>data>password
   github_ssh_key: __vault__::secret-open-courseware/global/github/ssh-deploy-key>data>value
   cron_mailto: mitx-devops@mit.edu
+  mirror:
+    fs_owner: ocwuser
+    fs_group: ocwuser
+    # rootdirectory corresponds to `rootdirectory' in the [Mirror] section of
+    # the engine server's engines.conf. It's apparently where the final files
+    # are put for being rsynced.
+    rootdirectory: /data2/prod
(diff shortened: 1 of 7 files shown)


blarghmatey:
Looks good to me. @shaidar anything I'm missing?

pillar/apps/ocw.sls
```diff
@@ -12,6 +12,37 @@ ocw:
...
+      - /data/prod/terms
+      - /data/prod/images
+      - /data/prod/jsp
+      - /data/prod/jw-player-free
+      - /data/prod/mathjax
+      - /data/prod/OcWeb
+      - /data/prod/scripts
+      - /data/prod/styles
+      - /data/prod/webfonts
+      - /data/prod/subscribe
```
shaidar: Shouldn't all those be `/data2/`?

salt/apps/ocw/mirror.sls
```diff
@@ -0,0 +1,23 @@
+
```
shaidar: nit: Why the blank line?

markbreedlove: no good reason! :-) I'll take it out.

pillar/apps/ocw.sls
markbreedlove: The `/data` paths are hardcoded in https://github.com/mitocw/ocwcms/blob/25f31dd2a15b6b658b0fa59d0cd8ebb8ebe0f7c7/mirror/scripts/ocw6%20Mirror%20Scripts/create_new_snapshot.sh

`/data2` exists alongside `/data`. It's configured in the engines.conf file. I don't know why these are two separate directories, but they have been that way on the legacy systems and I'm hesitant to change things until I can say better what the separation is about.

Issue #865, opened 2019-05-10
Scale up graders for 6.883


blarghmatey:
The MOOC that uses the 6.883 graders is starting on June 11th. We need to update the configuration to pull from the edX Xqueue on the residential instances and scale up the quantity to handle a larger group.

PR #873, opened 2019-05-24
Added configs for mooc graders to residential pillar and increased count

blarghmatey:
In order for us to support the 6.86x MOOC we need to run an increased number of graders for it. Rather than continuing to attempt to use Lightsail, and now that we have separated the graders for different courses, we will use the residential environment for running those xqueue-watchers.

#865 

#### What's this PR do?
Adds the configuration for the MOOC xqueue to the residential watcher instances and increases the number to deploy


shaidar:
👍

PR #879, opened 2019-06-07
Add "JSON for mobile" and Zope restart cron jobs

markbreedlove:
See https://github.com/mitocw/ocwcms/pull/81

A new cron job has been added. We've decided in Slack that we will try restarting the CMS on cron before it runs to avoid issues with its known high memory usage. This will restart both CMS 1 and 2. It might be nice to have that done, anyway, cron job or not.


blarghmatey:
:+1:

PR #884, opened 2019-06-12
Remove hubble-salt as it's been depractaed

shaidar:
#### What's this PR do?
Removed hubble-salt from salt-master config as we don't use it and it's [no longer maintained](https://github.com/hubblestack/hubble-salt)


blarghmatey:
:+1:

Issue #868, opened 2019-05-15
Restart residential workers after creds are updated


shaidar:
There's a scheduled task that runs on the residential worker instances every 5 days to update the credentials. However, that tasks does not restart the workers and when the credentials expire, even though they are not on disk, the worker process would still be using the expired credentials. We need to restart the worker processes so that they use the valid credentials.

pdpinch:
This question is a digression, but what kind of monitoring do we have for the workers? Shouldn't there be an error and an alert when their credentials have expired?

shaidar:
We have the usual monitoring as far as alerts being triggered based on Datadog metrics on those instances. There are also the elastalerts that are triggered, but doesn't appear that we have one for this particular case. Will get the logs and put in a PR for an alert related to this.

shaidar:
The elastalert is more involved than I initially assumed as we are only shipping the logs under `/edx/var/log/{cms,lms}/edx.log` and the error is actually logged in one of the files under `/edx/var/log/supervisor/lms*-stderr.log`.
Will create a separate issue to track the alert and log shipping on the worker side.

shaidar:
Deployed

PR #885, opened 2019-06-12
Restart edx_worker if config changes

shaidar:
[Issue#868](https://github.com/mitodl/salt-ops/issues/868)

#### What's this PR do?
Restart the edx worker service if/when the config changes.

### DIFF
--- a/salt/edx/run_ansible.sls
+++ b/salt/edx/run_ansible.sls
@@ -101,6 +101,8 @@ restart_edx_worker_service:
     - name: all
     - restart: True
     - bin_env: '/edx/bin/supervisorctl'
-    - onchanges_in:
-      - file: {{ conf_file }}
+    - onchanges:
+      - salt: place_ansible_environment_configuration
+    - require:
+      - cmd: run_ansible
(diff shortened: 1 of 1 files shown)


salt/edx/run_ansible.sls
```diff
@@ -94,3 +94,13 @@ run_ansible:
...
     - env:
         HOME: /root
+
+{% if 'edx-worker' in salt.grains.get('roles') %}
+restart_edx_worker_service:
+  supervisord.running:
+    - name: all
+    - restart: True
+    - bin_env: '/edx/bin/supervisorctl'
+    - onchanges_in:
```
blarghmatey: This would be `onchanges` for this case. The `_in` suffix would be if you added it to the conf file state id. Also, the argument should use the state id for that step.

```
    - onchanges:
      - file: place_ansible_environment_configuration
```

blarghmatey: This should also have a `require` statement for the `run_ansible` step.

blarghmatey:
:+1:

Issue #877, opened 2019-06-06
Configure xPro alerting when OpenedX account fails provisioning


shaidar:
When a new user signs up to xPro and as part of the process, an account is created on Heroku xPro side which then triggers a task to create an account on the associated OpenedX instance for that environment. If there are any [configuration](https://github.com/mitodl/mitxpro/blob/master/docs/configure_open_edx.md) discrepancies or some issues on the OpenedX side, the account could potentially fail without a clear indication to the end user about this problem. We should be alerted when this error is raised in the Heroku logs.
We've come across different error messages but I think a good start would be to raise a P3 alert when this is raised in `heroku.xpro` fluentd_tag: `courseware.exceptions.OpenEdXOAuth2Error`

shaidar:
Deployed, but before closing this issue, I would like to test it out and verify by creating a mismatch between Heroku and openedx.

PR #886, opened 2019-06-12
Added alert when account provisioning fails

shaidar:
[Issue#877](https://github.com/mitodl/salt-ops/issues/877)

#### What's this PR do?
Sends a `P3` OpsGenie alert when something is abnormal with account provisioning on the openedx instances for MIT xPRO


blarghmatey:
:+1:

PR #891, opened 2019-06-14
Added sqs engine config to salt-master

shaidar:
#### What's this PR do?
Added `sqs_engine` to salt-master config for mitxpro-production autoscaling.

### DIFF
--- a/pillar/salt_master.sls
+++ b/pillar/salt_master.sls
@@ -1,4 +1,5 @@
 {% set ONE_WEEK = 604800 %}
+{% set sqs_mitxpro_production_queue = 'edxapp-xpro-production-mitxpro-production-autoscaling' %}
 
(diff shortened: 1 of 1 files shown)


pillar/salt_master.sls
```diff
@@ -165,6 +166,17 @@ salt_master:
             - salt://reactors/slack/post_event.sls
         - salt/engine/sqs/mitxpro-production-autoscaling:
             - salt://reactors/mitxpro/edxapp_ec2_autoscale.sls
+    engine:
```
blarghmatey: nitpick: I would pluralize this since we will probably add other engines over time

blarghmatey:
:+1:

pillar/salt_master.sls
shaidar: Funny u mention it. I had it pluralized, but was worried that it might fail for some reason due to duplicate keys

PR #892, opened 2019-06-17
Added fix for xqwatcher logging

blarghmatey:
Added fix for being able to write out xqwatcher logs to disk so we can pull them with FluentD


shaidar:
👍

PR #894, opened 2019-06-21
Fix ownerships of OCW course dirs on origin server

markbreedlove:
After the rsync of files from the ocwcms working copy, make sure that various subdirectories of `/var/www/ocw/courses` are owned by `fsuser`, so that the publishing process (which copies files as `fsuser`) can copy files into these directories.

@blarghmatey Do you think this is an OK way to do this?

### DIFF
--- a/salt/apps/ocw/sync_repo.sls
+++ b/salt/apps/ocw/sync_repo.sls
@@ -65,11 +65,24 @@ ensure_that_webroot_is_writable_by_fsuser:
     - user: fsuser
     - group: www-data
 
-ensure_that_courses_dir_is_writable_by_fsuser:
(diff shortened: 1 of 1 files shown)


blarghmatey:
Rather than having to list the different subdirectories under `/courses` you can just add the following to the `file.directory statement:

```
  file.directory:
    - name: /courses
    - user: fsuser
    - group: fsuser
    - recurse:
      - user
      - group
```

markbreedlove:
The only issue with recurse is it can take a long time if there are a lot of files in the directory. I was once using the `recurse` option and it was taking over 15 minutes to run the state on production, where `courses` has 29G of data. I forgot to mention that!

blarghmatey:
In that case :+1:

PR #896, opened 2019-06-25
Added config for blocking visibility of xpro course catalog

blarghmatey:
In order to prevent anonymous users from viewing the course catalog on the edX deployment for xpro we are setting the visibility level to staff only.


blarghmatey:
@pdpinch I'm assuming we don't need to set `COURSE_ABOUT_VISIBILITY_PERMISSION` but I can set that as well if desired.

pdpinch:
Let's set `COURSE_ABOUT_VISIBILITY_PERMISSION` to False as well.

We can test it on sandbox/qa before we go live.

blarghmatey:
@shaidar this is ready for review

shaidar:
👍

PR #898, opened 2019-06-26
Tweaks for asg to highstate new instances

shaidar:
#### What's this PR do?
Added an additional reactor to highstate mitxpro instances once they've been created by the autoscaling group. Tried replacing `cloud.create` with the `cloud.profile` directive as it had an option to highstate the instance once the minion was installed, however it was creating two instances, one through salt, and one through auto-scaling, so had to stick with `cloud.create`.

### DIFF
--- a/salt/reactors/mitxpro/edxapp_highstate.sls
+++ b/salt/reactors/mitxpro/edxapp_highstate.sls
@@ -1,0 +1,4 @@
+edxapp_highstate:
+  local.state.apply:
+    - tgt: 'edx-*mitxpro-production-xpro-production-*'
+    - queue: True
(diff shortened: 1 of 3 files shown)


salt/reactors/mitxpro/edxapp_highstate.sls
```diff
@@ -0,0 +1,4 @@
+edxapp_highstate:
+  local.state.apply:
+    - tgt: 'edx-*mitxpro-production-xpro-production-*'
+    - queue: True
```
blarghmatey: I think it's worth either modifying pillar data that applies only to the non-base instances, or appending pillar data here to specify the ansible flags to only run the configuration updates.

blarghmatey:
:+1

PR #902, opened 2019-06-28
Added COMMON_ENABLE_DATADOG to xpro

shaidar:
Related to issue described [here](https://github.com/mitodl/configuration/pull/85) and this is a temp fix.


blarghmatey:
I think we should just add this to the ansible/init.sls so that it's global across envs since this will end up being a problem everywhere.

PR #903, opened 2019-07-01
Added reactor for pruning vault cache on instance termination

blarghmatey:
When deleting an instance we should purge the associated cache in Vault to keep it clean without manual intervention.


shaidar:
👍

PR #904, opened 2019-07-01
Add Elastalert alert for OCW CMS error

markbreedlove:
This adds an Elastalert alert for when an "invalid literal for int()" error is encountered by OCW's Zope CMS. This usually means that the CMS's Zope service needs to be restarted.


blarghmatey:
:+1:

Issue #907, opened 2019-07-02
Fix noisy Elastalert alerts for OCW


markbreedlove:
I'm afraid I did not spend enough time thinking through the alert that I configured in https://github.com/mitodl/salt-ops/pull/904 .  It sends out too many OpsGenie alerts in too short a time. Additionally, I have been informed that the search query picks up too many false positives.

The configuration needs to be changed so that OpsGenie sends out just one alert when it encounters a number of matching messages, and then waits for the alert to be cleared before doing so again. The Elasticsearch query also needs to do an untokenized search (e.g. on `messages.raw` if it exists) to match the exact phrase "invalid literal for int".

markbreedlove:
OpsGenie's alert API has an `alias` property, for deduplication: https://docs.opsgenie.com/docs/alert-deduplication

[Elastalert has a property named `opsgenie_alias`](https://elastalert.readthedocs.io/en/latest/ruletypes.html?highlight=opsgenie_alias), which will set this alias.

markbreedlove:
https://github.com/mitodl/salt-ops/pull/910 doesn't use an untokenized field, but uses a querystring query with the "AND" operator, instead.

markbreedlove:
Here's the Elasticsearch query that the Elastalert should be using:
```
{
	"query": {
		"bool": {
			"must": {
				"query_string": {
					"default_field": "message",
					"query": "error AND invalid AND literal AND int"
				}
			},
			"filter": {
				"term": {"fluentd_tag": "ocwcms.zope.event"}
			}
		}
	}
}
```

markbreedlove:
I've deployed #910 and will close this after we have evidence it's working correctly.

markbreedlove:
I've tested this by manually adding numerous fake log messages. I only got one alert because OpsGenie de-duplicated them.

PR #910, opened 2019-07-03
Fix OCW CMS Elastalert

markbreedlove:

https://github.com/mitodl/salt-ops/issues/907

#### What's this PR do?

Fix OCW CMS Elastalert for OOM or cache exhaustion errors that manifest themselves as "invalid literal for int()" errors.

* Get rid of false positives by using a better query
* Get rid of redundant alerts by using the OpsGenie `alias` field.

### DIFF
--- a/pillar/elastalert.sls
+++ b/pillar/elastalert.sls
@@ -331,6 +331,7 @@ elastic_stack:
             overconsumption -- memory or disk cache.
           opsgenie_key: {{ opsgenie_key }}
           opsgenie_priority: P3
+          opsgenie_alias: ocw_invalid_literal_for_int
           type: frequency
           index: logstash-ocw-*
           num_events: 1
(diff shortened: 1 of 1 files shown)


markbreedlove:
Here's the Elasticsearch query that the Elastalert should be using:
```
{
  "query": {
    "bool": {
      "must": {
        "query_string": {
          "default_field": "message",
          "query": "error AND invalid AND literal AND int"
        }
      },
      "filter": {
        "term": {"fluentd_tag": "ocwcms.zope.event"}
      }
    }
  }
}
```

blarghmatey:
:+1:

markbreedlove:
I've just tested it with `elastalert-test-rule` over a span of the past 7 days, and it indicates that it would have caught 31 matches only on 2019-06-27, which looks right. I think that we have to deploy it to see how well OpsGenie deduplicates the alerts with the `alias` field.

blarghmatey:
Sounds good. :shipit:

PR #911, opened 2019-07-03
Update the list of nginx ciphers

shaidar:
#### What's this PR do?
Updates the Nginx ssl_ciphers based on [OWASP's broad compatibility list](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/TLS_Cipher_String_Cheat_Sheet.md).
I had a question regarding the existing list as far as when to specify the ciphers that are not accepted (ex. `!DSS`) and on further [reading](https://www.nginx.com/blog/nginx-https-101-ssl-basics-getting-started/#CipherSuites), my understanding was that `ssl_ciphers` becomes practically a whitelist when `ssl_prefer_server_ciphers` is set to `on` which is on all our Nginx configs. So, we don't have to specifically specify the ciphers that we don't accept.


blarghmatey:
👍

Issue #862, opened 2019-05-08
Assign saltenv configuration to newly deployed instances


blarghmatey:

blarghmatey:
The saltenv system proved to not be fully baked and we abandoned that approach in favor of distinct master servers.

Issue #861, opened 2019-05-08
Assign saltenv configuration values to existing infrastructure


blarghmatey:
Assign the `saltenv` configuration parameter to the minion config for existing instances so that production instances use the `production` saltenv and appropriate values for other minions.

blarghmatey:
This is obsoleted by our current approach of building separate salt masters.

Issue #860, opened 2019-05-08
Update configuration for pillar and file server backends to specify salt environments


blarghmatey:
Add configuration for defining salt environments as separate branches of the salt-ops repository. For both pillar and fileserver backends.

blarghmatey:
This is obsoleted by our move to separate salt masters.

PR #917, opened 2019-07-09
Added settings for controlling salt environments

blarghmatey:
Added configurations for specifying salt envs based on git branches in order to avoid accidentally pushing changes to production before they are ready

#862 #860 #861 

#### What's this PR do?
Specifies salt envs based on git branches and updates minion settings to specify their assigned envs

Merge, deploy configs to pre-production minions, make a minor change in master and verify that it isn't reflected

Because of the fact that we have configuration updates that run automatically we run the risk of accidentally applying changes to production infrastructure before it has been fully vetted. This is designed to avoid that situation.

### DIFF
--- a/pillar/common.sls
+++ b/pillar/common.sls
@@ -1,3 +1,5 @@
+{% set environment = salt.grains.get('environment') %}
+
 mine_functions:
   network.ip_addrs: [eth0]
   network.get_hostname: []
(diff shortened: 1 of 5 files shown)


shaidar:
As far as I can tell, it's a 👍

blarghmatey:
That gets passed in as a context variable in the `file.managed` call in the orchestrate scripts that use the file.

markbreedlove:
OK, thanks!

Issue #772, opened 2019-03-14
Finish configuring logrotate for OCW


markbreedlove:
`pillar/logrotate/ocw_cms.sls` exists already but is apparently unused, because the referenced files have not been managed on the OCW CMS and DB instances. Ensure that the logrotate configurations are managed by Salt.

This may mean creating a formula to call `logrotate.set`: http://docs.saltstack.com/en/latest/ref/states/all/salt.states.logrotate.html#module-salt.states.logrotate (Please update this discription with the correct remedy if this is not right.)

markbreedlove:
FYI, I have manually created some logrotate files for now.

markbreedlove:
Somehow since this issue was created the logrotate state started working, but PR #918 is still necessary to rotate files that have since been created or discovered.

markbreedlove:
Needs PR #926 before being resolved.

PR #918, opened 2019-07-09
Fix evaluation of role, OCW logrotate pillar

markbreedlove:

https://github.com/mitodl/salt-ops/issues/772

#### What's this PR do?

Fixes how it's determined that logrotate configuration files need to be written for logs that are only found on the "engine" CMS server for OCW.

I'm making a variable assignment for `ocw_cms_role` for readability and to be more consistent with other `.sls` files, and I'm fixing the equality evaluation because `ocw_cms_role` is a list, not a string.

### DIFF
--- a/pillar/logrotate/ocw_cms.sls
+++ b/pillar/logrotate/ocw_cms.sls
@@ -21,7 +21,7 @@ logrotate:
       - notifempty
       - compress
       - delaycompress
-  {% if if 'engine' in ocw_cms_role %}
+  {% if 'engine' in ocw_cms_role %}
   ocw_publishing_logs:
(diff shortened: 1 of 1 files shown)


pillar/logrotate/ocw_cms.sls
```diff
@@ -20,7 +21,7 @@ logrotate:
       - notifempty
       - compress
       - delaycompress
-  {% if salt.grains.get('ocw-cms-role') == 'engine' %}
+  {% if if 'engine' in ocw_cms_role %}
```
blarghmatey: double `if`

markbreedlove: Ah!  I'll fix that.

PR #928, opened 2019-07-11
Resize reddit instance to t3.large

shaidar:
#### What's this PR do?
I have already manually resized the `reddit-production` instances from `t2.medium` to `t3.large` and this is to keep things in line as far instance type.

We have been continually getting notifications in the `devops-notfications` slack channel about restarting the reddit surface due to low memory on the reddit-production instances.


blarghmatey:
:+1:

PR #929, opened 2019-07-11
Changes to deploy xpro analytics instance

shaidar:
[Issue#767](https://github.com/mitodl/mitxpro/issues/767)

#### What's this PR do?
Removed the `residential` keyword from analytics since we want to use this for xpro as well. Also made some changes to state file and yml one to be in line with rest of states and map files.


blarghmatey:
:+1:

PR #930, opened 2019-07-11
Add conditional to for fluentd gitreload

shaidar:
#### What's this PR do?
Adds an `if` statement to check whether the instance environment is `mitxpro` or not as those instances don't use gitreload.

Fluentd service was running, however it was throwing errors because the `gr` folder wasn't found. xPro instances don't use gitreload.

### DIFF
--- a/pillar/fluentd/mitx.sls
+++ b/pillar/fluentd/mitx.sls
@@ -1,5 +1,6 @@
 {% from "fluentd/record_tagging.jinja" import record_tagging with context %}
 {% from "fluentd/auth_log.jinja" import auth_log_source, auth_log_filter with context %}
+{% set ENVIRONMENT = salt.grains.get('environment', 'mitx-qa') %}
 
 fluentd:
   overrides:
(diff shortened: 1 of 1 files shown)


pillar/fluentd/mitx.sls
```diff
@@ -68,6 +69,7 @@ fluentd:
             - format1: '/^(?<time>\w{3}\s+\d{1,2}\s+\d{1,2}:\d{2}:\d{2}) (?<hostname>[^ ]+?) \[service_variant=(?<service_variant>\w+?)\]\[(?<namespace>[a-zA-Z._-]+?)\]\[env:(?<logging_env>[a-zA-Z-_.]+)\]\ (?<log_level>\w+) \[[^ ]+\s+\d+\] \[(?<filename>[a-zA-Z0-9-_.]+):(?<line_number>\d+)\] - (?<message>.*)$/'
             - time_format: '%b %d %H:%M:%S'
             - multiline_flush_interval: '5s'
+        {% if 'mitxpro' not in ENVIRONMENT %}
```
blarghmatey: Rather than being conditional on the environment this might be more robust if we make it conditional on the presence of the directory, e.g. `if salt.file.directory_exists('/edx/var/log/gr`)`

blarghmatey:
:+1:

PR #931, opened 2019-07-12
Change ocw-origin instance type in AWS cloud profile

markbreedlove:
Change its instance type to m4.large. m5.large would have been ideal, but that type is not available in us-east-e.


blarghmatey:
:+1:

Issue #932, opened 2019-07-12
Fix fluentD parsing of OCW origin Nginx access log


markbreedlove:
Our fluentD log shipper is not parsing OCW origin server Nginx access logs correctly. With the Elasticsearch index's dynamic mapping, this causes a lot of garbage fields to be added to the index.

The fluentD configuration that I added in https://github.com/mitodl/salt-ops/pull/768 is wrong for [ocworigin_nginx_access_log](https://github.com/mitodl/salt-ops/blob/c3066bb1cd949d3c4180907b848f9088c43b47bb/pillar/fluentd/ocw_origin.sls#L11-L26).

Here is an example Nginx access log record. Note the bad field names under `_source`:
```
{
  "_index": "logstash-ocw-2019.07.12",
  "_type": "fluentd",
  "_id": "5u5A52sBec1clCmN7iWZ",
  "_version": 1,
  "_score": null,
  "_source": {
    "200": null,
    "47305": null,
    "10.100.0.22": null,
    "-": null,
    "[12/Jul/2019:17:34:58": null,
    "+0000]": null,
    "\"GET /courses/packageslip.xml HTTP/1.1\"": null,
    "\"-\"": null,
    "\"ocw staging engine\"": null,
    "minion_id": "ocw-qa-ocw2",
    "environment": "ocw",
    "roles": "ocw-origin",
    "@timestamp": "2019-07-12T17:34:58.000000000+00:00",
    "fluentd_tag": "ocworigin.nginx.access"
  },
  "fields": {
    "@timestamp": [
      "2019-07-12T17:34:58.000Z"
    ]
  },
  "highlight": {
    "fluentd_tag": [
      "@kibana-highlighted-field@ocworigin.nginx.access@/kibana-highlighted-field@"
    ]
  },
  "sort": [
    1562952898000
  ]
}
```

This is probably a performance and filesystem storage issue, if Elasticsearch has to maintain indices on so many fields. Plus, the logs are unusable.

We should have this fixed before we migrate the OCW site to AWS and we unleash the Akamai CDN on the production origin server.

TODO:

- [x] merge PR
- [x] deploy and then run `nginx.ng` Salt state on the origin servers
- [x] verify new log entries in Kibana
- [x] dump old OCW logstash indices with bad fields

PR #933, opened 2019-07-12
Fix OCW origin server's fluentD access log parsing

markbreedlove:

https://github.com/mitodl/salt-ops/issues/932

#### What's this PR do?

~~Fix the parsing attributes for the fluentD configuration that parses OCW origin server Nginx access logs.~~

~~Use the fluentD bundled 'apache2' type, which is a misnomer: it's the standard Combined Log Format, which is used by our Nginx servers. It's the same log format that has been working fine with `ocs_cms.sls`.~~

Fix OCW origin server's access log format in Nginx configuration so that fluentD can parse it correctly. Include the `nginx` pillar in top.sls under 'roles:ocw-origin' so that the `nginx.ng` state knows the correct log format parameters.

See https://github.com/mitodl/salt-ops/issues/932

### DIFF
--- a/pillar/fluentd/ocw_origin.sls
+++ b/pillar/fluentd/ocw_origin.sls
@@ -17,13 +17,8 @@ fluentd:
             - nested_directives:
               - directive: parse
                 attrs:
-                  - '@type': ltsv
-                  - null_value_pattern: '-'
+                  - '@type': apache2
                   - keep_time_key: 'true'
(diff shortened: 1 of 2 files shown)


markbreedlove:
I've updated this PR's description to describe a revised strategy.

Per @blarghmatey in our recent phone conversation, this is OK to merge.

PR #934, opened 2019-07-15
Improve file_limits sls, apply it to log cluster

markbreedlove:
* Improve `file_limits.sls` by doing a `file.replace` instead of `file.append`, in case the `fs.file-max` line is already present.
* Apply the state to the Operations Elasticsearch logging cluster, where we have run into problems due to the open-files limit having been exceeded.


blarghmatey:
:+1:

Issue #935, opened 2019-07-15
fluentD not parsing some OCW Nginx log fields correctly


markbreedlove:
fluentD is not correctly parsing the `request` and `referer` fields of some Nginx log entries, when the `@type` is `ltsv`.

Given the following request:
```
time=2019-07-15T16:19:36+00:00 client=18.20.162.137 method=GET request="GET /high-school HTTP/1.1" request_length=747 status=200 bytes_sent=34867 body_bytes_sent=34626 referer=https://ocw-origin.odl.mit.edu/high-school/ user_agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.100 Safari/537.36" upstream_addr=- upstream_status=- request_time=0.000 request_id=13f04bf99e14a0054a50ac8d45abe9cb upstream_response_time=- upstream_connect_time=- upstream_header_time=-
```
... the `request`, `referer`, and `time` properties are `null`, though they are defined in the log line:
```
[ ... ]
{
  "_source": {
    "time": null,
    "client": "18.20.162.137",
    "method": "GET",
    "request": null,
    "request_length": "747",
    "status": "200",
    "bytes_sent": "34867",
    "body_bytes_sent": "34626",
    "referer": null,
    "user_agent": "\"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.100 Safari/537.36\"",
    "upstream_addr": null,
    "upstream_status": null,
    "request_time": "0.000",
    "request_id": "13f04bf99e14a0054a50ac8d45abe9cb",
    "upstream_response_time": null,
    "upstream_connect_time": null,
    "upstream_header_time": null,
    "minion_id": "ocw-origin-ocw-0",
    "environment": "ocw",
    "roles": "ocw-origin",
    "@timestamp": "2019-07-15T16:19:36.000000000+00:00",
    "fluentd_tag": "ocworigin.nginx.access"
  },
[ ... ]
}
```

PR #936, opened 2019-07-15
Fix fluentD parsing of Nginx access log null value

markbreedlove:

Fixes https://github.com/mitodl/salt-ops/issues/935

#### What's this PR do?

Fix the regex that fluentD uses to determine a null value in a field in an Nginx access log.

With the `null_value_pattern` option, fluentD takes a field with a whole value of "-" and stores it as null. The problem was that the `null_value_pattern` parameter is a regex, so it was registering anything _containing_ a hyphen as null.

I have taken the liberty of modifying some other fluentD configurations for other products that had the same `null_value_pattern`.

I have made the change manually to `/etc/fluent/fluent.d/ocworigin.conf` on our origin webserver and have observed that it fixes the problem.

### DIFF
--- a/pillar/fluentd/cas.sls
+++ b/pillar/fluentd/cas.sls
@@ -19,7 +19,7 @@ fluentd:
                 - directive: parse
                   attrs:
                     - '@type': ltsv
-                    - null_value_pattern: '-'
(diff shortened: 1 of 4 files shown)


shaidar:
👍

PR #939, opened 2019-07-16
Enable Datadog on OCW CMS & DB, production

markbreedlove:

Recent Slack conversations about monitoring OCW servers for OOM problems.

#### What's this PR do?

Adds Datadog agent to OCW CMS and DB servers, in production.
Only adds the base system monitoring, not any integrations / plugins.

### DIFF
--- a/pillar/top.sls
+++ b/pillar/top.sls
@@ -270,3 +270,5 @@ base:
     - match: grain
     - logrotate.ocw_cms
     - fluentd.ocw_db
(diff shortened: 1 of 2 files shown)


shaidar:
The one thing you might want to do before running the `datadog` state on those minions is possibly rename the `environment` grain on the minions as that's what the [datadog pillar data uses to tag the instance](https://github.com/mitodl/salt-ops/blob/master/pillar/datadog/init.sls#L7). Other than looks good to me.

PR #940, opened 2019-07-18
Fix problem with OCW environment names

markbreedlove:
Add include syntax to `environment_settings.yml` to work around a
subtlety with environment name grains.

Commit 7c834d5badd51b4ac18e6aaa82401bba597c1d6a introduced a problem
with OCW environment names. Per https://github.com/mitodl/salt-ops/pull/939#issuecomment-511955312 I modified the Salt grains on our OCW instances from "ocw" to "ocw-production" or "ocw-qa". As a result, Nginx templates now fail to render in the nginx.ocw_origin pillar, where it tries to look up data in `environment_settings.yml`.

### DIFF
--- a/salt/environment_settings.yml
+++ b/salt/environment_settings.yml
@@ -898,7 +898,7 @@ environments:
           parameter_group_name: default.redis3.2
           failover_enabled: False
           business_unit: odl-video
-  ocw:
+  ocw: &ocw
     business_unit: open-courseware
     network_prefix: '10.100'
     vpc_name: OCW
@@ -921,6 +921,10 @@ environments:
         security_groups:
           - webapp
           - ssh
+  ocw-production:
+    <<: *ocw
+  ocw-qa:
+    <<: *ocw
 
 business_units:
   - bootcamps


shaidar:
I think doing it this way complicates things a little as u'd have production domain under `ocw-qa` as well as `ocw-production`.

markbreedlove:
Yeah, it's a little complicated. If I abandon this, do you think I can run the Datadog state again with the environment being `ocw` instead of `ocw-production`? I just need to accomplish two things as soon as possible: 1) have Datadog alerts set up, and 2) deploy `ocwcms` to OCW again to fix some images before we go live.

shaidar:
Yeah guess we can do that and then fix datadog manually later (if possible) or when u're not rushed for time split the environments like we have with the rest of the apps. Can you just open an issue for that as `tech debt` so that we don't lose track of it.

markbreedlove:
Yes, thank you! Sounds good.
I'll leave this branch so that I can follow up later by reverting my commit and fixing it the right way.

PR #942, opened 2019-07-22
Move xpro settings changes to run_ansible state

shaidar:
#### What's this PR do?
When a scheduled configuration task runs, it overwrites changes made to `production.py` and thus the supervisord services fail to start. Changes to that file were made through the `hacks` state however that does not runs as part of the scheduled configuration task. So, I moved those directives to the `edx.run_ansible` state to resolve the problem.


shaidar:
I was mistaken. Closing this

PR #943, opened 2019-07-22
Restart edx-workers to use new creds

shaidar:
#### What's this PR do?
Restart instances with the `edx-worker` role on a schedule to pick up new/updated credentials.


blarghmatey:
:+1:

PR #944, opened 2019-07-23
Fix fluentD parsing of nginx error logs

markbreedlove:

None. I noticed the problem while looking at our logs in Kibana.

#### What's this PR do?

Fix the fluentD parse directives for various products' nginx error logs.

The one that has been working correctly is in `pillar/fluentd/cas.sls`. Correct the regular expressions in the other ones to match the one in `cas.sls`.

Filter on the following `fluentd_tag` values in Kibana: `cas.nginx.error`, `ocworigin.nginx.error`, `odlvideo.nginx.error`, and `reddit.nginx.error`. You'll see that `cas.nginx.error` is the only one that shows up, and the `cas.sls` configuration is the one that is different than the others.

### DIFF
--- a/pillar/fluentd/ocw_origin.sls
+++ b/pillar/fluentd/ocw_origin.sls
@@ -36,7 +36,7 @@ fluentd:
               - directive: parse
                 attrs:
                   - '@type': regexp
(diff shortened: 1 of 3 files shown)


blarghmatey:
:+1:

PR #945, opened 2019-07-25
Set SESSION_COOKIE_SECURE to true

shaidar:
#### What's this PR do?
Sets session cookie to only be sent over `https` connections only.


blarghmatey:
:+1:

PR #956, opened 2019-07-30
Moved JWT_AUTH to hacks.sls

shaidar:
#### What's this PR do?
Came across an issue where by the time the vault values were written out to `lms.env.json` lms was throwing errors for an invalid `JWT_PRIVATE_SIGNING_JWK`. It appeared that some `'` and `\n` were being added along the line (probably ansible rendering). The suggestion was to use the values as specified [here](https://github.com/edx/configuration/blob/open-release%2Fironwood.master/playbooks/roles/edxapp/defaults/main.yml#L1213-L1223)


blarghmatey:
:+1:

PR #957, opened 2019-07-30
Move JWT config to production.py

shaidar:
Even after using the JWT settings based on [this](https://github.com/edx/configuration/blob/open-release%2Fironwood.master/playbooks/roles/edxapp/defaults/main.yml#L1213-L1223), when the values were written to disk, it was still doing something weird with the rendering by adding quotes and a `\n` to the private and public key sets which was throwing a `ValueError: No JSON object could be decoded` in the lms error log.
This just moves the values to the `production.py` file


blarghmatey:
:+1:

PR #959, opened 2019-07-31
Tweaks for mitxpro autoscaling

shaidar:
#### What's this PR do?
Those are two fixes that I had already made manually in the AWS console and just now committing them to the repo.


blarghmatey:
:+1:

PR #960, opened 2019-07-31
Added JWT alert to xpro

shaidar:
#### What's this PR do?
When there's a misconfiguration with the JWT JWK set, an error is logged in lms about it. I added an alert as that's a major problem and can easily go unnoticed for a bit and can create xpro account issues down the line.
Additionally, I refactored a few residential alerts to always trigger on xpro.

### DIFF
--- a/pillar/elastalert.sls
+++ b/pillar/elastalert.sls
@@ -67,12 +67,12 @@ elastic_stack:
                 must_not:
                   - match:
                       message: ichuang
-      - name: mitx_operational_failure
+      - name: edx_operational_failure
         settings:
-          name: Operational Failure on mitx instances
+          name: Operational Failure on MITx or xPRO instances
           description: >-
             Send a message anytime an 'Operational Failure' message is
-            encountered in the mitx production logs.
+            encountered in the MITx or xPRO production logs.
           opsgenie_key: {{ opsgenie_key }}
           opsgenie_priority: P1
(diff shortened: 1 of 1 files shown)


pillar/elastalert.sls
```diff
@@ -82,14 +82,14 @@ elastic_stack:
...
-          alert_text: "Operational Failure on mitx-production detected"
+          alert_text: "Operational Failure on mitx or xpro production detected"
           filter:
             - bool:
                 must:
                   - match:
                       message: OperationFailure
                   - term:
-                      environment.raw: mitx-production
+                      environment.raw: mitx(pro)?-production
```
blarghmatey: Does this support regex matching?

pillar/elastalert.sls
```diff
@@ -314,14 +314,39 @@ elastic_stack:
...
+      - name: xpro_jwt_error
+        settings:
+          name: xPRO potential misconfigured JWT
+          description: >-
+            xPRO openedx instances appear to have
+            misconfigured JWT keys under lms.env.json
+          opsgenie_key: {{ opsgenie_key }}
+          opsgenie_priority: P2
+          type: frequency
+          index: logstash-*
```
blarghmatey: Should we change this to only alert on the mitxpro logstash indices?

blarghmatey:
:+1:

PR #963, opened 2019-08-01
Refactored to add edx-worker instances

shaidar:
#### What's this PR do?
Refactored autoscaling group code to also create a group for the worker instances. One caveat to it is that the alarm based on CPU load isn't really the ideal way to determine load on those instance and it would be better to base it on queue length, but that would require some integration between CloudWatch and rabbit and we can revisit at a later date.


blarghmatey:
:+1:

Issue #955, opened 2019-07-30
Change OCW production's Akamai netstorage hostname


markbreedlove:
Now that we've migrated OCW over to AWS, the production publishing engine should not be uploading media to the dummy netstorage server. The engines configuration needs to be changed to use the real hostname.

Change https://github.com/mitodl/salt-ops/blob/8e5c33b7fd4971b2cc579f8eade3ac68c44159af/pillar/apps/ocw-production.sls#L37 and the perform an rsync to upload anything that's been published in the time since the CMS freeze was lifted.

markbreedlove:
Rsync done yesterday, 2019-08-05

PR #964, opened 2019-08-05
Change OCW Akamai netstorage host, production

markbreedlove:
Change the hostname of the Akamak netstorage host, as configured in the OCW engine configuration file.

See https://github.com/mitodl/salt-ops/issues/955

@JoeMartis Is there any reason I should not merge this and update the configuration right now so that any production publishes run from now on are pushing files to the real netstorage?


JoeMartis:
Yes, we should update this asap. Also should copy whatever has been published to the faux netstorage since launch. There have been a few courses published and their zips will need to be copied over to netstorage, or the courses will need to be republished.

PR #965, opened 2019-08-06
Tune zope mem and cache

markbreedlove:

https://github.com/mitocw/ocwcms/issues/115

#### What's this PR do?

It tidies up the `zope.conf` file a little bit, and then parameterizes some memory settings.
The settings are defined by Salt grains, so that they can be different on individual CMS servers within the same environment.

The changes have already been made manually, and this adds them to `salt-ops`. I'll need to add grains for the relevant servers.


blarghmatey:
:+1:

PR #966, opened 2019-08-06
Update OCW Zope base site URLs

markbreedlove:
Update the environment variable definitions in the Zope configuration that govern the base site URLs that will be written to HTML meta tags.

The canonical URL meta tag values are wrong on the production site, and this should fix them.


blarghmatey:
:+1:

PR #967, opened 2019-08-07
Remove unnecessary OCW cron jobs from engines.sls

markbreedlove:
Remove unnecessary cron.present states.

* Remove cron job that runs `check_cache_size.sh`, which is unnecessary because we have a dedicated volume for the blobstorage directory (the cache) and we monitor it with Datadog.

* Remove commented-out cron job that used to run `copy_mitx_archived_courses_xml_from_CMS.sh`, which is unnecessary because the relevant files are on a shared filesystem.

See also https://github.com/mitocw/ocwcms/pull/121

### DIFF
--- a/salt/apps/ocw/engines.sls
+++ b/salt/apps/ocw/engines.sls
@@ -89,17 +89,6 @@ daily_broken_links_update_cronjob:
     - minute: 7
     - hour: 4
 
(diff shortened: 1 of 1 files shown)


blarghmatey:
:+1:

Issue #946, opened 2019-07-25
OCW CMS access logs not being shipped or received


markbreedlove:
OCW CMS access logs for HTTP (written by Apache and Zope) are not showing up in KIbana after July 16th.

The relevant `fluentd_tag` values are `ocwcms.zope.access` and `ocwcms.apache.access`. They were being saved correctly from July 14th through July 16h, but then they just drop off.

Looking at `ocw-production-cms-1`, it would appear that FluentD is following the  access logs, because I see `.pos` files with up-to-date timestamps -- `/var/log/apache2/access.log.pos` and `/usr/local/Plone/zeocluster/var/client_Z2.log.pos`. FluentD is configured to use its standard `apache2` parsing type, which seems to be correct. Other logs on `ocw-production-cms-1` are being sent without issue, proving that FluentD has no issue shipping to its aggregator over the network.

This issue impacts our ability to conveniently analyze errors that may be reported by users of the CMS.

PR #970, opened 2019-08-08
Fix FluentD shipping, OCW Zope and Apache access logs

markbreedlove:

Fixes https://github.com/mitodl/salt-ops/issues/946

#### What's this PR do?

Fix how FluentD ships Apache2-format access logs from OCW's Zope and Apache acccess logs.

Stop it from sending the `time` field, which has special significance for Elasticsearch, which tries to parse it as a long int and fails to save the log message with a parsing error.

FluentD's default for the `keep_time_key` parameter is `false`, but we had it set to `true`.

I have removed the `keep_time_key` parameter from a FluentD config file in QA and have observed log messages being shipped as a result.

### DIFF
--- a/pillar/fluentd/ocw_cms.sls
+++ b/pillar/fluentd/ocw_cms.sls
@@ -18,7 +18,6 @@ fluentd:
               - directive: parse
                 attrs:
                   - '@type': apache2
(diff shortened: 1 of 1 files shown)


blarghmatey:
:+1:

PR #972, opened 2019-08-09
Fix Elasticsearch log timestamp in FluentD config

markbreedlove:

I noticed this issue after committing https://github.com/mitodl/salt-ops/commit/ef621cdb505fafc846f1c0e8ff3bc2cf01f8e2f3 to get FluentD to follow the logging cluster's logs.

The Operations logging cluster's logs were not being shipped, though the correct logs were now being tailed.

#### What's this PR do?

Fix FluentD's regular expression for Elasticsearch server messages to allow timestamps with a "T", such as `2019-08-09T15:06:11,483`. Also allow a space in that position. It appears that our operations Elasticsearch cluster (version 6.x) has timestamps with "T" but other instances like those in MITx have timestamps with " ".

I've tested the revised regex in Rubular, using a sample line from the Operations Elasticsearch logging cluster.

I assume that the regex with the space character in the timestamp has been working fine for other Elasticsearch clusters, so I didn't want to break that. I'm allowing either a space or a "T" for this reason.

### DIFF
--- a/pillar/fluentd/elasticsearch.sls
+++ b/pillar/fluentd/elasticsearch.sls
@@ -15,8 +15,8 @@ fluentd:
             - path: /usr/share/elasticsearch/logs/*.log
             - pos_file: /usr/share/elasticsearch/logs/elasticsearch_fluentd.log.pos
             - format: multiline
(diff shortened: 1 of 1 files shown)


blarghmatey:
:+1:

PR #974, opened 2019-08-12
Set from bulk email address in lms production

shaidar:
[Issue#986](https://github.com/mitodl/mitxpro/issues/986)

#### What's this PR do?
This sets the `from` email address to a fixed value as opposed to adding the class prefix to the sent from address which can cause emails to mislabeled as spam.


blarghmatey:
:+1:

PR #975, opened 2019-08-12
Updated app deployments to use release ids

blarghmatey:
Added a release-id specifier in instance/minion names in order to simplify build/deploy of instances and applications.

As things stand currently, if we wish to deploy a new set of instances it requires either increasing the instance count temporarily, or performing some manual intervention. By adding an updateable and arbitrary release-id it makes it easier to deploy multiple releases in parallel.

N/A

#### What's this PR do?
Simplifies deploying new instances of a service alongside existing ones

Redeploy an existing service


blarghmatey:
One minor concern here would be that this will result in forcing a new deployment of any service that we run the corresponding orchestrate for since none of the instances have a matching version specifier in their instance names. That shouldn't be a problem however, because if we are running the orchestrate then it is likely that the intent is to actually roll out new instances anyway.

salt/orchestrate/apps/deploy.sls
```diff
@@ -11,6 +11,8 @@
     ).subnets|map(attribute='id')|list %}
 {% set security_groups = env_data.purposes[app_name].get('security_groups', []) %}
 {% do security_groups.extend(['salt_master', 'consul-agent']) %}
+{{ set release_id = salt.sdb.get('sdb://consul/' ~ app_name ~ '/' ~ ENVIRONMENT ~ '/release-id')|default('v1') }}
```
shaidar: Shouldn't this be a `%` as opposed to an extra `{`?

salt/orchestrate/services/cassandra.sls
```diff
@@ -9,6 +9,8 @@
     vpc_id=salt.boto_vpc.describe_vpcs(
         name=env_data.vpc_name).vpcs[0].id
     ).subnets|map(attribute='id')|list %}
+{{ set release_id = salt.sdb.get('sdb://consul/' ~ app_name ~ '/' ~ ENVIRONMENT ~ '/release-id')|default('v1') }}
```
shaidar: Another one

salt/orchestrate/services/consul.sls
```diff
@@ -9,6 +9,8 @@
     vpc_id=salt.boto_vpc.describe_vpcs(
         name=env_data.vpc_name).vpcs[0].id
     ).subnets|map(attribute='id')|list %}
+{{ set release_id = salt.sdb.get('sdb://consul/' ~ app_name ~ '/' ~ ENVIRONMENT ~ '/release-id')|default('v1') }}
```
shaidar: Another one

salt/orchestrate/services/elasticsearch.sls
```diff
@@ -10,6 +10,8 @@
     vpc_id=salt.boto_vpc.describe_vpcs(
         name=env_data.vpc_name).vpcs[0].id
     ).subnets|map(attribute='id')|list %}
+{{ set release_id = salt.sdb.get('sdb://consul/' ~ app_name ~ '/' ~ ENVIRONMENT ~ '/release-id')|default('v1') }}
```
shaidar: Another one

salt/orchestrate/services/elasticsearch.sls
```diff
@@ -69,12 +72,12 @@ deploy_elasticsearch_nodes:
 sync_external_modules_for_elasticsearch_nodes:
   salt.function:
     - name: saltutil.sync_all
-    - tgt: 'G@roles:elasticsearch and G@environment:{{ ENVIRONMENT }} and G@launch-date:{{ launch_date }}'
+    - tgt: '{{ target_string }} and G@launch-date:{{ launch_date }}'
```
shaidar: Why are we keeping the `launch-date` here unlike the rest of them?

blarghmatey: That's because with elasticsearch and Rabbit we might be running the state to expand an existing cluster, in which case we only want to highstate the newly launched instances.

shaidar:
👍

blarghmatey:
@markbreedlove to your comments, the map file is generated for each orchestrate run because it renders a different set of values for each app and environment so it is necessary to be updated each time.

The release id is something that we will update via Salt using the `sdb.set` function, but it is not required to be present which is where the `default('v1')` filter comes into play. The release IDs are arbitrary and not necessarily tied to any actual release version, they just serve as a way to differentiate between running and new instances.

PR #978, opened 2019-08-14
Added filter to avoid launching in us-east-1e

blarghmatey:
Our mapping of the 1e AZ is for a datacenter which doesn't support newer instance types. Avoiding launching in that to prevent deployment failures when using newer instance types.


markbreedlove:
I have yet to experiment with other AZs, so I'm not sure if any of the other Northern Virginia ones have the same issue.

blarghmatey:
@markbreedlove I don't believe we've had issues in any of the other AZs and we've definitely launched a number of the new instance types in them.

shaidar:
👍

PR #979, opened 2019-08-14
Upgraded testinfra version after fixing extension

shaidar:
#### What's this PR do?
Updates the testinfra version after having merged the fix on the extension repo. I have already updated and tested the new version manually on the salt-master and all appears to be working.


blarghmatey:
There might be some spots in other formulas that might need to be updated, but :+1:

Issue #864, opened 2019-05-08
Adding guard rails for Salt environments


blarghmatey:
In order to reduce the occurrence of accidental configuration mismatches we need to implement strict separations between different environments and applications. To make this work effectively without introducing a significant cognitive burden we are going to use a set of approaches.

- [ ] Create an additional salt master that is exclusively for communicating with production infrastructure
- [ ] Implement Salt environments to separate code to separate branches for pre-prod and prod deployments
- [ ] Add node group specifications for more easily defining a set of targets

shaidar:
While working on salt auto-scaling, I came across salt's [external authentication system](https://docs.saltstack.com/en/latest/topics/eauth/index.html#acl-eauth) and I was wondering whether we can leverage that (either to replace having multiple environments or as additional guard rails). The thinking would be that a separate account with privileges on production instances would have the rights to run specific salt commands.

PR #983, opened 2019-08-16
Salt envs

blarghmatey:
#864 

#### What's this PR do?
Retries adding separate env configurations for pillar and state files

Apply an environment config to a minion and verify proper pillar and state rendering


pillar/common.sls
```diff
@@ -16,3 +18,13 @@ salt_minion:
...
+    saltenvs:
+      {% if 'production' in environment %}
+      saltenv: production
+      lock_saltenv: True
+      pillarenv: production
+      {% else %}
+      saltenv: base
+      pillarenv_from_saltenv: True
+      {% endif %}
+      state_top_saltenv: base
```
shaidar: Not very clear on this, but since it's outside the conditional, would this not cause the base saltenv thus the master branch to apply and overwrite the one set for the production environment?

blarghmatey: So, that's actually for handling which top file gets used for targeting, not which state files/saltenv to use for rendering.

pillar/top.sls
```diff
@@ -1,4 +1,4 @@
-base:
+{{ saltenv }}:
```
shaidar: Based on pillar/common, this will always be `base` correct?

blarghmatey: No, for pillar it will be whatever the configured saltenv is

shaidar:
👍

Issue #756, opened 2019-02-24
MITxPro: enable automated export to git


pdpinch:
I think it would behoove us to enable automatic export to git and set up courses the use it. I think having versioning for courseware would benefit the xPro team.

pdpinch:
@mitodl/devops did this code get deployed?

shaidar:
Deployed to RC. Needs testing

shaidar:
@pdpinch This is deployed on RC, but need some info as far as repo/user that will be used for xPro courses.

shaidar:
Deployed to production as part of a new build. Once functionality is confirmed (when repo is created) this can be closed

pdpinch:
Thanks Sar. If I recall correctly, the repo is configured per course, using an advanced setting in the course. 

I think the user needs to be specially configured. Do you think we could reuse the user we have for residential MITx?

shaidar:
I'm closing this now as I believe it's done. If not, please fell free to re-open

PR #999, opened 2019-08-29
Added git export functionality to xpro

shaidar:
[Issue#756](https://github.com/mitodl/salt-ops/issues/756)

#### What's this PR do?
Adds git export functionality similar to how it is configured on residential.

* Note: ssh-key has not been generated and still needs to be added to vault


blarghmatey:
Looks good. There's a decent chance that we can use the same SSH key for both envs, so maybe just populate that Vault path with the same data.

PR #1003, opened 2019-09-04
Applied bulk email hack to all envs

shaidar:
#### What's this PR do?
This is required so that the setting in `lms.env.json` gets read in otherwise when the bulk email task runs it'll throw an error about the value not being set.


blarghmatey:
:+1:

PR #1004, opened 2019-09-05
Manage edx buckets in separate state file

shaidar:
#### What's this PR do?
Moved edx (residential and xpro) bucket orchestration to its own state file while removing that part from the `build_ami` is it was overwriting bucket permissions and causing read permission error on video thumbnails.

### DIFF
--- a/salt/orchestrate/aws/s3_buckets/edx.sls
+++ b/salt/orchestrate/aws/s3_buckets/edx.sls
@@ -4,6 +4,27 @@
 {% for purpose in env_data.purposes %}
 {% set bucket_prefix = env_data.secret_backends.aws.bucket_prefix %}
 
+{% for use in bucket_uses %}
+create_edx_s3_bucket_{{ use }}_{{ purpose }}_{{ ENVIRONMENT }}:
+  boto_s3_bucket.present:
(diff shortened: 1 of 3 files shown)


salt/orchestrate/aws/s3_buckets/edx.sls
```diff
@@ -0,0 +1,50 @@
...
+    - Policy:
+        Version: "2012-10-17"
+        Statement:
+          - Sid: "PublicRead"
+            Effect: "Allow"
+            Principal: "*"
+            Action: "s3:GetObject"
+            Resource: "arn:aws:s3:::{{ bucket_prefix }}-edx-video-upload-{{ purpose }}-{{ env }}/*"
+
+{% for use in bucket_uses %}
```
blarghmatey: This loop is going to override the values set in the prior step for the video uploads bucket, the same as what we ran into previously.

We can either flip the ordering, add a conditional to exclude uploads from this loop, or add a conditional to specify the policy within this loop rather than having it as a separate step.

shaidar: That loop specifies a different bucket name though - `xpro-app-{{ env }}`

blarghmatey:
Good enough  :smile: :+1:

PR #1005, opened 2019-09-05
Fixed issues with xpro worker auto-scaling

shaidar:
#### What's this PR do?
Came across some issues when trying to create the xpro worker autoscaling group and its related launch config. Not a fan of the way I fixed the issue with the security group, but think fixing it properly might require major refactoring of the envs file unless I'm missing something.

### DIFF
--- a/pillar/salt_master.sls
+++ b/pillar/salt_master.sls
@@ -1,5 +1,6 @@
 {% set ONE_WEEK = 604800 %}
-{% set sqs_mitxpro_production_queue = 'edxapp-xpro-production-mitxpro-production-autoscaling' %}
+{% set sqs_edxapp_mitxpro_production_queue = 'edxapp-xpro-production-mitxpro-production-autoscaling' %}
+{% set sqs_edx_worker_mitxpro_production_queue = 'edx-worker-xpro-production-mitxpro-production-autoscaling' %}
 
 slack_api_token: __vault__::secret-operations/global/slack/slack_api_token>data>value
 slack:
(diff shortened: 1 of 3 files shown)


pillar/salt_master.sls
```diff
@@ -179,7 +180,11 @@ salt_master:
         key: use-instance-role-credentials
       engines:
         - sqs_events:
-            queue: {{ sqs_mitxpro_production_queue }}
+            queue: {{ sqs_edxapp_mitxpro_production_queue }}
+            profile: sqs
+            tag: salt/engine/sqs/mitxpro-production-autoscaling
```
blarghmatey: Should there be different tags for each queue to avoid crossing the threads on which instance type is being managed?

salt/orchestrate/aws/mitxpro_autoscaling.sls
```diff
@@ -9,7 +9,11 @@
 {% set sqs_queue = env_data.provider_services[app_name].sqs.queue ~ '-' ~ ENVIRONMENT ~ '-autoscaling' %}
 {% set sns_topic = env_data.provider_services[app_name].sns.topic ~ '-' ~ ENVIRONMENT ~ '-autoscaling' %}
 {% set edx_codename = purpose_data.versions.codename %}
-{% set security_groups = purpose_data.get('security_groups', []) %}
+{% if app_name == 'edxapp' %}
+{% set security_groups = 'edx' %}
+{% else %}
+{% set security_groups = 'edx-worker' %}
```
blarghmatey: Is there a material difference between the settings and uses of these security groups?

pillar/salt_master.sls
shaidar: Not really sure what we'd gain by it, so left it as is.

salt/orchestrate/aws/mitxpro_autoscaling.sls
shaidar: Yeah there are quite some differences between the two that it's probably best to keep them as is

blarghmatey:
:+1:

PR #1012, opened 2019-09-17
Upgrade ubuntu and python

shaidar:
#### What's this PR do?
Point new instances to use Ubuntu Bionic (18.04) and install and use python 3.7 for built virtual envs.

### DIFF
--- a/pillar/edx/xqwatcher_600.sls
+++ b/pillar/edx/xqwatcher_600.sls
@@ -2,7 +2,7 @@
 {% set environment = salt.grains.get('environment', 'mitx-qa') %}
 {% set env_data = env_settings.environments[environment] %}
 {% set xqwatcher_venv_base = '/edx/app/xqwatcher/venvs' %}
-{% set python3_version = '3.7' %}
+{% set python3_version = 'python3.7' %}
 
 edx:
   ansible_vars:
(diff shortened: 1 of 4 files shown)


pillar/edx/xqwatcher_600.sls
```diff
@@ -2,6 +2,7 @@
 {% set environment = salt.grains.get('environment', 'mitx-qa') %}
 {% set env_data = env_settings.environments[environment] %}
 {% set xqwatcher_venv_base = '/edx/app/xqwatcher/venvs' %}
+{% set python3_version = '3.7' %}
```
blarghmatey: This will result in the Python path being `/usr/bin/3.7` rather than `/usr/bin/python3.7`

pillar/edx/xqwatcher_600.sls
```diff
@@ -32,7 +33,7 @@ edx:
             CODEJAIL:
               name: mit-600x
               user: mit-600x
-              lang: python3
+              lang: {{ python3_version }}
```
blarghmatey: This needs to stay as `python3`. It's not a path attribute, it's an input to a selection list which only supports `python2` and `python3`

blarghmatey:
:+1:

Issue #984, opened 2019-08-19
Configure Sentry monitoring for edX deployments


blarghmatey:
In order to bet better visibility into the errors of our edX installations we should add the configuration necessary to enable the Sentry integration. This is for both residential and xPro deployments.

shaidar:
That should be deployed now.

PR #1013, opened 2019-09-18
Add sentry to openedx instances

shaidar:
[Issue#984](https://github.com/mitodl/salt-ops/issues/984)

#### What's this PR do?
Using the raven client, it integrates sentry with our openedx deployment for better visibility. I tested this by manually configuring the xpro sandbox instance and was able to trigger and observe events in sentry.
I was tempted to add the `raven` install to base requirements, but didn't want to add more changes to our own fork that might differ from upstream.

* Note: When this PR is merged, a new project for residential (openedx-xpro has already been created for testing sandbox) needs to be created in sentry. Additionally, the sentry DSN's for both projects need to be added to vault.

### DIFF
--- a/pillar/edx/ansible_vars/next_residential.sls
+++ b/pillar/edx/ansible_vars/next_residential.sls
@@ -39,3 +39,5 @@ edx:
         - name: git+https://github.com/Stanford-Online/xblock-in-video-quiz@release/v0.1.7#egg=xblock-in-video-quiz
           extra_args: -e
         - name: xblock-image-modal==0.4.2
+        # Python client for Sentry
+        - name: raven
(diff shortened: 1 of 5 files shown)


pillar/edx/ansible_vars/next_residential.sls
```diff
@@ -39,3 +39,5 @@ edx:
         - name: git+https://github.com/Stanford-Online/xblock-in-video-quiz@release/v0.1.7#egg=xblock-in-video-quiz
           extra_args: -e
         - name: xblock-image-modal==0.4.2
+        # Python client for Sentry
+        - name: raven
```
blarghmatey: Should we add this to the cloud_deployments pillar?

blarghmatey:
Aside from the one comment :+1:

pillar/edx/ansible_vars/next_residential.sls
shaidar: I didn't add it there as I didn't want to change how we're deploying the additional apps and was thinking of sticking to the existing pattern

blarghmatey: Works for me.

PR #1017, opened 2019-09-25
Updated to allow for mapping anonymized emails in xpro

blarghmatey:
N/A

#### What's this PR do?
Adds configuration for allowing email mapping script to run on more applications


shaidar:
Looks good as far as I can tell 👍

Issue #1007, opened 2019-09-11
Start shipping stdout and stderr logs for edX app and worker nodes


blarghmatey:
For the edX services the application logs typically elide the tracebacks which are most useful for our debugging purposes. We would like to start shipping the stdout/stderr logs that are stored under the supervisor directory to the Elasticsearch cluster.

markbreedlove:
There are stderr logs for `cms` and `lms`, which seem useful and easy to parse.
There are also stderr logs for `forum`, but these have mixed line formatting, including lines with differently-styled timestamps. Are the forum stderr logs worth trying to ingest, too?

blarghmatey:
We can ignore the forum logs this time around.

PR #1018, opened 2019-09-26
Watch edX stderr log files with Fluentd

markbreedlove:

https://github.com/mitodl/salt-ops/issues/1007

#### What's this PR do?

It makes FluentD tail various stderr logfiles that are generated by the edX app.

It does not look like the stdout logs are useful. They are too free-form and tend to lack timestamps.

N/A, but I have tested the regular expressions with https://rubular.com


blarghmatey:
:+1:

PR #1019, opened 2019-09-27
Backslash-escape hyphens in MITx FluentD regexes

markbreedlove:
Backslash-escape some hyphen characters in regex character classes in the FluentD pillar file for MITx.

This gets rid of warnings that FluentD issues about the hyphens when it starts up, which should make it quicker for someone to focus on more serious errors, should they arise.


blarghmatey:
:+1:

PR #1020, opened 2019-09-30
Changes necessary to use mailgun for email in xpro

shaidar:
[Issue#101](https://github.mit.edu/xpro/xpro-issues/issues/101)

#### What's this PR do?
Running into problems on the xpro end sending bulk emails to non-institution addresses. The idea is to route email through Mailgun for xpro and keep residential as is. Decided to leverage vault paths for that given that some values are secrets and think it's better than adding `if` statements to check for the different environments especially when Mailgun has different sets of SMTP creds for different domains.

### DIFF
--- a/pillar/edx/ansible_vars/cloud_deployment.sls
+++ b/pillar/edx/ansible_vars/cloud_deployment.sls
@@ -10,9 +10,8 @@
 {% set bucket_prefix = env_data.secret_backends.aws.bucket_prefix %}
 {% set bucket_uses = env_data.secret_backends.aws.bucket_uses %}
 {% set sentry_dsn = salt.vault.read('secret-operations/global/' ~ business_unit ~ '/sentry-dsn').data.value %}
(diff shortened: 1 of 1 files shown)


pillar/edx/ansible_vars/cloud_deployment.sls
```diff
@@ -10,9 +10,8 @@
 {% set bucket_prefix = env_data.secret_backends.aws.bucket_prefix %}
 {% set bucket_uses = env_data.secret_backends.aws.bucket_uses %}
 {% set sentry_dsn = salt.vault.read('secret-operations/global/' ~ business_unit ~ '/sentry-dsn').data.value %}
+{% set smtp_config = salt.vault.read('secret-' ~ business_unit ~ '/' ~ environment ~ '/smtp') %}
```
blarghmatey: Minor nit, you can add the `.data` access to the end of the `vault.read` call so you don't have to include `.data` in the access for each of the other lines. We can also use the `__vault__` syntax in place of the Jinja code.

blarghmatey:
👍

PR #1023, opened 2019-10-01
Changed pgbouncer configs to work with ci restrictions

shaidar:
#### What's this PR do?
Noticed that scheduled config var updates on `discussions-ci` were failing with the following error:
`
 django.db.utils.OperationalError: FATAL:  too many connections for role
`

For ci, we usually use the Heroku postgresql add-on which allows for a max of 20 connections, however with the recent change to pgbouncer for an issuer we came across in production, the number was increased to 500. That caused a problem on the ci side. This PR sets different configs based on environment.


blarghmatey:
:+1:

PR #1028, opened 2019-10-02
Added opsgenie_alias to reduce multiple alerts for same issue

shaidar:
#### What's this PR do?
This add the `opsgenie_alias` to the alerts which I believe @markbreedlove came across and would help reduce some of the noise when an issue is raised.

With certain alerts, we sometimes get inundated with acknowledging the same alert every minute while having to work on resolving the problem. Being able to group the same alerts into a single one would help reduce the noise and allow us to focus on resolving the problem instead of switching back and forth between working on the issue and acknowledging the alert over and over.


blarghmatey:
:+1:

PR #1029, opened 2019-10-02
Changes to resize residential elasticache

shaidar:
#### What's this PR do?
A few weeks ago we experienced a tremendous load on the residential instances caused by what was determined to be a quiz for 8.01L. Another one is scheduled for this coming Monday and we have to decided to make the following changes:
- Add an additional 6 edxapp instances to the existing pool
- Build a new elasticache memcached cluster (can't resize individual nodes) using instances types with more memory and the latest engine as we've seen some cache evictions, albeit a low number, but worth a change.

* Note: While trying to figure out what to name the new cluster given the 20 char limitation faced previously, I noticed that the limitation is now 50 chars. I tried creating a cluster through the console and also confirmed that by checking out [boto3's elasticache library](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elasticache.html#ElastiCache.Client.create_cache_cluster)

### DIFF
--- a/salt/environment_settings.yml
+++ b/salt/environment_settings.yml
@@ -513,13 +513,13 @@ environments:
           purpose: mitx-cas
       elasticache:
         - engine: memcached
-          engine_version: '1.5.16'
-          node_type: cache.r5.large
-          num_cache_nodes: 3
+          engine_version: '1.4.34'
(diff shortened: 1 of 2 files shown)


blarghmatey:
:+1:

PR #1032, opened 2019-10-04
Added memcached config options

shaidar:
#### What's this PR do?
This is tied to the [addition](https://github.com/mitodl/configuration/pull/87) added to the `edx/configuration` fork to be able to configure the memcached server max value from its default of 1MB to 2MB. This is a temporary fix as we need to refactor our elasticache state since the value specified here needs to match or at least not exceed the value set in the parameter group of the elasticache cluster.


blarghmatey:
:+1:

PR #1033, opened 2019-10-08
Updated instance type and count for Redash workers

blarghmatey:
Redash periodically has a flood of tasks which causes a bottleneck and poor user experience. Increasing instance size for more CPU cores and adding to instance count.

N/A

#### What's this PR do?
Increases instance size and count for Redash to scale Celery

N/A


shaidar:
👍

Issue #859, opened 2019-05-08
Define node groups for infrastructure targets in Salt


blarghmatey:
To simplify targeting of sets of instances we should define node groups that allow us to use short names for groupings of servers.

https://docs.saltstack.com/en/latest/topics/targeting/nodegroups.html

markbreedlove:
Are there any particular groups that we need the most?

There are a number of matchers in the topfiles that are pretty eclectic, where I'm not sure what to name them. I assume that some of these should not be defined as nodegroups because they're only relevant in the topfiles; but others seem to be good candidates.

I am working on a number of groups that I know from experience would be useful on the commandline, mainly for working with OCW and the logging cluster.

blarghmatey:
Some good candidates would be things like:
- ovs-rc
- ovs-production
- consul-servers-prod
- consul-servers-qa
- rabbit-servers-prod

Probably best if we all 3 sit down to discuss. Maybe look at bash history to see things that we target frequently.

PR #1036, opened 2019-10-09
Add node groups

markbreedlove:

https://github.com/mitodl/salt-ops/issues/859

#### What's this PR do?

Adds node groups, per the ticket.

A few notes about the groups I've added:

The groups suggested by @blarghmatey in the ticket had hyphen characters, but I've used underscores, thinking the underscore has less risk of being sytactically problematic in some cases, but I don't know if that's warrented. I have a vague memory of this having been the case for me somewhere.

I added some groups that I think would have been useful for me in the past.

I replaced some of the selectors in the topfiles with some of the new nodegroups, but most of the nodegroups aren't used in the topfiles.

After merging this, we'll want to run the `nodegroups.sls` state.


salt/master_utils/nodegroups.sls
```diff
@@ -0,0 +1,47 @@
+manage_nodegroups:
+  file.managed:
```
blarghmatey: This state isn't necessary. You can add the nodegroups definitions as YAML to the salt_master.sls pillar file under the `extra_configs` section (https://github.com/mitodl/salt-ops/blob/redash_scaling/pillar/salt_master.sls#L102). The key name becomes the filename and everything nested under that gets written to the associated configuration file on the salt master via the `master.config` state from the `master-formula` repository.

salt/master_utils/nodegroups.sls
```diff
@@ -0,0 +1,47 @@
...
+              - 'and'
+              - 'G@environment:operations'
+            consul_qa:
+              - 'G@roles:consul_server'
+              - 'and'
+              - 'P@environment:.*-qa$'
+            consul_production:
+              - 'G@roles:consul_server'
+              - 'and'
+              - 'P@environment:.*-production$'
```
blarghmatey: This wouldn't match the `production-apps` environment, and I'm not sure of the usefulness of targeting all consul servers in all production environments at once. Similarly for the rabbit nodegroup below. Typically we would address one environment's cluster at a time.

salt/master_utils/nodegroups.sls
markbreedlove: OK, I've moved it.

salt/master_utils/nodegroups.sls
markbreedlove: I was attempting to define the `consul-servers-prod`, `consul-servers-qa`, and `rabbit-servers-prod` that are mentioned in https://github.com/mitodl/salt-ops/issues/859#issuecomment-533706185

I'm not that familiar with the consul and rabbitmq servers and how they'd typically be configured. Can you clarify what those groups should be; or should I disregard?

markbreedlove: I've just pushed some commits, based on our conversation today...

bootstrap.sh
```diff
@@ -14,6 +14,7 @@ sudo salt-call --local service.restart salt-minion
 sudo salt-key -A -y
 sudo salt -G 'roles:master' state.sls master_utils.libgit
 sudo salt -G 'roles:master' state.sls master_utils.bootstrap
+sudo salt -G 'roles:master' state.sls master_utils.nodegroups
```
blarghmatey: This line can be deleted now since that state got removed

blarghmatey:
:+1:

bootstrap.sh
markbreedlove: Aha, thanks for catching that.

PR #1040, opened 2019-10-23
Tweaks to pki roles

shaidar:
#### What's this PR do?
Due to the way things are rendered, I needed to reformat lists as strings and also made some minor changes to the allowed names passed in with the `common_name` key to generate a cert


blarghmatey:
:+1:

PR #1041, opened 2019-10-28
Fixing list rendering in vault pki

shaidar:
#### What's this PR do?
Vault pki roles is rendering keys that have list values in the following fashion:
`allowed_domains":["['fluentd.service.consul'" ...`

I'm hoping this PR would address the problem as I made changes based on load_filters in [this](https://docs.saltstack.com/en/latest/ref/renderers/all/salt.renderers.jinja.html) doc.


blarghmatey:
:+1:

Issue #981, opened 2019-08-15
Create read replicas for databases used in BI


blarghmatey:
In order to avoid load issues from BI and problems with locks that block migrations we want to have read replicas configured (at least until we get some ETL set up).

- Identify which databases are used in BI (https://bi.odl.mit.edu/data_sources)
- Create read replicas for RDS instances

blarghmatey:
As a corollary to this, we might look at adding support in our orchestrate scripts for RDS deployment to specify when to create a read replica.

blarghmatey:
As part of this work, it would be useful to get a ballpark number on cost for the read replicas so that we can use that when planning next steps for analytical data storage.

markbreedlove:
Starting to look at costs.

markbreedlove:
## Cost analysis

I am avoiding putting detailed information about RDS instances into this issue, which is in a public tracker. Please contact me if you want details.

There are six RDS instances that are probably candidates for read replicas. One RDS instance, for MicroMasters, already has a read replica, which may serve as an example for sizing. Its master is db.m4.large and its replica is db.t2.medium.

The candidate RDS instances range in size, and probably want different-sized replicas. Hypothetically,  the three of them are m5.large or m4.large could have replicas of t2.medium. Two are t2.small, and could have replicas of t2.small, and one is t2.micro, and could have a t2.micro replica. This means:
* 3 x t2.medium
* 2 x t2.small
* 1 x t2.micro

The existing read replica for MicroMasters is paid for by an RI that I can't track down, because there are no RIs listed in our EC2 control panel. It's possible that some of the replicas above could be handled by RIs. However, the monthly on-demand costs would be as follows:
* 3 x t2.medium = $53 x 3 = $159 / month
* 2 x t2.small = $26 x 2 = $52 / month
* 1 x t2.micro = $13 / month

Total: $224 / month

(The fees above are reported by ec2instances.info.)

markbreedlove:
By the way, there's a MySQL datasource in Redash named "TechTV". TechTV is a thing of the past -- do we want a replica of that database, as well?

markbreedlove:
I've merged https://github.com/mitodl/salt-ops/pull/1047

The next step is to create a commit for adding an RDS replica and a parameter group to an existing QA database, with the intention of reverting this commit after we've done our testing.

markbreedlove:
I've made this commit for testing:
[...]

PR #1047, opened 2019-11-04
Add RDS read replicas, update RDS orchestration data

markbreedlove:

https://github.com/mitodl/salt-ops/issues/981

#### What's this PR do?

* Updates `environment_settings.yml` with missing RDS data, and with new values that follow our naming conventions.
* Adds orchestration to create RDS read replicas in `salt/orchestrate/aws/rds.sls`.

### DIFF
--- a/salt/environment_settings.yml
+++ b/salt/environment_settings.yml
@@ -138,6 +138,7 @@ sandbox_mitxpro_versions: &sandbox_mitxpro_versions
   codename: ironwood
   ami_id: ami-80861296
 
+
 environments:
   mitxpro-qa:
     business_unit: mitxpro
@@ -603,6 +604,20 @@ environments:
       - aws
     purposes:
       - micromasters
+    backends:
+      rds:
+        - name: micromasters
+          multi_az: True
+          engine: postgres
+          vault_plugin: postgresql-database-plugin
+          db_instance_class: db.m4.large
+          allocated_storage: 100
+          public_access: True
+          engine_version: '9.6.11'
+          purpose: micromasters
+          mount_point: postgresql-micromasters
+          replica:
+            db_instance_class: db.t2.medium
   bootcamps:
     business_unit: bootcamps
     network_prefix: '10.11'
@@ -614,6 +629,20 @@ environments:
       - aws
     purposes:
       - bootcamps
+    backends:
+      rds:
+        - name: bootcamps
+          multi_az: True
+          engine: postgres
+          vault_plugin: postgresql-database-plugin
+          db_instance_class: db.t2.micro
+          allocated_storage: 50
+          public_access: True
+          engine_version: '9.6.11'
+          purpose: bootcamps
+          mount_point: postgresql-bootcamps
+          replica:
+            db_instance_class: db.t2.micro
   operations:
     business_unit: operations
     network_prefix: '10.0'
@@ -688,11 +717,14 @@ environments:
           multi_az: True
           engine: postgres
           vault_plugin: postgresql-database-plugin
+          mount_point: postgres-operations-redash
           db_instance_class: db.t2.small
           allocated_storage: 100
           public_access: False
-          engine_version: '10.1'
+          engine_version: '10.6'
           purpose: redash
+          replica:
+            db_instance_class: db.t2.small
         - name: techtvcopy
           multi_az: False
           engine: mariadb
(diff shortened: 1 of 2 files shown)


blarghmatey:
This is looking good so far. I think the only piece remaining is to create the custom parameter group as part of the orchestrate and assign it to the RDS instance that gets created.

blarghmatey:
It's also worth noting here which existing instances need renaming so that we can highlight the overall deployment process here.

salt/orchestrate/aws/rds.sls
```diff
@@ -59,6 +61,22 @@ set_{{ name }}_master_password_in_vault:
 {% set master_pass = master_pass.data.value %}
 {% endif %}
 
+{# If engine is postgres and version is 10, then use major version only for
+   parameter group family. All others use major.minor. #}
+{% if engine == 'postgres' and engine_version.startswith('10') %}
```
blarghmatey: This is fragile as the current version of Postgres is now at 12, so it will fail for versions 11, 12, etc. Maybe verify on something like `len(engine_version.split('.')[0]) > 1`

salt/orchestrate/aws/rds.sls
```diff
@@ -59,6 +61,22 @@ set_{{ name }}_master_password_in_vault:
...
+{% set fam_version = engine_version.partition('.')[0] %}
+{% else %}
+{% set fam_version = engine_version.rpartition('.')[0] %}
+{% endif %}
+{% set db_parameter_group_family = engine ~ fam_version %}
+
+
+create_{{ ENVIRONMENT }}_{{ name }}_parameter_group:
+  boto_rds.parameter_present:
```
blarghmatey: In addition to creating the parameter group we need to apply it to the RDS instance as well.

salt/orchestrate/aws/rds.sls
markbreedlove: I've just pushed a commit that I think will fix this. It uses `int(engine_version.partition('.')[0]) > 9`.

salt/orchestrate/aws/rds.sls
markbreedlove: I've just pushed a commit that will apply it in the `boto_rds.present` state.

salt/orchestrate/aws/rds.sls
```diff
@@ -57,14 +61,35 @@ set_{{ name }}_master_password_in_vault:
...
+   parameter group family. All others use major.minor. #}
+{% if engine == 'postgres' and int(engine_version.partition('.')[0]) > 9 %}
+{% set fam_version = engine_version.partition('.')[0] %}
+{% else %}
+{% set fam_version = engine_version.rpartition('.')[0] %}
+{% endif %}
+{% set db_parameter_group_family = engine ~ fam_version %}
+{% set db_parameter_group_name = ENVIRONMENT ~ '_' ~ engine ~ '_' ~ name ~ '_parameters' %}
+
+{% if parameters %}
```
blarghmatey: We always want the parameter group to be created. If there needs to be a conditional it should only apply to setting the customized parameters on the parameter group (line 79). This also means unconditionally applying the parameter group to the RDS instance on line 91.

salt/orchestrate/aws/rds.sls
```diff
@@ -57,14 +61,35 @@ set_{{ name }}_master_password_in_vault:
...
 
+{# If engine is postgres and version is 10, then use major version only for
+   parameter group family. All others use major.minor. #}
+{% if engine == 'postgres' and int(engine_version.partition('.')[0]) > 9 %}
+{% set fam_version = engine_version.partition('.')[0] %}
+{% else %}
+{% set fam_version = engine_version.rpartition('.')[0] %}
+{% endif %}
+{% set db_parameter_group_family = engine ~ fam_version %}
+{% set db_parameter_group_name = ENVIRONMENT ~ '_' ~ engine ~ '_' ~ name ~ '_parameters' %}
```
blarghmatey: For the case where we are upgrading an RDS instance we might want to use the parameter group family in the name of the group, replacing `engine` with `db_parameter_group_family`.

salt/environment_settings.yml
```diff
@@ -138,6 +138,31 @@ sandbox_mitxpro_versions: &sandbox_mitxpro_versions
   codename: ironwood
   ami_id: ami-80861296
 
+rds_postgres_parameter_group_defaults: &rds_postgres_parameter_group_defaults
```
blarghmatey: I think it's better for these defaults to live in the orchestrate state as a Jinja mapping and then if there are additional custom parameters for a given instance then we can specify them at that point and merge them into the mapping as part of the orchestrate state. Having them defined here in YAML I think adds extra bloat to this file and creates unnecessary visual noise.

Having it as a mapping also simplifies enforcing that if there are version specific overrides then we can define them there and more atomically ensure that they get applied rather than manually tracking version updates for the engine and the parameter group in this file.

salt/environment_settings.yml
```diff
@@ -602,6 +643,23 @@ environments:
...
+          vault_plugin: postgresql-database-plugin
+          db_instance_class: db.t2.micro
+          allocated_storage: 50
+          public_access: True
+          engine_version: '9.6.11'
+          purpose: bootcamps
+          mount_point: postgresql-bootcamps
+          parameters:
+            <<: *rds_postgres_9_parameter_group_defaults
+            rds.force_ssl: 1
```
blarghmatey: This should be a default parameter as it is useful even inside a VPC.

salt/environment_settings.yml
markbreedlove: I saw the yaml merge technique being used in `pillar/devstack.sls` and was emulating how it was done there. Can you point me at an example of how to do it with the jinja map method?

blarghmatey: In Jinja it's essentially just a Python dict so you would use the `.update` method to merge the dicts.

e.g.
```
{% set parameter_map = {'postgres_default_parameters': {'character_set_encoding': 'utf8'},
'mysql_default_parameters': {'character_set_encoding': 'utf8'}} %}
{% set extra_params = db_config.pop('parameters', {}) %}
{% do paramater_map.update(extra_params) %}
```

markbreedlove: OK, thanks 👍

salt/orchestrate/aws/rds.sls
markbreedlove: I've changed this in commit ace521b.

salt/orchestrate/aws/rds.sls
markbreedlove: I've changed this in ace521b.

salt/environment_settings.yml
markbreedlove: I've changed this in commits ff310fb and 7b39348.

salt/environment_settings.yml
markbreedlove: I've added this to the PostgreSQL defaults in ff310fb.

blarghmatey:
:+1:

markbreedlove:
The instances that will need renaming are as follows:

* Open Discussions: `production-apps-rds-postgresql-opendiscussions` -> `production-apps-rds-postgres-opendiscussions`
* Open Discussions Reddit: `production-apps-rds-postgresql-reddit` -> `production-apps-rds-postgres-reddit`
* Bootcamps: `bootcamps-rds-postgresql` -> `bootcamps-rds-postgres-bootcamps`

We're skipping the MicroMasters database for now, because it already has a read replica, but it should at some point be renamed from `micromasters-db` to `micromasters-rds-postgres-micromasters`.

I believe that the procedure for the ones that need renaming will be:

1. Find a suitable maintenance window when the app can withstand a few minutes' downtime
1. Rename the RDS database in the control panel. This appears to change the hostname.
1. Change the Heroku app's environment variable for the database endpoint, resulting in redeployment
1. Run orchestration, which should just create a read replica and assign a custom parameter group
1. Change the data source in Redash

PR #1056, opened 2019-11-07
Rearrange preset ids

shaidar:
Removed the last preset_id that was added recently as it's only audio and moved the other two to the beginning as that impacts the order of the dropdown selector.


blarghmatey:
:+1:

PR #1060, opened 2019-11-15
Fix issue with old db backend roles

shaidar:
#### What's this PR do?
This should fix the issue whereby the roles configured for the old vault database backend are properly populated


blarghmatey:
:+1:

PR #1063, opened 2019-11-18
Updated Consul DNS for operations services to simplify discovery

blarghmatey:
With the introduction of the operations-qa environment there were a number of hardcoded DNS records that would fail to resolve for QA services. Added a query template to allow for querying operations and operations-qa environments based on what is reachable to the given client, removing the need to generate a mapping of which envs talk to which other envs.

#1050 

#### What's this PR do?
Adds a query template for Consul DNS to simplify operations services discovery

Apply the template and run a dig command for the relevant services

### DIFF
--- a/salt/consul/query_template.sls
+++ b/salt/consul/query_template.sls
@@ -18,3 +18,29 @@ create_query_template_for_nearest_service:
     - match_type: pcre
     - decode: True
     - raise_error: False
+
+create_query_template_for_ops_service:
+  http.query:
+    - name: 'http://consul.service.consul:8500/v1/query'
+    - method: POST
+    - data: >-
+        {
+          "Name": "operations",
+          "Service": {
+            "Failover": {
+              "Datacenters": [
+                "operations",
+                "operations-qa"
+              ]
+            },
+            "Service": "${match(1)}"
+          },
+          "Template": {
+            "Regexp": "^operations-(.*?)$",
+            "Type": "name_prefix_match"
+          }
+        }
+    - match: (ID|existing)
+    - match_type: pcre
+    - decode: True
+    - raise_error: False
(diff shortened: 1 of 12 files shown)


salt/consul/query_template.sls
```diff
@@ -18,3 +18,29 @@ create_query_template_for_nearest_service:
...
+            "Failover": {
+              "Datacenters": [
+                "operations",
+                "operations-qa"
+              ]
+            },
+            "Service": "${match(1)}"
+          },
+          "Template": {
+            "Regexp": "^operations-(.*?)$",
```
shaidar: Is this meant to match against `operations-qa` and `operations`? Just wondering about the `-`

blarghmatey: This is intended to allow for DNS records of the form `operations-elasticsearch.query.consul` similar to our `nearest-elasticsearch.query.consul` prepared query. That allows us to avoid having to write conditionals in our pillar data for things like `elasticsearch.service.operations.consul` vs `elasticsearch.service.operations-qa.consul`.

shaidar: And consul will know which `operations` (prod or qa) to point to?

blarghmatey: The way that will work is that a given VPC will only be connected to the appropriate operations environment, meaning that the query will only be able to resolve from the one network. This saves us from having to embed that logic in our pillar data and push it to the Consul layer.

shaidar: Are there any safeguards in place in case the secgroups/routing gets screwed up? Might be overkill, but just thinking about potential issues

blarghmatey: In other words, the QA environments will be updated to peer their networks with operations-qa, meaning that the query will then start resolving for the fluentd/elasticsearch/etc. services from the operations-qa environment instead of the existing operations environment.

shaidar:
👍

salt/consul/query_template.sls
blarghmatey: As far as safeguards, it will try to resolve one DC and then fail over to the other if it can't reach it. It doesn't really handle whether the secgroups are configured properly, but that's why we're testing in QA first :)

PR #1064, opened 2019-11-20
Added configuration for Open youtube integration

rhysyngsun:
Fixes https://github.com/mitodl/open-discussions/issues/2349

#### What's this PR do?
Adds a variable for configuring per-environment developer keys. Please double-check that I wrote the vault variable correctly based on how you guys structure those

For each environment, we'll need separate developer keys, instructions on how to create these are here:

https://developers.google.com/youtube/v3/getting-started#before-you-start

They can be testing with this curl command (bad credentials will 403, otherwise you should see data):

```bash
curl "https://www.googleapis.com/youtube/v3/channels?id=UCEBb1b_L6zDS3xTUrIALZOw&part=snippet&key=$YOUR_API_KEY"
```


blarghmatey:
:+1:

PR #1072, opened 2019-11-22
Fixed missing celery task logs

shaidar:
#### What's this PR do?
Noticed that the celery task logs weren't showing up `/var/log/uwsgi/apps/odl-video-service.log`. After reading docs and testing the fix by adding `-l LOG_LEVEL` to the launch daemon, celery task logs started showing up.


blarghmatey:
:+1:

PR #1074, opened 2019-11-26
Updated xPro/MM hosts for data etl to use SDB/consul

rhysyngsun:
N/A

#### What's this PR do?
Updates the host values for external integrations into xPro/MM so that they can be configured without code changes. This will support configuring these against production in the short term but permit changes to other environments for testing purposes as well.

Deploy the code.

SDB values for all environments should be `xpro.mit.edu` and `micromasters.mit.edu`


blarghmatey:
:+1:

PR #1079, opened 2019-12-04
Adjust nginx max_body_size

shaidar:
In response to the [sentry issue](https://sentry.io/organizations/mit-office-of-digital-learning/issues/1369981344/?project=216201) reported 

#### What's this PR do?
This increases the default of 1mb to 20mb. I picked 20 cause that's what we have in edx, not that is any indication of a good measure, but mainly for parity. The [setting](http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) can be in either the http, server, or location contexts and I chose the one that's most restrictive.


blarghmatey:
:+1:

PR #1080, opened 2019-12-04
Rotate Kibana's logfile

markbreedlove:
Rotate Kibana's `/var/log/kibana.log` logfile, which is currently unhandled.


blarghmatey:
:+1:

PR #1081, opened 2019-12-04
First pass at cleaning fluentd aggreator for qa

shaidar:
#### What's this PR do?
This is a first pass at removing some of the bits that aren't relevant to the fluentd aggregators in qa. It doesn't include yet anything dealing with pki as I would like to get things flowing first and then looking into adding that piece


blarghmatey:
:+1:

PR #1082, opened 2019-12-06
Tweaks to edx.tracking

shaidar:
#### What's this PR do?
This is in response to Ike's comment about the tracking logs in BigQuery missing the time stamp. The logs showing up in Kibana do have the `@timestamp` in but they don't appear in the S3 files. Also noticed that the xpro-production s3 bucket doesn't have the latest tracking logs.


blarghmatey:
That will make sure they get uploaded hourly but doesn't seem to handle adding the timestamp. In ES the time attribute gets added by the ES index

blarghmatey:
:+1:

PR #1083, opened 2019-12-06
Ensure same version for Elastic Stack products

markbreedlove:
Per recent changes to elastic-stack-formula, use one version variable for all Elastic Stack products.

See https://github.com/mitodl/elastic-stack-formula/pull/9


blarghmatey:
Other than my comment above this is :+1:

PR #1085, opened 2019-12-09
Moving established settings out of next-residential and adding pdf xblock

blarghmatey:
N/A

#### What's this PR do?
Moves edX settings into the residential pillar, adds the PDF xblock


shaidar:
👍

PR #1090, opened 2019-12-10
Splitting schedules on the master for prod vs. qa

blarghmatey:


pillar/master/production_schedule.sls
```diff
@@ -47,3 +21,19 @@ schedule:
     function: state.sls
     args:
       - edx.maintenance_tasks
+  scan_for_expiring_vault_leases:
+    maxrunning: 1
+    when: Monday 9:00am
+    function: vault.scan_leases
+    kwargs:
+      time_horizon: {{ ONE_WEEK }}
```
shaidar: Wouldn't this be undefined?

shaidar:
Just a nitpick but why is the order different for scanning vault leases in production and qa?

blarghmatey:
Because that's where I copy pasted it :)

shaidar:
👍

PR #1091, opened 2019-12-10
Set Elasticsearch cluster.initial_master_nodes for v7

markbreedlove:
In the Elastic Stack for Elasticsearch pillar, set cluster.initial_master_nodes if the version of Elastic Stack is 7 or above.

This setting has been introduced with version 7 and the cluster will not run without it.

See https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#_cluster_bootstrapping_is_required_if_discovery_is_configured

I'm not sure if I've got the discovery of the cluster hostnames right. The new convention in ES 7 is for the node name to equal the hostname. I imitated some code that I saw elsewhere that employs `mine.get` for a similar purpose.

I could probably improve the `if` statement that evaluates the version number. It's hardcoded to check for version 7 specifically, which won't be right when version 8 comes out.

### DIFF
--- a/pillar/elastic_stack/elasticsearch/init.sls
+++ b/pillar/elastic_stack/elasticsearch/init.sls
@@ -1,4 +1,13 @@
 {% set ENVIRONMENT = salt.grains.get('environment', 'rc-apps') %}
+{% set es_hostnames = [] %}
+{% for host, addr in salt.saltutil.runner(
+    'mine.get',
+    tgt='G@roles:elasticsearch and G@environment:{}'.format(ENVIRONMENT),
+    fun='grains.item',
+    tgt_type='compound').items() %}
+{% do es_hostnames.append(host) %}
+{% endfor %}
+{% version_data = salt.cp.get_file_str('salt://elastic_stack/version_{}.sls'.format(ENVIRONMENT))|load_yaml %}
 
 elastic_stack:
   elasticsearch:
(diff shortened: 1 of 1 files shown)


markbreedlove:
I've changed the way it determines the Elasticsearch version, but I'm not sure it will work for a fresh installation, where the package has not been installed yet.

pillar/elastic_stack/elasticsearch/init.sls
```diff
@@ -1,13 +1,5 @@
...
-{% set es_hostnames = [] %}
-{% for host, addr in salt.saltutil.runner(
-    'mine.get',
-    tgt='G@roles:elasticsearch and G@environment:{}'.format(ENVIRONMENT),
-    fun='grains.item',
-    tgt_type='compound').items() %}
-{% do es_hostnames.append(host) %}
-{% endfor %}
-{% version_data = salt.cp.get_file_str('salt://elastic_stack/version_{}.sls'.format(ENVIRONMENT))|load_yaml %}
+{% set version = salt.pkg.version('elasticsearch').split('.')[0]|int %}
```
blarghmatey: To guard against the case where the package isn't installed you can leave off the `int` filter here, and then below update the conditional to something like:
`if version and int(version) > 6`

markbreedlove: Do you think that commit 9515ee7 will do it?

PR #1092, opened 2019-12-10
Make separate pillar for QA Elasticsearch logging cluster

markbreedlove:
Have a separate pillar for the QA environment's Elasticsearch logging cluster. We may want to have a different version of Elasticsearch, different numbers or sizes of nodes, or just play around with different settings in general.

I've removed the evaluation of the installed Elasticsearch version in the QA pillar file.


blarghmatey:
:+1:

PR #1097, opened 2019-12-13
Changed xpro sheets task timing settings

gsidebo:
Reset xpro sheets settings so that tasks run every hour by default in all environments, and the wait time in between successive tasks in 2 minutes


blarghmatey:
:+1:

PR #1098, opened 2019-12-17
Adding values for SSL host override in xPRO

blarghmatey:
N/A

#### What's this PR do?
Adds settings to override the host used in SSL redirects for xPRO


rhysyngsun:
:+1: lgtm

PR #1103, opened 2019-12-18
Mods for fluentd qa server and pki

shaidar:
#### What's this PR do?
This is a basic first pass for the fluentd qa changes for enabling TLS using our pki infra. Additionally, some of the other changes are related to the newer version of fluentd. Last but not least, since this is a pillar file for the qa aggregators, I took out the tracking log shipping to s3 since we only need to do that for production.

### DIFF
--- a/pillar/fluentd/server_operations_qa.sls
+++ b/pillar/fluentd/server_operations_qa.sls
@@ -1,7 +1,7 @@
 {% set ENVIRONMENT = salt.grains.get('environment') %}
 {% set mailgun_webhooks_token = salt.vault.read('secret-operations/{}/mailgun_webhooks_token'.format(ENVIRONMENT)).data.value %}
 {% set es_hosts = 'operations-elasticsearch.query.consul' %}
-{% set cert = salt.vault.cached_write('pki-intermediate-operations/issue/fluentd-server', common_name='{}'.format(es_hosts)) %}
+{% set cert = salt.vault.cached_write('pki-intermediate-operations/issue/fluentd-server', common_name='operations-fluentd.query.consul') %}
 {% set fluentd_cert_path = '/etc/fluent/fluentd.crt' %}
 {% set fluentd_key_path = '/etc/fluent/fluentd.key' %}
 {% set ca_cert_path = '/etc/fluent/ca.crt' %}
--- a/pillar/fluentd/server_operations_qa.sls
+++ b/pillar/fluentd/server_operations_qa.sls
@@ -1,20 +1,9 @@
-{% set app_name = "fluentd-aggregators" %}
-{% set fluentd_shared_key = salt.vault.read('secret-operations/operations-qa/fluentd_shared_key').data.value %}
-{% set heroku_http_token = salt.vault.read('secret-operations/operations-qa/heroku_http_token').data.value %}
-{% set mailgun_webhooks_token = salt.vault.read('secret-operations/operations-qa/mailgun_webhooks_token').data.value %}
-{% set odl_wildcard_cert = salt.vault.read('secret-operations/global/odl_wildcard_cert') %}
-{% import_yaml 'fluentd/fluentd_directories.yml' as fluentd_directories %}
-
-schedule:
-  refresh_{{ app_name }}_configs:
-    # Needed to ensure that S3 credentials remain valid
-    days: 5
-    function: state.sls
-    args:
-      - fluentd.config
+{% set ENVIRONMENT = salt.grains.get('environment') %}
+{% set mailgun_webhooks_token = salt.vault.read('secret-operations/{}/mailgun_webhooks_token'.format(ENVIRONMENT)).data.value %}
+{% set es_hosts = 'operations-elasticsearch.query.consul' %}
+{% set cert = salt.vault.cached_write('pki-intermediate-operations/issue/fluentd-server', common_name='{}'.format(es_hosts)) %}
 
 fluentd:
-  persistent_directories: {{ fluentd_directories|tojson }}
   overrides:
     nginx_config:
       server_name: logs-qa.odl.mit.edu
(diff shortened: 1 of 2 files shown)


pillar/fluentd/server_operations_qa.sls
```diff
@@ -1,38 +1,30 @@
...
     nginx_config:
       server_name: logs-qa.odl.mit.edu
       cert_file: log-input.crt
       key_file: log-input.key
       cert_contents: __vault__::secret-operations/global/odl_wildcard_cert>data>value
       key_contents: __vault__::secret-operations/global/odl_wildcard_cert>data>key
+  cert:
+    - fluentd_cert: {{ cert.data.certificate }}
+    - fluentd_key: {{ cert.data.private_key }}
+    - ca_cert: {{ cert.data.issuing_ca }}
```
blarghmatey: Did you intend to format this as a list and just iterate over it blindly? If you wanted to access these values by key path then you'll want to format it as a dictionary.

pillar/fluentd/server_operations_qa.sls
```diff
@@ -1,38 +1,30 @@
...
   plugins:
     - fluent-plugin-heroku-syslog-http
-    - fluent-plugin-s3
     - fluent-plugin-elasticsearch
   proxied_plugins:
     - route: heroku-http
       port: 9000
-      token: {{ heroku_http_token }}
+      token: __vault__::secret-operations/{{ ENVIRONMENT }}/heroku_http_token>data>value
```
blarghmatey: Recognizing that these values are already set, should we make this a `gen_if_missing`?

pillar/fluentd/server_operations_qa.sls
```diff
@@ -95,10 +89,11 @@ fluentd:
                 directive_arg: '**'
                 attrs:
                   - '@id': es_outbound
-                  - '@type': elasticsearch_dynamic
+                  - '@type': elasticsearch
+                  - scheme: https
```
blarghmatey: Will this work? I don't believe the logging ES hosts are running as HTTPS.

pillar/fluentd/server_operations_qa.sls
```diff
@@ -95,10 +89,11 @@ fluentd:
                 directive_arg: '**'
                 attrs:
                   - '@id': es_outbound
-                  - '@type': elasticsearch_dynamic
+                  - '@type': elasticsearch
```
blarghmatey: Does the newer elasticsearch builtin support the dynamic functionality now? We had used that plugin to allow for programmatically generating the logstash_prefix.

pillar/fluentd/server_operations_qa.sls
```diff
@@ -48,40 +40,42 @@ fluentd:
...
-                  - private_key_path: '/etc/ssl/certs/log-input.key'
-                  - private_key_passphrase: ''
+                - directive: security
+                  attrs:
+                    - self_hostname: {{ salt.grains.get('external_ip') }}
+                - directive: transport
+                  directive_arg: tls
+                  attrs:
+                    - cert_path: /etc/fluent/fluentd.crt
+                    - private_key_path: /etc/fluent/fluentd.key
```
blarghmatey: If we're going to put the keys/certs under /etc/fluent then we should probably have a state to ensure that ownership and access to that directory is only available to the user running the fluent process.

pillar/fluentd/server_operations_qa.sls
shaidar: Think we need to have it already set as we need to pass that token when we create the drain for that heroku app

pillar/fluentd/server_operations_qa.sls
shaidar: Reverted, but looks like we should be maybe handling things on the ES side instead based on their [docs](https://github.com/uken/fluent-plugin-elasticsearch#dynamic-configuration)

pillar/fluentd/server_operations_qa.sls
shaidar: Yeah was planning on doing that bit through the fluentd formula that will be creating those files.

pillar/fluentd/server_operations_qa.sls
shaidar: I was hoping that we'd leverage the pki infra for use on those ES hosts and thus be able to use `https`.

pillar/fluentd/server_operations_qa.sls
shaidar: Changed

pillar/fluentd/server_operations_qa.sls
```diff
@@ -1,38 +1,39 @@
...
-  refresh_{{ app_name }}_configs:
-    # Needed to ensure that S3 credentials remain valid
-    days: 5
-    function: state.sls
-    args:
-      - fluentd.config
+{% set ENVIRONMENT = salt.grains.get('environment') %}
+{% set mailgun_webhooks_token = salt.vault.read('secret-operations/{}/mailgun_webhooks_token'.format(ENVIRONMENT)).data.value %}
+{% set es_hosts = 'operations-elasticsearch.query.consul' %}
+{% set cert = salt.vault.cached_write('pki-intermediate-operations/issue/fluentd-server', common_name='{}'.format(es_hosts)) %}
```
blarghmatey: Why are we setting the elastic search hosts as the name for the clients certs?

shaidar: This would be the cert that the aggregators are using and the fluentd clients would be shipping their logs to that hostname. Unless I'm missing something, but believe that the name on the cert that the aggregators are using should match the hostname that the fluentd client are using to ship logs, no?

blarghmatey: Right, so shouldn't it be operations-fluentd.query.consul?

shaidar: That was a `duh` moment. Fixed

blarghmatey:
:+1:

Issue #1106, opened 2019-12-20
xPRO: set SHOW_UNREDEEMED_COUPON_ON_DASHBOARD to true in production


pdpinch:
Now that https://github.com/mitodl/mitxpro/pull/1356 is merged, we should turn on the feature.

pdpinch:
I believe the key needs to be added in 3 places:

1. CI: https://github.com/mitodl/salt-ops/blob/master/pillar/heroku/xpro.sls#L6
2. RC: https://github.com/mitodl/salt-ops/blob/master/pillar/heroku/xpro.sls#L33
3. production: https://github.com/mitodl/salt-ops/blob/master/pillar/heroku/xpro.sls#L60

PR #1108, opened 2019-12-20
add feature flag SHOW_UNREDEEMED_COUPON_ON_DASHBOARD

tasawernawaz:
Actual issue: https://github.com/mitodl/mitxpro/pull/1356
Fixes https://github.com/mitodl/salt-ops/issues/1106

#### What's this PR do?
add feature flag SHOW_UNREDEEMED_COUPON_ON_DASHBOARD

- Create coupon codes.
- Upload CSV with user's emails to bulk enroll.
- User(s) listed in the CSV should see a new section on the dashboard with a list of unredeemed codes.

### DIFF
--- a/pillar/heroku/xpro.sls
+++ b/pillar/heroku/xpro.sls
@@ -26,7 +26,6 @@
       'MITXPRO_BASE_URL': 'https://xpro-ci.odl.mit.edu',
       'MITXPRO_SECURE_SSL_HOST': 'xpro-ci.odl.mit.edu',
       'ENABLE_ORDER_RECEIPTS': True,
-      'SHOW_UNREDEEMED_COUPON_ON_DASHBOARD': True,
       'vault_env_path': 'rc-apps',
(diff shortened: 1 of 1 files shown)


pillar/heroku/xpro.sls
```diff
@@ -171,6 +174,7 @@ heroku:
     SHEETS_DATE_TIMEZONE: America/New_York
     SHEETS_MONITORING_FREQUENCY: 3600
     SHEETS_TASK_OFFSET: 120
+    SHOW_UNREDEEMED_COUPON_ON_DASHBOARD: {{ env_data.SHOW_UNREDEEMED_COUPON_ON_DASHBOARD }}
```
blarghmatey: If we want it true in all environments then you can just set it to that value on this line, rather than putting it in the per-env dictionaries. Those are only used for cases where we need a different value between the different environments.

blarghmatey:
:+1:

PR #1109, opened 2019-12-20
Added pruning of backups to backup scripts

blarghmatey:
We have been keeping all backups which ends up ballooning the backup repo and slowing down backup/restore. This adds commands to prune old backups beyond 5 weeks to keep things in check.


markbreedlove:
:+1:

PR #1110, opened 2019-12-23
Pruning list of services no longer using CAS

blarghmatey:
With our move to directly integrating with the edX SAML implementation there is no need to maintain this list of URLs for support with CAS. This serves as one step toward removing CAS entirely.

#610 

#### What's this PR do?
Removes residential URLs from list of services allowed in our CAS deployment


shaidar:
👍

PR #1111, opened 2019-12-26
Broke out some directives into separate files

shaidar:
#### What's this PR do?
- Updates consul port to match fluentd
- Breaks out fluentd logging directive into its own file
- Breaks out match all pattern including the tls config into its own file
- Updates fluentd-qa aggregator config based on testing
- Updates mitx fluentd client based on testing

Note: Still additional tweaks required (probably changing the jinja files into sls files) and adding some filtering but wanted to get this reviewed and deployed and then do some more tweaking.


pillar/fluentd/fluentd_log.sls
```diff
@@ -0,0 +1,15 @@
+fluentd:
+  configs:
+    - name: fluentd_log
```
blarghmatey: If we're applying these pillar files to the same host then the list items will override each other. To be able to split them out like this we'll need to change the formula to take pillar data in the format:

```
fluentd:
  configs:
    fluentd_log:
      settings:
         foo: bar
     math_all_tls:
       settings:
         baz: foo
```

pillar/fluentd/mitx.sls
```diff
@@ -7,9 +7,8 @@ fluentd:
       - ruby2.3
       - ruby2.3-dev
       - build-essential
-  plugins:
-    - fluent-plugin-secure-forward
   configs:
+    - '@include': 'fluentd_log'
```
blarghmatey: This won't work at the top level of the `configs` object. If we want it included for all fluentd configs we would add it to the `init.sls` file, otherwise we would need to move it down into the `settings:` object under the edx config.

pillar/fluentd/server_operations_qa.sls
```diff
@@ -45,54 +45,81 @@ fluentd:
             - '@type': monitor_agent
             - bind: 127.0.0.1
             - port: 24220
+    - name: fluentd_server_log
```
blarghmatey: I'm confused about the reason for having the `FLUENT_LOG` label here and in the other SLS file.

shaidar: This is a way to capture fluentd logs (https://docs.fluentd.org/deployment/logging#capture-fluentd-logs)

pillar/fluentd/server_operations_qa.sls
```diff
@@ -45,54 +45,81 @@ fluentd:
...
+                        attrs:
+                          - flush_interval: '10s'
+                          - flush_thread_count: 2
     - name: elasticsearch
       settings:
         - directive: source
           attrs:
             - '@id': heroku_logs_inbound
             - '@type': heroku_syslog_http
-            - tag: heroku_logs
```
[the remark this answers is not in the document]
shaidar: Re-added

pillar/fluentd/server_operations_qa.sls
blarghmatey: My confusion was because of the fact that you had the label twice. Once here and once in the other config which you moved to the mitx config. I understand it now :)

blarghmatey:
:+1:

PR #1112, opened 2019-12-30
Removing logzio configuration as it is no longer being used

blarghmatey:
The logz.io account was closed due to inactivity because the course team wasn't using it.


shaidar:
👍

PR #1114, opened 2020-01-02
Add mitxpro redirect

markbreedlove:

https://github.com/mitodl/mitxpro/issues/1370

#### What's this PR do?

It adds an Nginx virtual host for redirecting mitxpro.mit.edu to xpro.mit.edu. The virtual host is added to the same Nginx server that is used for `amps-web.amps.ms.mit.edu`. I have added a letsencrypt certificate for mitxpro.

### DIFF
--- a/pillar/nginx/mitxpro_redirect.sls
+++ b/pillar/nginx/mitxpro_redirect.sls
@@ -1,0 +1,40 @@
+nginx:
+  ng:
+    servers:
+      managed:
+        mitxpro-redirect:
+          enabled: True
+          config:
+            - server:
+              - server_name: mitxpro.mit.edu
+              - listen:
+                  - 80
+              - listen:
+                  - 443
+                  - ssl
+              - listen:
+                - '[::]:80'
+                - '[::]:443'
+                - ssl
+                # The certificate uses subject alternative names, including mitxpro.mit.edu.
+                - ssl_certificate: /etc/letsencrypt/live/amps-web.amps.ms.mit.edu/cert.pem
+                - ssl_certificate_key: /etc/letsencrypt/live/amps-web.amps.ms.mit.edu/privkey.pem
+                - ssl_stapling: 'on'
+                - ssl_stapling_verify: 'on'
+                - ssl_session_timeout: 1d
+                - ssl_session_tickets: 'off'
+                - ssl_protocols:
+                    - TLSv1.2
+                    - TLSv1.3
+                - ssl_ciphers: "TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256\
+                    :DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384\
+                    :ECDHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-SHA256:DHE-RSA-AES128-SHA256\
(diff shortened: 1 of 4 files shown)


pillar/nginx/mitxpro_redirect.sls
```diff
@@ -0,0 +1,39 @@
...
+                - 80
+            - listen:
+                - 443
+                - ssl
+            - listen:
+              - '[::]:80'
+              - '[::]:443'
+              - ssl
+              # The certificate uses subject alternative names, including mitxpro.mit.edu.
+              - ssl_certificate: /etc/letsencrypt/live/amps-web.amps.ms.mit.edu/cert.pem
```
blarghmatey: Should this be set to a cert for the mtixpro domain?

pillar/nginx/mitxpro_redirect.sls
```diff
@@ -0,0 +1,39 @@
...
+              - ssl_protocols:
+                  - TLSv1.2
+                  - TLSv1.3
+              - ssl_ciphers: "TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256\
+                  :DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384\
+                  :ECDHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-SHA256:DHE-RSA-AES128-SHA256\
+                  :ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256"
+              - ssl_prefer_server_ciphers: 'on'
+              - resolver: 1.1.1.1
+              - location ~ /.well-known:
```
blarghmatey: It shouldn't be necessary to do a regex match on this path. Should be able to just have it be `location /.well-known`. Usually I set the parameters for that location to use `alias` e.g.

```
              - location /.well-known/:
                  - alias: /var/www/.well-known/
```

pillar/nginx/mitxpro_redirect.sls
```diff
@@ -0,0 +1,39 @@
...
+                - allow: all
+              - location /certificates/
+                - rewrite ^(.*)$ https://courses.edx.org$1 permanent
```
blarghmatey: A possibly simpler option is to use `return 301 https://courses.edx.org$request_uri`

blarghmatey: That removes the need to do a regex match

pillar/nginx/mitxpro_redirect.sls
markbreedlove: I think that I planned it this way so that the certificate could be acquired through Letsencrypt prior to the DNS change. The domain `mitxpro.mit.edu` will still point to the edX servers when we first request the certificate. If we make the request for `amps-web.amps.ms.mit.edu` with a SAN of `mitxpo.mit.edu` I expect it to succeed because `amps-web.amps.ms.mit.edu` resolves correctly.

I'm not sure how much the downtime matters. If we get the certificate for `mitxpro.mit.edu` then there will be some time between when DNS cuts over and then we run the `certbot` script when someone could get an SSL error.

markbreedlove: It's been a while since I worked on this last, and another thing that I admit I can't figure out now is why I added the `letsencrypt` state to `salt/top.sls` but there's no pillar data to really configure the certificate. It could be that I just got distracted and never finished that work.

markbreedlove:
per @blarghmatey , OK to merge now.

PR #1116, opened 2020-01-02
Added ca_chain config for fluentd

shaidar:
#### What's this PR do?
In order not to simplify our work with PKI and not have to include the entire chain in the fluentd client config, we opted to add the chain to the local ca store. This change, in conjunction with one to fluentd-formula will write the ca_chain.crt file to the trusted store path and update that config.


blarghmatey:
:+1:

PR #1118, opened 2020-01-03
Additional minor tweaks

shaidar:
#### What's this PR do?
Moved fluentd log under the same file instead of duplicating another match statement that would be required.


blarghmatey:
:+1:

PR #1120, opened 2020-01-06
Fix bad X-Forwarded-For header in Kibana Nginx config

markbreedlove:
Fix a bad value written into an Nginx `proxy_set_header` directive for the `X-Forwarded-For` header. The bad value (with the contents of a TLS CA certificate) must have been the result of a bad copy-and-paste operation, in commit c3ab7c399b94d53ca9a9c8d725a990a9152a09ae ("Update Elastic Stack pillars and states").


blarghmatey:
:+1:

PR #1121, opened 2020-01-06
Additional fluentd tweaks and yaml sdb

shaidar:
#### What's this PR do?
- Sets up a yaml sdb
- Updates the auth_log_filter syntax
- Adds a tls_forward.jinja file to use with all fluentd configs
- Updates config for fluentd clients to use common jinja files
- Moved some repetitive configs to the init file

### DIFF
--- a/pillar/fluentd/consul.sls
+++ b/pillar/fluentd/consul.sls
@@ -15,6 +15,7 @@ fluentd:
             - format: syslog
             - tag: consul.server
         - {{ auth_log_source('syslog.auth', '/var/log/auth.log') }}
+        - {{ auth_log_filter('grep', 'ident', 'consul', 'consul.server', 'regexp') }}
         - {{ auth_log_filter('grep', 'ident', '/CRON/') }}
         - {{ record_tagging |yaml() }}
         - {{ tls_forward |yaml() }}
--- a/pillar/fluentd/consul.sls
+++ b/pillar/fluentd/consul.sls
@@ -15,7 +15,6 @@ fluentd:
             - format: syslog
             - tag: consul.server
         - {{ auth_log_source('syslog.auth', '/var/log/auth.log') }}
-        - {{ auth_log_filter('grep', 'ident', '/consul', 'consul.server') }}
         - {{ auth_log_filter('grep', 'ident', '/CRON/') }}
         - {{ record_tagging |yaml() }}
         - {{ tls_forward |yaml() }}
--- a/pillar/fluentd/consul.sls
+++ b/pillar/fluentd/consul.sls
@@ -1,9 +1,8 @@
 {% from "fluentd/record_tagging.jinja" import record_tagging with context %}
 {% from "fluentd/auth_log.jinja" import auth_log_source, auth_log_filter with context %}
+{% from "fluentd/tls_forward.jinja" import tls_forward with context %}
 
 fluentd:
-  plugins:
-    - fluent-plugin-secure-forward
   configs:
     - name: consul_server
       settings:
@@ -15,24 +14,8 @@ fluentd:
             - pos_file: /var/log/syslog.pos
             - format: syslog
             - tag: consul.server
-        - directive: filter
-          directive_arg: '**'
-          attrs:
-            - '@type': grep
-            - regexp1: ident consul
         - {{ auth_log_source('syslog.auth', '/var/log/auth.log') }}
-        - {{ auth_log_filter('grep', 'ident', 'CRON') }}
+        - {{ auth_log_filter('grep', 'ident', '/consul', 'consul.server') }}
+        - {{ auth_log_filter('grep', 'ident', '/CRON/') }}
         - {{ record_tagging |yaml() }}
-        - directive: match
-          directive_arg: '**'
-          attrs:
-            - '@type': secure_forward
-            - self_hostname: {{ salt.grains.get('ipv4')[0] }}
-            - secure: 'false'
-            - flush_interval: '10s'
-            - shared_key: __vault__::secret-operations/global/fluentd_shared_key>data>value
-            - nested_directives:
-              - directive: server
-                attrs:
-                  - host: operations-fluentd.query.consul
-                  - port: 5001
+        - {{ tls_forward |yaml() }}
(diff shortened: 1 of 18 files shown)


pillar/fluentd/init.sls
```diff
@@ -1,3 +1,10 @@
+{% set ENVIRONMENT = salt.grains.get('environment') %}
+{% set minion_id = salt.grains.get('id', '') %}
+{% set cert = salt.vault.cached_write('pki-intermediate-{}/issue/fluentd-client'.format(ENVIRONMENT), common_name='fluentd.{}.{}'.format(minion_id, ENVIRONMENT)) %}
```
blarghmatey: We should be passing the `cache_prefix` parameter with the minion_id for all of these cached_write calls

pillar/fluentd/init.sls
```diff
@@ -1,3 +1,10 @@
...
+{% set fluentd_cert_path = salt.sdb.yaml.get('fluentd:cert_path') %}
```
blarghmatey: We don't have to use `sdb.get` in jinja as Salt supports passing an SDB string directly in most of its files. So this would end up being:

```sls
cert:
  fluentd_cert:
    content: |
      {{ cert.data.certificate | indent(8) }}
    path: sdb://yaml/fluentd:cert_path
```

pillar/fluentd/consul.sls
```diff
@@ -15,24 +14,8 @@ fluentd:
...
             - format: syslog
             - tag: consul.server
-        - directive: filter
-          directive_arg: '**'
-          attrs:
-            - '@type': grep
-            - regexp1: ident consul
         - {{ auth_log_source('syslog.auth', '/var/log/auth.log') }}
-        - {{ auth_log_filter('grep', 'ident', 'CRON') }}
+        - {{ auth_log_filter('grep', 'ident', '/consul', 'consul.server') }}
```
blarghmatey: I don't think we want to use this block for consul because the directive that you're removing to replace this with was to set it to only pass through consul logs, whereas this will pass through everything _except_ consul logs.

pillar/fluentd/ocw_cms.sls
```diff
@@ -1,4 +1,6 @@
 {% from "fluentd/record_tagging.jinja" import record_tagging with context %}
+{% from "fluentd/tls_forward.jinja" import tls_forward with context %}
+{% from "fluentd/tls_forward.jinja" import tls_forward with context %}
```
blarghmatey: Double paste

pillar/fluentd/reddit.sls
```diff
@@ -1,13 +1,8 @@
 {% from "fluentd/record_tagging.jinja" import record_tagging with context %}
+{% from "fluentd/tls_forward.jinja" import tls_forward with context %}
 
 fluentd:
-  overrides:
-    pkgs:
-      - ruby2.3
-      - ruby2.3-dev
-      - build-essential
```
blarghmatey: Any reason you're removing the ruby package name overrides?

pillar/fluentd/tls_forward.jinja
```diff
@@ -0,0 +1,23 @@
+{% set tls_forward = {
+    'directive': 'match',
+    'directive_arg': '**',
+    'attrs': [
+        {'@type': 'forward'},
+        {'transport': 'tls'},
+        {'tls_client_cert_path': salt.sdb.yaml.get('fluentd:cert_path')},
+        {'tls_client_private_key_path': salt.sdb.yaml.get('fluentd:cert_key_path')},
```
blarghmatey: These calls wouldn't be `sdb.yaml.get` it would be `sdb.get('sdb://yaml/fluentd:cert_key_path')`, etc.

pillar/fluentd/init.sls
shaidar: Fixed call to sdb, but left the `cert` config as is, since it feels cleaner to make changes to a set variable as opposed to dig through the whole file (granted this is a small file) and make changes somewhere in there.

pillar/fluentd/reddit.sls
shaidar: This came across as inconsistent as some files have it and others don't. When looking at the `map.jinja` file in the fluentd formula, those are listed under the Debian pkgs.

pillar/fluentd/consul.sls
```diff
@@ -15,24 +14,8 @@ fluentd:
...
             - format: syslog
             - tag: consul.server
-        - directive: filter
-          directive_arg: '**'
-          attrs:
-            - '@type': grep
-            - regexp1: ident consul
         - {{ auth_log_source('syslog.auth', '/var/log/auth.log') }}
-        - {{ auth_log_filter('grep', 'ident', 'CRON') }}
+        - {{ auth_log_filter('grep', 'ident', 'consul', 'consul.server', 'regexp') }}
```
blarghmatey: I think we want to move this above the auth_log_source line so that we don't allow only Consul messages from the auth log as well.

pillar/fluentd/init.sls
```diff
@@ -1,3 +1,10 @@
+{% set ENVIRONMENT = salt.grains.get('environment') %}
+{% set minion_id = salt.grains.get('id', '') %}
+{% set cert = salt.vault.cached_write('pki-intermediate-{}/issue/fluentd-client'.format(ENVIRONMENT), common_name='fluentd.{}.{}'.format(minion_id, ENVIRONMENT), cache_prefix=minion_id) %}
+{% set fluentd_cert_path = sdb.get('sdb://yaml/fluentd:cert_path') %}
+{% set fluentd_cert_key_path = sdb.get('sdb://yaml/fluentd:cert_key_path') %}
+{% set ca_cert_path = sdb.get('sdb://yaml/fluentd:ca_cert_path') %}
```
blarghmatey: The function call should be `salt.sdb.get. I don't think `sdb` is injected directly into the Jinja context.

pillar/fluentd/tls_forward.jinja
```diff
@@ -0,0 +1,23 @@
+{% set tls_forward = {
+    'directive': 'match',
+    'directive_arg': '**',
+    'attrs': [
+        {'@type': 'forward'},
+        {'transport': 'tls'},
+        {'tls_client_cert_path': sdb.get('sdb://yaml/fluentd:cert_path')},
+        {'tls_client_private_key_path': sdb.get('sdb://yaml/fluentd:cert_key_path')},
+        {'tls_ca_cert_path': sdb.get('sdb://yaml/fluentd:ca_cert_path')},
```
blarghmatey: `salt.sdb.get`

pillar/fluentd/consul.sls
shaidar: But the filter is using the `consul.server` tag which should only apply to the /var/log/syslog file and not auth.log one

blarghmatey: Fair point, missed that.

blarghmatey:
:+1:

PR #1122, opened 2020-01-06
Decreased xpro sheets monitoring frequency for non-prod envs

gsidebo:
No ticket

#### What's this PR do?
See title

Now that we aren't in an active testing period for the sheets feature, we don't need to be polling our Drive folders for changes in certain Sheets


blarghmatey:
:+1:

PR #1124, opened 2020-01-07
Changed fluentd file from list to dict

shaidar:
#### What's this PR do?
In order to be able to specify common fluentd config files in `init.sls` and the specific files, we needed to change the file type from list to dict. This change is in conjunction with a change submitted to the fluentd formula.

Note: The one thing I'm not 100% sure about is the change to `server_operations_qa.sls` filename. I made it the same as the filename specified in `init.sls` as I'm expecting that the configs in the server config overwrite the contents of init, but could be wrong and it might throw a duplicate key error.

### DIFF
--- a/pillar/fluentd/cas.sls
+++ b/pillar/fluentd/cas.sls
@@ -4,7 +4,7 @@
 
 fluentd:
   configs:
-    - name: auth_server
+    auth_server:
       settings:
         - directive: source
           attrs:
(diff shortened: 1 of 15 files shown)


blarghmatey:
:+1: 
The `fluentd_log` block in the `server_operations_qa.sls` will merge with the values defined in the `init.sls` file. Because the contents of that key are a list, that means that the `init.sls` settings will be overridden in the server case as it is specified after `init.sls` in the pilar top file.

PR #1127, opened 2020-01-10
Fix OCW Mirror Nginx pillar rendering

markbreedlove:
Fix the Nginx pillar for OCW's mirror servers. Though there is a "qa" minion for the ocw-mirror role, its Nginx configuration is superficial, because the only reason for the "qa" mirror to exist is to verify that mirror publishing scripts work. The QA mirror isn't accessed over HTTP. Therefore, there only has to be one "production" Nginx configuration that actually works.

I'm keeping an nginx configuration on `ocw-qa-cms-2` because it's still a good idea for making sure that rendering works. It just doesn't need to have the correct domain name. Is this a bad idea?

Without the modification in this commit, there is a Jinja2 rendering error about a dict having no 'qa' attribute. That's because the ocw-environment grain on the QA minion is "qa" and there's no such property of `environments:ocw:purposes:ocw-mirror`.

### DIFF
--- a/pillar/nginx/kibana.sls
+++ b/pillar/nginx/kibana.sls
@@ -52,7 +52,7 @@ nginx:
               - ssl_client_certificate: /etc/ssl/certs/mitca.pem
               - ssl_verify_client: 'on'
               - set $authorized: 'no'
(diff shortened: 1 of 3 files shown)


blarghmatey:
:+1:

PR #1128, opened 2020-01-13
Changed default datadog ports on fluentd aggregators

shaidar:
#### What's this PR do?
By default, datadog-agent uses the same port as the one we're using for the fluentd aggregators. This PR changes the default ports for datadog-agent.


blarghmatey:
:+1:

PR #1129, opened 2020-01-13
Do not run Nginx role on OCW Mirror QA

markbreedlove:
Only apply Nginx states to the production instance of OCW's Mirror server.


blarghmatey:
:+1:

PR #1130, opened 2020-01-14
OCW chalkradio redirects

markbreedlove:
See https://odl.zendesk.com/agent/tickets/44331

I'm adding these to amps-redirect because this is already a place where we manage assorted redirects, and it seems likely that one of us would look there if we need to maintain them.

### DIFF
--- a/pillar/nginx/chalkradio_redirect.sls
+++ b/pillar/nginx/chalkradio_redirect.sls
@@ -1,0 +1,33 @@
+nginx:
+    servers:
+    managed:
+      chalkradio-redirect:
+        enabled: True
+        config:
+          - server:
+              - server_name: chalkradiopodcast.com chalkradiopodcast.org chalkradiopodcast.net
+              - listen: 80
+              - listen: '443 ssl'
+              - listen: '[::]:80'
+              - listen: '[::]:443 ssl'
+              # The certificate uses subject alternative names.
+              - ssl_certificate: /etc/letsencrypt/live/amps-web.amps.ms.mit.edu/cert.pem
+              - ssl_certificate_key: /etc/letsencrypt/live/amps-web.amps.ms.mit.edu/privkey.pem
+              - ssl_stapling: 'on'
+              - ssl_stapling_verify: 'on'
+              - ssl_session_timeout: 1d
+              - ssl_session_tickets: 'off'
+              - ssl_protocols: 'TLSv1.2 TLSv1.3'
+              - ssl_ciphers: >-
+                  TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:
+                  TLS_AES_128_GCM_SHA256:DHE-RSA-AES256-GCM-SHA384:
+                  DHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:
+                  ECDHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-SHA256:
+                  DHE-RSA-AES128-SHA256:ECDHE-RSA-AES256-SHA384:
+                  ECDHE-RSA-AES128-SHA256
+              - ssl_prefer_server_ciphers: 'on'
+              - resolver: 1.1.1.1
+              - location /.well-known/:
+                - alias: /usr/share/nginx/html/.well-known/
+              - location /:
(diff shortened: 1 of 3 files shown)


pillar/nginx/chalkradio_redirect.sls
```diff
@@ -0,0 +1,34 @@
...
+
+nginx:
+    servers:
+    managed:
+      {% for tld in tlds %}
+      chalkradio-{{ tld }}-redirect:
+        enabled: True
+        config:
+          - server:
+              - server_name: chalkradiopodcast.{{ tld }}
```
blarghmatey: Rather than defining a separate server block for each domain you can just have all of the domains in a space separated list in the `server_name` parameter.

pillar/nginx/chalkradio_redirect.sls
```diff
@@ -0,0 +1,34 @@
...
+              - listen: '[::]:443 ssl'
+              # The certificate uses subject alternative names.
+              - ssl_certificate: /etc/letsencrypt/live/amps-web.amps.ms.mit.edu/cert.pem
+              - ssl_certificate_key: /etc/letsencrypt/live/amps-web.amps.ms.mit.edu/privkey.pem
+              - ssl_stapling: 'on'
+              - ssl_stapling_verify: 'on'
+              - ssl_session_timeout: 1d
+              - ssl_session_tickets: 'off'
+              - ssl_protocols: 'TLSv1.2 TLSv1.3'
+              - ssl_ciphers: "TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256\
```
blarghmatey: If you want to split this across lines it's better to use the `>-` operator in YAML than the backslash escape.
For example:

```yaml
ssl_ciphers: >-
  TLS_AES_356_GCM_SHA384:
  TLS_CHACHA20_POLY1305_SHA256
```

pillar/nginx/chalkradio_redirect.sls
```diff
@@ -0,0 +1,34 @@
...
+                  :DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384\
+                  :ECDHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-SHA256:DHE-RSA-AES128-SHA256\
+                  :ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256"
+              - ssl_prefer_server_ciphers: 'on'
+              - resolver: 1.1.1.1
+              - location /.well-known/:
+                - alias: /usr/share/nginx/html/.well-known/
+              - location /:
+                - return: 301 https://ocw.mit.edu/educator/chalk-radio-podcast/
```
blarghmatey: Do we want all routes on these domains to redirect specifically to that path, or do we want the path to be appended to that base?

pdpinch: My understanding from the ZenDesk ticket is that they want all routes to redirect to this path. The use case is they announce the domain "chalktalkpodcast" on the podcast and the user gets to the this page on the OCW site.

pillar/nginx/chalkradio_redirect.sls
markbreedlove: I just copied that from the original amps-redirect config, so we'd want to change that, too.

pillar/nginx/chalkradio_redirect.sls
markbreedlove: Oh, didn't realize that. will change ...

pillar/nginx/chalkradio_redirect.sls
markbreedlove: I've changed the line wrapping in this configuration, but I'll do another PR outside of this one to fix all of the others in `pillar/nginx`.

pillar/nginx/chalkradio_redirect.sls
```diff
@@ -0,0 +1,34 @@
+nginx:
+    servers:
+    managed:
+      {% for tld in tlds %}
```
blarghmatey: This for loop can go now that we've got all of the domains in the server_name field

blarghmatey: The templated `tld` in the next line will also need to be removed

blarghmatey:
:+1:

PR #1131, opened 2020-01-14
Added tls_forward macro to accomodate env without consul

shaidar:
#### What's this PR do?
We aren't running a consul cluster in the OCW VPC and for fluentd clients to work they need to be able to resolve `operations-fluent.query.consul`. This PR gives us the ability to override the static consul setting in the `tls_forward.jinja` so that we can pass a resolvable in pillar for OCW fluentd configs.


blarghmatey:
:+1:

PR #1138, opened 2020-01-16
Make production logging Elasticsearch conf like QA

markbreedlove:
Change some settings in the Operations Elasticsearch (logging) cluster's configuration to bring it in-line with the one on QA, which we set up first, and which we know works as desired.


blarghmatey:
:+1:

Issue #1133, opened 2020-01-15
Add elastalert for expired SAML metadata


blarghmatey:
When the SAML metadata in the residential edX instances expires it causes authentication failures. To gain early warning of this situation we would like to set up an elastalert trigger on logs of the form "No SAMLProviderData found"

https://logs.odl.mit.edu/goto/8e6ef6a22d695e66c8e4af3ac6294776

Issue #1132, opened 2020-01-15
Add scheduled pull of SAML metadata for residential


blarghmatey:
In order to ensure that we have up to date SAML metadata on residential we need to add a scheduled job to run the `saml --pull` management command on LMS for both draft and live servers in QA and production.

PR #1139, opened 2020-01-17
Alert and task to refresh saml provider metadata

shaidar:
- [Issue#1133](https://github.com/mitodl/salt-ops/issues/1133)
- [Issue#1132](https://github.com/mitodl/salt-ops/issues/1132)

### DIFF
--- a/pillar/edx/init.sls
+++ b/pillar/edx/init.sls
@@ -47,3 +47,10 @@ schedule:
     args:
       - etl.mitx
   {% endif %}
+  {% if 'mitx-' in environment %}
+  refresh_saml_provider_metadata:
+    days: 5
+    function: state.sls
+    args:
+      - edx.refresh_saml_provider_metadata
+  {% endif %}
--- a/pillar/edx/init.sls
+++ b/pillar/edx/init.sls
@@ -4,6 +4,7 @@
 {% set environment = salt.grains.get('environment', 'mitx-qa') %}
 {% set purpose_data = env_settings.environments[environment].purposes[purpose] %}
 {% set roles = salt.grains.get('roles') %}
+{% set business_unit = salt.grains.get('business_unit') %}
 
 edx:
   config:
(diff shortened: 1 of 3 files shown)


pillar/edx/init.sls
```diff
@@ -47,3 +47,10 @@ schedule:
     args:
       - etl.mitx
   {% endif %}
+  {% if 'mitx-' in environment %}
```
blarghmatey: To be more explicit about this we can filter on `business_unit == residential`

pillar/elastalert.sls
```diff
@@ -124,6 +124,36 @@ elastic_stack:
...
+        settings:
+          name: Authentication with MIT Kerberos is currently unavailable
+          description: >-
+            Send a message anytime an error message containing No SAMLProviderData
+            found is encountered in the mitx production logs.
+          opsgenie_key: {{ opsgenie_key }}
+          opsgenie_priority: P1
+          opsgenie_alias: mitx_saml_error
+          type: frequency
+          index: logstash-*
```
blarghmatey: Should we segment this to logstash-mitx-production-*?

pillar/elastalert.sls
```diff
@@ -124,6 +124,36 @@ elastic_stack:
...
+          num_events: 5
```
blarghmatey: Should this be triggered on 1 event?

pillar/edx/init.sls
shaidar: Sure, but feels as if we're more consistent with the environment grain than the business_unit, but I'll set it as much

pillar/elastalert.sls
shaidar: We want the alert for QA too though, no?

blarghmatey: Well, we only want to get paged for production

blarghmatey: Also, the environment name is already being set to production at the bottom of this stanza

blarghmatey:
:+1:

PR #1141, opened 2020-01-22
Upgrade edX next-residential to Juniper Alpha release

markbreedlove:
Upgrade the next-residential edx app to the Juniper Alpha release.

### DIFF
--- a/salt/orchestrate/edx/build_ami.sls
+++ b/salt/orchestrate/edx/build_ami.sls
@@ -4,7 +4,7 @@
 {% set env_settings = env_dict.environments[ENVIRONMENT] %}
 {% set VPC_NAME = salt.environ.get('VPC_NAME', env_settings.vpc_name) %}
 {% set BUSINESS_UNIT = salt.environ.get('BUSINESS_UNIT', env_settings.business_unit) %}
-{% set RUN_MIGRATIONS = salt.environ.get('RUN_MIGRATIONS') %}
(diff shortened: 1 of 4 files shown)


blarghmatey:
:+1:

markbreedlove:
@blarghmatey : Thanks for giving this the thumbs up earlier. I have since added one commit that adds that `RUN_MIGRATIONS` flag we were talking about.

salt/orchestrate/edx/build_ami.sls
```diff
@@ -4,6 +4,7 @@
 {% set env_settings = env_dict.environments[ENVIRONMENT] %}
 {% set VPC_NAME = salt.environ.get('VPC_NAME', env_settings.vpc_name) %}
 {% set BUSINESS_UNIT = salt.environ.get('BUSINESS_UNIT', env_settings.business_unit) %}
+{% set RUN_MIGRATIONS = salt.environ.get('RUN_MIGRATIONS') %}
```
blarghmatey: Should add a default value of `False` just to help clarify that it's a boolean.

blarghmatey: We should also remove the EDX_VERSION line above as it no longer has any purpose.

markbreedlove: OK, I've made those changes.

blarghmatey:
:+1:

PR #1142, opened 2020-01-23
Specify python binary for virtualenv

markbreedlove:
Does this look like a good way to get the`virtualenv.managed` state module to use Python 3.5?

I'm getting an error running it that states, `ERROR: Package 'setuptools' requires a different Python: 2.7.12 not in '>=3.5'`.


markbreedlove:
Closing this in favor of another approach, which is to add a next_residential pillar like this one and specify Python 3: https://github.com/mitodl/salt-ops/blob/master/salt/edx/prod.sls#L16-L28

PR #1143, opened 2020-01-23
Add OS package dependency data to next_residential pillar

markbreedlove:
Add `edx:dependencies:os_packages` to `pillar/edx/ansible_vars/next_residential.sls` to try to get the AMI build orchestration to create a virtualenv with python 3.5 instead of 2.7.

### DIFF
--- a/pillar/edx/ansible_vars/next_residential.sls
+++ b/pillar/edx/ansible_vars/next_residential.sls
@@ -42,8 +42,7 @@ edx:
       - libssl-dev
       - python3.5
       - python3.5-dev
-      - python-pip
-      - python-virtualenv
+      - python3-pip
+      - python3-virtualenv
       - nfs-common
       - postfix
-      - memcached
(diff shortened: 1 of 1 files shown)


pillar/edx/ansible_vars/next_residential.sls
```diff
@@ -32,3 +32,18 @@ edx:
...
+    os_packages:
+      - git
+      - libmysqlclient-dev
+      - mariadb-client-10.0
+      - landscape-common
+      - libssl-dev
+      - python3.5
+      - python3.5-dev
+      - python-pip
+      - python-virtualenv
```
blarghmatey: These should be `python3-pip` and `python3-virtualenv`

pillar/edx/ansible_vars/next_residential.sls
```diff
@@ -32,3 +32,18 @@ edx:
...
+      - nfs-common
+      - postfix
+      - memcached
```
blarghmatey: We can actually get rid of memcached now as we are using elasticache for handling a shared cache across isntances.

blarghmatey:
:+1:

PR #1145, opened 2020-01-23
Added ELASTICSEARCH_INDEXING_CHUNK_SIZE key and set nginx value

shaidar:
#### What's this PR do?
- Added `ELASTICSEARCH_INDEXING_CHUNK_SIZE` as it was missing from Heroku config in addition to a comment that the value for it should match what's set in nginx
- Set the nginx `client_max_body_size` to the same value as the chunk_size


blarghmatey:
:+1:

PR #1146, opened 2020-01-24
Adding state and pillar to handle installing requirements to xqwatcher

blarghmatey:
There are two different sets of requirements for graders to work. One set that is used in the student submissions and one that is used by the grader code. Each of those are needed in different virtualenvs to get everything working. This PR adds a state to install packages into the xqueue-watcher virtualenv for the graders and the pillar values that we need for each of the courses.

Trello ticket 

#### What's this PR do?
Handles installing grader dependencies in xqueue-watcher virtualenv


shaidar:
👍

PR #1147, opened 2020-01-27
Reinstate Ansible interpreter, deps for edX next_residential

markbreedlove:
In the next_residential pillar for the edX application, reinstate the ansible_python_interpreter variable and the `virtualenv` Python dependency. We want the Python interpreter to be 3.5 again and we want `virtualenv` installed from pip.


blarghmatey:
:+1:

PR #1149, opened 2020-01-27
Replaced monit with datadog http-check

shaidar:
#### What's this PR do?
Leveraged Datadog http_checks to replace monit and added checks for the following production sites:
- residential
- latex2edx
- xpro
- discussions-reddit
- odl-video


blarghmatey:
:+1:

PR #1156, opened 2020-02-10
Add next-residential Ansible var for Nginx SSL ciphers

markbreedlove:
Add the Ansible variable `NGINX_SSL_CIPHERS` for next-residential, which will change `ssl_ciphers` in the Nginx configuration. The list of ciphers is being copied from what's been seen to work in production. Without this change, the SSL connection between the AWS loadbalancer and Nginx on the application server was failing.


markbreedlove:
Here's the relevant variable in the`configuration` project: https://github.com/mitodl/configuration/blob/ac2aeebe125622aa5e19913470717f7a8050788e/playbooks/roles/nginx/defaults/main.yml#L54

blarghmatey:
:+1:

PR #1157, opened 2020-02-10
Add JWT signing variables for next-residential

markbreedlove:
This is basically what's been done with the JWT variables in `xpro.sls`.


blarghmatey:
:+1:

PR #1160, opened 2020-02-12
Additions to deploy tika in rc-apps

shaidar:
#### What's this PR do?
- Adds Tika conf file
- Adds Tika config to rc-apps env
- Makes some changes to `deploy.sls` so that it can be used to also create route53 dns records in cases when a domain name(s) is/are specified in environment settings.

Note: I didn't add a sec_group as those instances should only be available to within the VPC to open instances.

### DIFF
--- a/salt/orchestrate/apps/deploy.sls
+++ b/salt/orchestrate/apps/deploy.sls
@@ -13,7 +13,7 @@
 {% do security_groups.extend(['master-ssh', 'consul-agent']) %}
 {% set release_id = (salt.sdb.get('sdb://consul/' ~ app_name ~ '/' ~ ENVIRONMENT ~ '/release-id') or 'v1') %}
 {% set target_string = app_name ~ '-' ~ ENVIRONMENT ~ '-*-' ~ release_id %}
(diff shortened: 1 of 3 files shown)


salt/orchestrate/apps/deploy.sls
```diff
@@ -89,3 +90,35 @@ build_{{ app_name }}_nodes:
...
+    tgt='G@roles:' ~ app_name ~ ' and G@environment:' ~ ENVIRONMENT, fun='grains.item', tgt_type='compound'
+    ).items() %}
+{% do hosts.append(grains['external_ip']) %}
+{% endfor %}
+
+{% set zone_list = salt.boto_route53.list_all_zones_by_name %}
+{% for server_domain_name in server_domain_names %}
+{% for zone in zone_list %}
+{% if zone in server_domain_name %}
+register_{{ server_domain_names }}_nodes_with_dns:
```
blarghmatey: Remove the plural from this so that it doesn't generate conflicting state ids

blarghmatey:
:+1:

PR #1161, opened 2020-02-12
Added nginx as proxy to tika

shaidar:
#### What's this PR do?
Since Tika instances will need to be accessible from Heroku and because it's not ideal to have them  publicly accessible by anyone, we ended up having to add Nginx to those instances to act as a proxy and check for an API token when calling the Tika service.
This PR includes the Nginx config, the needed changes to the top files and the changes to the environment settings file whereby I added a sec group and removed the domain name since we'll be using an AWS ALB to front those instances.

Note: Nginx pillar config was mostly copied from the reddit one with some mods.


pillar/nginx/tika.sls
```diff
@@ -0,0 +1,45 @@
...
+  servers:
+    managed:
+      default:
+        enabled: False
+        deleted: True
+        config: None
+      tika:
+        enabled: True
+        config:
+          - map $http_upgrade $connection_upgrade:
```
blarghmatey: I think this is only necessary for websockets which aren't relevant for Tika.

blarghmatey:
:+1:

PR #1165, opened 2020-02-18
Removing master key before snapshotting AMI

blarghmatey:
When moving from the QA to the production master the master key mismatches, causing production builds to fail. This adds a step to remove the master key to avoid that error case.


shaidar:
👍

PR #1167, opened 2020-02-20
Add netdata-formula & add to logging cluster

markbreedlove:
Add recently-created `netdata-formula` and associate it with the Elasticsearch logging cluster, so that we can use that as a test case in QA.

### DIFF
--- a/bootstrap/etc/salt/master
+++ b/bootstrap/etc/salt/master
@@ -24,6 +24,7 @@ gitfs_remotes:
   - https://github.com/mitodl/consul-formula
   - https://github.com/mitodl/vault-formula
   - https://github.com/mitodl/fluentd-formula
+  - https://github.com/mitodl/netdata-formula
   - https://github.com/mitodl/datadog-formula
 
 root_dir: bootstrap/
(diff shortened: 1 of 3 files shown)


bootstrap/etc/salt/master
```diff
@@ -24,6 +24,7 @@ gitfs_remotes:
   - https://github.com/mitodl/consul-formula
   - https://github.com/mitodl/vault-formula
   - https://github.com/mitodl/fluentd-formula
+  - https://github.com/mitodl/netdata-formula
```
blarghmatey: This isn't needed in the bootstrap. That's just for deploying new master servers, so it should be as barebones as we can make it.

salt/top.sls
```diff
@@ -49,6 +49,7 @@ base:
     - elastic-stack.elasticsearch
     - elastic-stack.elasticsearch.plugins
     - datadog.plugins
+    - netdata
```
blarghmatey: For targeting, we should just go ahead and add it to all of operations-qa

blarghmatey:
:+1:

PR #1168, opened 2020-02-25
Change xpro support email address

shaidar:
[Issue#1288](https://github.com/mitodl/mitxpro/issues/1288)

#### What's this PR do?
- Changes to support email address.
- Change Heroku app to use maligun smtp config.

* Note: I set the vault path for smtp_config that way to match what's specified [here](https://github.com/mitodl/salt-ops/blob/master/pillar/edx/ansible_vars/cloud_deployment.sls#L13) in order to avoid duplication

### DIFF
--- a/pillar/heroku/xpro.sls
+++ b/pillar/heroku/xpro.sls
@@ -148,11 +148,11 @@ heroku:
     MITXPRO_BASE_URL: {{ env_data.MITXPRO_BASE_URL }}
     MITXPRO_DB_CONN_MAX_AGE: 0
     MITXPRO_DB_DISABLE_SSL: True    # pgbouncer buildpack uses stunnel to handle encryption
-    MITXPRO_EMAIL_HOST: {{ smtp_config.relay_host }}
-    MITXPRO_EMAIL_PASSWORD: {{ smtp_config.relay_password }}
-    MITXPRO_EMAIL_PORT: {{ smtp_config.relay_port }}
+    MITXPRO_EMAIL_HOST: {{ smtp_config.data.relay_host }}
+    MITXPRO_EMAIL_PASSWORD: {{ smtp_config.data.relay_password }}
+    MITXPRO_EMAIL_PORT: {{ smtp_config.data.relay_port }}
     MITXPRO_EMAIL_TLS: True
-    MITXPRO_EMAIL_USER: {{ smtp_config.relay_username }}
+    MITXPRO_EMAIL_USER: {{ smtp_config.data.relay_username }}
     MITXPRO_ENVIRONMENT: {{ env_data.env_name }}
     MITXPRO_FROM_EMAIL: 'MIT xPRO <support@xpro.mit.edu>'
     MITXPRO_LOG_LEVEL: {{ env_data.app_log_level }}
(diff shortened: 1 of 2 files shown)


pdpinch:
Does the SMTP server address go into vault?

shaidar:
Yes that lives in vault and is specified [here](https://github.com/mitodl/salt-ops/blob/master/pillar/edx/ansible_vars/cloud_deployment.sls#L13)

shaidar:
@pdpinch I noticed that the xpro maligun app config wasn't using the mailgun smtp even though it was checked off in the Trello card. I made the changes, however I don't believe the address (especially for `MITXPRO_FROM_EMAIL` and `MITXPRO_REPLY_TO_ADDRESS`) is what you had in mind.

pillar/heroku/xpro.sls
```diff
@@ -147,20 +148,20 @@ heroku:
...
-    MITXPRO_EMAIL_PORT: 587
+    MITXPRO_EMAIL_HOST: {{ smtp_config.relay_host }}
+    MITXPRO_EMAIL_PASSWORD: {{ smtp_config.relay_password }}
+    MITXPRO_EMAIL_PORT: {{ smtp_config.relay_port }}
     MITXPRO_EMAIL_TLS: True
-    MITXPRO_EMAIL_USER: __vault__::secret-operations/global/mit-smtp>data>relay_username
+    MITXPRO_EMAIL_USER: {{ smtp_config.relay_username }}
     MITXPRO_ENVIRONMENT: {{ env_data.env_name }}
-    MITXPRO_FROM_EMAIL: 'MIT xPRO <xpro@mit.edu>'
+    MITXPRO_FROM_EMAIL: 'MIT xPRO <xpro@xpro.mit.edu>'
```
pdpinch: Ugh, I'm going to have to check on this and get back to you. Do you think this is the "from" address that's used whenever a course team sends an email?

pillar/heroku/xpro.sls
```diff
@@ -147,20 +148,20 @@ heroku:
...
     MITXPRO_LOG_LEVEL: {{ env_data.app_log_level }}
     MITXPRO_OAUTH_PROVIDER: 'mitxpro-oauth2'
     MITXPRO_REGISTRATION_ACCESS_TOKEN:  __vault__:gen_if_missing:secret-{{ business_unit }}/{{ env_data.openedx_environment }}/xpro-registration-access-token>data>value
-    MITXPRO_REPLY_TO_ADDRESS: 'MIT xPRO <xpro@mit.edu>'
+    MITXPRO_REPLY_TO_ADDRESS: 'MIT xPRO <xpro@xpro.mit.edu>'
```
pdpinch: I'll have to check on this as well. I assume this will route to zendesk, but I'm checking.

pillar/heroku/xpro.sls
shaidar: Yeah I'm pretty sure it is - https://github.com/mitodl/mitxpro/blob/master/app.json#L318

pillar/heroku/xpro.sls
pdpinch: I don't think I'm going to get answer today, and changing this again seems like it would be relatively minor. 

Let's go with: 

```suggestion
    MITXPRO_REPLY_TO_ADDRESS: 'MIT xPRO <support@xpro.mit.edu>'
```

pillar/heroku/xpro.sls
pdpinch: I don't think I'm going to get an answer today, and changing this again seems like it would be relatively minor. 

Let's go with this, so we can move ahead. 

```suggestion
    MITXPRO_FROM_EMAIL: 'MIT xPRO <support@xpro.mit.edu>'
```

shaidar:
@pdpinch I made the changes manually on the xpro-ci app if u'd like to test it out.

pillar/heroku/xpro.sls
```diff
@@ -97,6 +97,7 @@
 {% set env_data = env_dict[environment] %}
 {% set business_unit = 'mitxpro' %}
 {% set cybersource_creds = salt.vault.read('secret-' ~ business_unit ~ '/' ~ env_data.vault_env_path ~ '/cybersource') %}
+{% set smtp_config = salt.vault.read('secret-' ~ business_unit ~ '/' ~ business_unit ~ env_data.env_name ~ '/smtp_config') %}
```
blarghmatey: Needs `.data` at the end of the call

blarghmatey:
:+1:

PR #1170, opened 2020-02-27
Retry conditional system_site_packages for virtualenv

markbreedlove:
Retry what was attempted in commit
2aa0515126c7a8a1cefb1f5cc2aec051262f180e, but with a default value that
allows the template to be rendered if the `edx_codename' grain is
undefined.


blarghmatey:
:+1:

PR #1171, opened 2020-02-28
Add netdata ports to "master-ssh" security groups

markbreedlove:
Add rules to allow access on Netdata's port, to the existing "master-ssh" security groups. This allows us to access Netdata via SSH tunnels.

### DIFF
--- a/salt/orchestrate/aws/security_groups.sls
+++ b/salt/orchestrate/aws/security_groups.sls
@@ -27,6 +27,13 @@ create_master_ssh_security_group:
           cidr_ip:
             - 10.0.0.0/22
             - 10.1.0.0/22
+        # netdata:
+        - ip_prodocol: tcp
+          from_port: 19999
+          to_port: 19999
+          cidr_ip:
+            - 10.0.0.0/22
+            - 10.0.0.0/22
     - tags:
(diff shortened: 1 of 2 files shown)


blarghmatey:
I think we also need to add that to `orchestrate/aws/mitx.sls`. That's a separate file that we use for managing settings in the edX environments.

markbreedlove:
Yeah, just noticed that ...

salt/orchestrate/aws/security_groups.sls
```diff
@@ -27,6 +27,13 @@ create_master_ssh_security_group:
           cidr_ip:
             - 10.0.0.0/22
             - 10.1.0.0/22
+        # netdata:
+        - ip_prodocol: tcp
+          from_port: 19999
+          to_port: 19999
+          cidr_ip:
+            - 10.0.0.0/22
+            - 10.0.0.0/22
```
blarghmatey: This has the same CIDR listed twice

markbreedlove: oops. thanks, just fixed it.

blarghmatey:
:+1:

PR #1176, opened 2020-03-04
Moved VOUCHER_INTERNATIONAL_* keys to prod env

shaidar:
[Issue#1577 ](https://github.com/mitodl/mitxpro/issues/1577)

#### What's this PR do?
Moved `VOUCHER_INTERNATIONAL_*` keys to production env so that they only apply to that env while devs verify on CI/RC that those keys are no longer needed and then can be removed all together.


blarghmatey:
:+1:

PR #1179, opened 2020-03-10
Add Datadog HTTP check for OCW CMS

markbreedlove:
This adds a Datadog HTTP Check for the OCW CMS.  Is this adequate? These appear to run on the Kibana server, so I added a rule to allow it in the security group that the CMS belongs to.


shaidar:
👍

PR #1180, opened 2020-03-12
Add a Kibana instance

markbreedlove:
Add a Kibana instance, which will give us more than one Elastalert
agent, for better alerting reliability.

It will also give us a hot standby for Kibana.


markbreedlove:
I think I need to make sure it's not going to duplicate alerts with two nodes querying the same logs.

markbreedlove:
I'm changing my approach to the overall issue of Elastalert reliability. I'll open another PR for this branch.

PR #1182, opened 2020-03-13
Fix fluentd regex and monitor edX unregistered task errors

markbreedlove:
This PR does two things:
* It improves a regex in our FluentD configuration so that it parses a new timestamp format that's mixed in with the other formats we've already parsed.
* It adds an Elastalert alert for "Received unregistered task" messages that result from the code on the worker and app server being out of sync.

See https://trello.com/c/qZBWIpSf/57-add-an-elastalert-for-unregistered-tasks-in-edx

I've checked the regular expressions with rubular.com against the following sample lines:

(Already-handled)
```
2019-10-23 11:58:32,696 WARNING 7812 [enterprise.utils] utils.py:50 - Could not import Registry from third_party_auth.provider
```

(New line that is currently generating FluentD error messages)
```
[2020-03-11 13:55:03,108: ERROR/MainProcess] Received unregistered task of type 
```

### DIFF
--- a/pillar/elastalert.sls
+++ b/pillar/elastalert.sls
@@ -445,3 +445,23 @@ elastic_stack:
                 filter:
                   - term:
                       fluentd_tag: edx.cms
+      - name: edx_unregistered_task
+        settings:
+          name: edX task failing
+          description: >-
+            A task has failed due to a mismatch in the source code between the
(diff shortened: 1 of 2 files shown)


blarghmatey:
:+1:

PR #1183, opened 2020-03-13
Add alert for OCW media asset problem

markbreedlove:
Add an Elastalert alert for a problem in the OCW CMS where a misnamed media asset causes an endpoint to error out that is queried by the mirror engine. The alert will go to the ocw-eng slack channel.

https://trello.com/c/aIt6TfFz/35-fix-ocw-mirror-engine


blarghmatey:
:+1:

PR #1184, opened 2020-03-16
Change mailgun domain for OVS

shaidar:
#### What's this PR do?
As part of the work of moving the OVS app behind an ALB, we ran into an issue whereby we couldn't change the A record of `video.odl.mit.edu` that's pointing to an EC2 instance to a CNAME since that name has MX and TXT records pointing to mailgun. Created a new mailgun domain and added those records to Route 53 so that it doesn't conflict with `video.odl.mit.edu`. 
This minor change points to the newly created maligun domain.


blarghmatey:
:+1:

PR #1185, opened 2020-03-16
Added check for starcellbio

shaidar:
#### What's this PR do?
Adds a Datadog http check for the starcellbio app.


blarghmatey:
:+1:

PR #1187, opened 2020-03-17
Added moira app cert and key

shaidar:
[Issue#2681](https://github.com/mitodl/open-discussions/issues/2681)

#### What's this PR do?
Adds the Moira app cert that has already been obtained from IS&T along with another key-value `FEATURE_MOIRA` to CI/RC


blarghmatey:
:+1:

PR #1189, opened 2020-03-20
Exclude netdata folder from inotify

shaidar:
#### What's this PR do?
Exclude netdata folder from inotify as it's generating useless posts to slack


blarghmatey:
:+1:

PR #1190, opened 2020-03-20
Condition to skip worker service restart on base

shaidar:
#### What's this PR do?
When running the `build_ami` state, `orchestrate.edx.build_ami.build_edx_base_nodes` fails when restarting the supervisord services on the worker instances as I believe it's timing out waiting for a response. The change here isn't a fix but a patch to sidestep the restart while building the ami.


blarghmatey:
:+1:

PR #1191, opened 2020-03-20
Read-only access to ES rc-apps with token

shaidar:
### What's this PR do?
Provides read-only access to ES rc-apps with token generated and stored in vault.


blarghmatey:
:+1:

PR #1192, opened 2020-03-23
Changed OVS url for datadog healthcheck

shaidar:
#### What's this PR do?
Noticed a lot of false alerts being generated by the OVS Datadog http health check. I made the change to match what we currently have setup as a health check endpoint on the ALB. Think that's more accurate and should cut stop the incoming false alerts.


blarghmatey:
:+1:

PR #1193, opened 2020-03-23
Add uWSGI defaults for all configurations

markbreedlove:
We have abandoned recent changes to update `uwsgi-formula` and have that formula set defaults. This commit sets recommended values for all configurations.

See https://github.com/mitodl/uwsgi-formula/pull/4


blarghmatey:
:+1:

PR #1194, opened 2020-03-23
Add Elastalert: edX S3ResponseError

markbreedlove:
Add an Elastalert alert for when course content can not be exported to Git due to an S3 403 Forbidden error.

### DIFF
--- a/pillar/elastalert.sls
+++ b/pillar/elastalert.sls
@@ -500,9 +500,9 @@ elastic_stack:
                   - query_string:
                       default_field: message
                       query: S3ResponseError
-                filter:
-                  - term:
-                      fluentd_tag: edx.*
+                  - query_string:
+                      default_field: fluentd_tag
+                      query: edx.cms.* OR edx.lms.*
       - name: edx_unregistered_task
         settings:
           name: edX task failing
--- a/pillar/elastalert.sls
+++ b/pillar/elastalert.sls
@@ -475,6 +475,34 @@ elastic_stack:
                 filter:
                   - term:
                       fluentd_tag: edx.cms
+      - name: edx_s3_response_error
+        settings:
+          name: edX S3 Response Error
+          description: >-
+            An edX worker got an error from S3 while trying to export course
+            content to Git. This usually means that the uwsgi service needs to
+            be restarted so that Celery can pick up updated credentials.
+          index: logstash-mitx*-production*
+          type: frequency
+          num_events: 1
+          timeframe:
+            minutes: 5
+          alert:
+            - slack
+          alert_text: "Automated git export failure"
+          slack_webhook_url: {{ slack_webhook_url }}
+          slack_channel_override: '#mitx-tech-notifs'
+          slack_username_override: Elastalert
+          slack_msg_color: "warning"
+          filter:
+            - bool:
+                must:
+                  - query_string:
+                      default_field: message
+                      query: S3ResponseError AND async_export_to_git
+                filter:
+                  - term:
+                      fluentd_tag: edx.cms.stderr
       - name: edx_unregistered_task
         settings:
           name: edX task failing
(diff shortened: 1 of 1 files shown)


pillar/elastalert.sls
```diff
@@ -475,6 +475,34 @@ elastic_stack:
                 filter:
                   - term:
                       fluentd_tag: edx.cms
+      - name: edx_s3_response_error
+        settings:
+          name: edX S3 Response Error
+          description: >-
+            An edX worker got an error from S3 while trying to export course
+            content to Git. This usually means that the uwsgi service needs to
```
blarghmatey: For edX the celery workers are managed with supervisor, not uwsgi. Also, this isn't exclusive to the workers, so a more accurate representation is that the S3 credentials have expired, so the process might need to be restarted or the credentials might need to be refreshed.

pillar/elastalert.sls
```diff
@@ -475,6 +475,34 @@ elastic_stack:
...
+          slack_webhook_url: {{ slack_webhook_url }}
+          slack_channel_override: '#mitx-tech-notifs'
+          slack_username_override: Elastalert
+          slack_msg_color: "warning"
+          filter:
+            - bool:
+                must:
+                  - query_string:
+                      default_field: message
+                      query: S3ResponseError AND async_export_to_git
```
blarghmatey: The query should just be for the s3response error, as it may happen outside of the git export and we would still like to know about those events also.

pillar/elastalert.sls
```diff
@@ -475,6 +475,34 @@ elastic_stack:
...
+                filter:
+                  - term:
+                      fluentd_tag: edx.cms.stderr
```
blarghmatey: Given that we want any s3response error, we may want to just filter on edx.* or edx.cms.* and edx.lms.*

pillar/elastalert.sls
markbreedlove: OK, changed.

pillar/elastalert.sls
markbreedlove: OK

pillar/elastalert.sls
markbreedlove: OK

pillar/elastalert.sls
```diff
@@ -499,10 +499,10 @@ elastic_stack:
                 must:
                   - query_string:
                       default_field: message
-                      query: S3ResponseError AND async_export_to_git
+                      query: S3ResponseError
                 filter:
                   - term:
-                      fluentd_tag: edx.cms.stderr
+                      fluentd_tag: edx.*
```
blarghmatey: Does the wildcard work in a term block? I don't know the specifics of the query DSL so it may be fine but it's worth validating.

markbreedlove: Thanks for noticing that. The wildcard does _not_ work in this term filter. I wonder if a filter on the fluentd tag is really necessary in the first place.

markbreedlove: I'm doing some queries and observing that the filter does help. Confining the result to particular fluentd tags doesn't have to be done with a `filter` clause, so I'm looking into an alternate way.

markbreedlove: Commit f6ea9218 should do the trick. I've tested the query, as follows (where localhost is tunneled to the ES cluster):
```
curl --location --request GET 'localhost:9200/logstash-mitx*-production*/_search' \
--header 'Content-Type: application/json' \
--data-raw '{
	"query": {
		"bool": {
			"must": [
				{"query_string": {"query": "S3ResponseError", "default_field": "message"}},
				{"query_string": {"query": "edx.cms.* OR edx.lms.*", "default_field": "fluentd_tag"}}
			]
		}
	}
}'
```

blarghmatey:
:+1:

PR #1198, opened 2020-03-30
Sync all modules before installing consul

shaidar:
#### What's this PR do?
`build_ami` state run fails the first time trying to install consul because it hadn't synced our modules. So we then have to sync those and run the build again and if we're lucky, it works. This change syncs the modules prior to trying to install consul.


blarghmatey:
:+1:

PR #1201, opened 2020-04-02
Update security group names in Kibana orchestration

markbreedlove:
I hope that this will fix a problem I'm seeing running the Kibana orchestration.

It's failing silently on `load_pillar_data_on_kibana_nodes`.

The rendered cloud map file is missing security groups that the old Kibana server belongs to:

```
         SecurityGroupId:


            - sg-03b82776





            - sg-5e6bbc2f


 
```

### DIFF
--- a/salt/orchestrate/operations/services/kibana.sls
+++ b/salt/orchestrate/operations/services/kibana.sls
@@ -30,9 +30,9 @@ generate_{{ app_name }}_cloud_map_file:
         release_id: {{ salt.sdb.get('sdb://consul/{{ app_name }}/{{ ENVIRONMENT }}/release-id')|default('v1') }}
         securitygroupid:
           - {{ salt.boto_secgroup.get_group_id(
-            'webapp-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
+            'webapp-odl-vpn-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
           - {{ salt.boto_secgroup.get_group_id(
-            'master-ssh-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
+            'salt_master-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
           - {{ salt.boto_secgroup.get_group_id(
             'consul-agent-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
         subnetids: {{ subnet_ids|tojson }}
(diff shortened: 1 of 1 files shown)


salt/orchestrate/operations/services/kibana.sls
```diff
@@ -30,9 +30,9 @@ generate_{{ app_name }}_cloud_map_file:
         release_id: {{ salt.sdb.get('sdb://consul/{{ app_name }}/{{ ENVIRONMENT }}/release-id')|default('v1') }}
         securitygroupid:
           - {{ salt.boto_secgroup.get_group_id(
-            'webapp-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
+            'webapp-odl-vpn-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
           - {{ salt.boto_secgroup.get_group_id(
-            'master-ssh-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
+            'salt_master-{}'.format(ENVIRONMENT), vpc_name=VPC_NAME) }}
```
blarghmatey, line 35: We don't necessarily want this to be applied as such in QA. Another option is to create the master-ssh security group in the operations VPC

markbreedlove: The pre-existing Kibana instance is a member of the `consul-agent-operations`, `salt_master-operations`,  and `webapp-odl-vpn-operations` security groups. This has a `format()` that will apply `salt_master-operations` to the new instance, as well.

Should we not be using this security group? All of the other production operations instances seem to be a member. I was having trouble with the orchestration because of what I thought looked like a network timeout.

markbreedlove: Scratch that, I see how there's no `salt_master-operations-qa` -- there's `salt-master-operations-qa`. There's  This is inconsistent, so shouldn't we just fix the name of the security group?

blarghmatey: It's inconsistent because the operations VPC predates the others. We recently updated the convention for the security group to be `master-ssh-<environment>` but did not create it in the operations VPC yet. The salt-master-operations-qa security group only applies to the salt master instance and serves a different purpose, which is why we decided to change the name for the SSH security group to be more explicit in its purpose. The quickest fix is to just create that master-ssh-oeprations security group in the operations VPC, even manually is fine.

markbreedlove: OK! I have created the `master-ssh-operations` security group and have reversed the change to that line in `kibana.sls`.

blarghmatey:
:+1:

PR #1204, opened 2020-04-06
Make newer OpenSSL accept 1024-bit MIT CA key

markbreedlove:
For verifying client certificates, allow verification to proceed with MIT's 1024-bit CA key. Debian 10 uses a newer version of Openssl than Debian 9, and requires a 2048-bit CA key, by default, when it verifies client certificates. Applications like Kibana, in its Nginx configuration, verify client certificates, which is delegated to OpenSSL.

Without this change, we get the error `CA certificate key too weak` when we have info-level logging turned on in Nginx.


blarghmatey:
:+1:

PR #1208, opened 2020-04-09
Added patch of upload size to run_ansible state

blarghmatey:
We have run into situations where updating configuration values reverts the Nginx upload size setting. This adds that state into the run_ansible script to avoid that occurrence.

### DIFF
--- a/salt/edx/run_ansible.sls
+++ b/salt/edx/run_ansible.sls
@@ -97,6 +97,20 @@ run_ansible:
     - env:
         HOME: /root
 
+update_max_upload_size_for_lms:
+  file.replace:
+    - name: /etc/nginx/sites-enabled/lms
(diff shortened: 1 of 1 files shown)


salt/edx/run_ansible.sls
```diff
@@ -97,6 +97,20 @@ run_ansible:
     - env:
         HOME: /root
 
+update_max_upload_size_for_lms:
+  file.replace:
+    - name: /etc/nginx/sites-enabled/lms
+    - pattern: 'client_max_body_size\s+\d+M;'
+    - repl: 'client_max_body_size {{ salt.pillar.get("edx:edxapp:max_upload_size", "20") }}M;'
```
shaidar, line 104: Should we make the default 50 since that's we now have in on the LMS side?

PR #1214, opened 2020-04-16
Added env to app cert vault path

shaidar:
#### What's this PR do?
Obtained separate app cert for CI/RC and production and added vault path to point to those. Additionally added `business_unit` key similar to other salt states and replaced it where necessary. Cert and key have already been added where necessary in vault clusters.


blarghmatey:
:+1:

PR #1215, opened 2020-04-22
Add csrf cookie in residential nginx config

shaidar:
#### What's this PR do?
Adds a config to Nginx to set an expired csrf token with a domain set to `.mit.edu`. This is meant to overwrite some students having that csrf token set and causing them issues by logging them out of residential as that's the token used instead of the one set by the more specific one set by the residential platform.

### DIFF
--- a/salt/edx/run_ansible.sls
+++ b/salt/edx/run_ansible.sls
@@ -97,7 +97,7 @@ run_ansible:
     - env:
         HOME: /root
 
-update_max_upload_size_for_lms:
+update_max_upload_size_and_set_csrf_for_lms:
   file.replace:
     - name: /etc/nginx/sites-enabled/lms
     - pattern: 'client_max_body_size\s+\d+M;'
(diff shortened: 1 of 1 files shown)


salt/edx/run_ansible.sls
```diff
@@ -97,19 +97,24 @@ run_ansible:
...
-update_max_upload_size_for_lms:
+update_max_upload_size_and_set_csrf_for_lms:
   file.replace:
     - name: /etc/nginx/sites-enabled/lms
     - pattern: 'client_max_body_size\s+\d+M;'
     - repl: 'client_max_body_size {{ salt.pillar.get("edx:edxapp:max_upload_size", "20") }}M;'
     - backup: False
     - require:
         - cmd: run_ansible
+  file.line:
```
blarghmatey, line 108: This won't work as it's 2 different `file` module statements with the same ID. It will need to be split into a separate call.

blarghmatey:
:+1:

PR #1218, opened 2020-04-23
Fix for residential csrftoken issue

shaidar:
#### What's this PR do?
Fixes the logout issues reported in residential due to a conflicting csrf token. This has now been tested on Chrome, Firefox, and Safari. It appears that Safari disregards the `Expires` key however responds to `max-age`. Also added a comment explaining why we've added this Nginx directive.

### DIFF
--- a/salt/edx/run_ansible.sls
+++ b/salt/edx/run_ansible.sls
@@ -106,12 +106,16 @@ update_max_upload_for_lms:
     - require:
         - cmd: run_ansible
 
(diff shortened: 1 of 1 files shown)


blarghmatey:
This looks good, I would just want to double check functionality with IE 11

shaidar:
This has now been tested in IE 11 and it seems to be working as well.

blarghmatey:
:+1:

PR #1219, opened 2020-04-29
Change ODL video log location for FluentD

markbreedlove:

https://trello.com/c/qEbo79DI/78-add-odl-video-service-uwsgi-log-to-fluentd

#### What's this PR do?

See https://github.com/mitodl/odl-video-service/pull/883

It changes the location of the log that's read by FluentD.

It ensures that the log file is writable by the `deploy` account, under which the Django app workers run.

I started to parameterize the logfile location, but I balked because I wasn't sure if that was necessary, and also because I recalled that it is hard to include pillar data from one pillar file in another pillar file.  Can you suggest any way to have the file path coded in only one place?

### DIFF
--- a/pillar/apps/odlvideo.sls
+++ b/pillar/apps/odlvideo.sls
@@ -120,7 +120,7 @@ django:
     ODL_VIDEO_FROM_EMAIL: MIT ODL Video <odl-video-support@mit.edu>
     ODL_VIDEO_LOG_LEVEL: {{ env_data.log_level }}
     ODL_VIDEO_SUPPORT_EMAIL: MIT ODL Video <odl-video-support@mit.edu>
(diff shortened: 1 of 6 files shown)


markbreedlove:
I've just thought of something else. If the changes above are basically OK, then I should add a logrotate configuration to rotate this new file.

blarghmatey:
I would create a directory for the log files that we can set appropriate ownership on. e.g. /var/log/odl-video/django.log

markbreedlove:
@blarghmatey OK, I've changed the location and added a directory.

blarghmatey:
:+1:

markbreedlove:
@blarghmatey : I've just added a logrotate config for that new Django log.

blarghmatey:
:+1:

PR #1221, opened 2020-04-30
Adding IAM role definition for MIT Open

blarghmatey:
Added role for granting permissions on S3 buckets needed by MIT Open app

### DIFF
--- a/pillar/vault/roles/aws.sls
+++ b/pillar/vault/roles/aws.sls
@@ -125,4 +125,43 @@ vault:
       name: odl-video-service-{{ env }}
       options:
         policy: '{{ ovs_policy|json }}'
+    {% load_json as mit_open_policy %}
+    {
+      "Statement": [
+        {
+          "Resource": [
+            "arn:aws:s3:::odl-discussions-{{ env }}",
+            "arn:aws:s3:::odl-discussions-{{ env }}/*"          ],
+          "Action": [
+            "s3:HeadObject",
+            "s3:Get*",
+            "s3:List*",
+            "s3:PutObject",
+            "S3:DeleteObject",
+          ],
+          "Effect": "Allow"
+        },
+        {
+          "Resource": [
+            "arn:aws:s3:::mitx-etl-xpro-production-mitxpro-production",
+            "arn:aws:s3:::mitx-etl-xpro-production-mitxpro-production/*",
+            "arn:aws:s3:::ol-olx-course-exports",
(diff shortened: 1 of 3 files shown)


pillar/vault/roles/aws.sls
```diff
@@ -125,4 +125,43 @@ vault:
...
+        {
+          "Resource": [
+            "arn:aws:s3:::odl-discussions-{{ env }}",
+            "arn:aws:s3:::odl-discussions-{{ env }}/*"          ],
+          "Action": [
+            "s3:HeadObject",
+            "s3:Get*",
+            "s3:List*",
+            "s3:PutObject",
+            "S3:DeleteObject",
```
shaidar, line 140: Are those actions enough or does this need more extensive ones similar to OVS?

pillar/vault/roles/macros.jinja
```diff
@@ -1,23 +1,21 @@
...
-        {% raw %}
         sql: >-
-          CREATE USER "{{name}}" WITH PASSWORD '{{password}}' VALID UNTIL '{{expiration}}';
-          GRANT "{{name}}" TO "bootcamp-ecommerce" WITH ADMIN OPTION;
-          GRANT "bootcamp-ecommerce" TO "{{name}}";
-          GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO "{{name}}";
-          GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA public TO "{{name}}";
-          ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT ALL PRIVILEGES ON TABLES TO "{{name}}";
-          ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT ALL PRIVILEGES ON SEQUENCES TO "{{name}}";
+          {% raw %}CREATE USER "{{name}}" WITH PASSWORD '{{password}}' VALID UNTIL '{{expiration}}';{% endraw %}
```
shaidar, line 3: Why the change from having the entire block using `raw` to one per line?

blarghmatey: So that we can use the approle parameter in the rendered Jinja

pillar/vault/roles/aws.sls
blarghmatey: This should be plenty. The listing for OVS is every individual S3 permission, most of which it doesn't need.

shaidar:
👍

PR #1223, opened 2020-05-04
Fix ALLOWED_HOSTS for bootcamps

rhysyngsun:
N/A

#### What's this PR do?
Fixes the `ALLOWED_HOSTS` value to quote the wildcard so that the app will parse it.

- Verify in CI/RC that you're getting a Bad Request error trying to hit the site
- Apply this in CI/RC and verify the site is reachable now


shaidar:
👍

PR #1228, opened 2020-05-13
Doc explaining used cybersource keys

shaidar:
#### What's this PR do?
Overview of the CyberSource used in some of our apps to clarify some of the mixups that occur when we try and request the renewal of some of those keys.

### DIFF
--- a/docs/runbooks/cybersource_keys.md
+++ b/docs/runbooks/cybersource_keys.md
@@ -1,0 +1,63 @@
+# CyberSource
+CyberSource is the payment processing and fraud management third party that the Institute uses for its online transactions. Additionally, CyberSource provides the ability to run OFAC checks for compliance which we leverage in our apps.
+
+Currently, any requests for new keys or renewal of existing ones need to go through Peter as he is the one with an account to the CyberSource portal.
+
+## Vault
+Ideally, we should store all cybersource related values under the following vault path:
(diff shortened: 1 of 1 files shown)


blarghmatey:
Can you just add a sample email to our Cybersource contact to identify the phrasing the gives us the information that we need and avoids the confusion that we've dealt with in the past?

shaidar:
Well I purposefully didn't include it given that this is a public doc and wasn't sure whether it's appropriate to include the name and address of said contact

shaidar:
The main confusion has typically been around the `TRANSACTION_KEY` and I included that this is referred to as the SOAP key so that would be the thing to keep an eye on. I also added the character lengths of the keys so that we know what to expect

blarghmatey:
I wasn't thinking of including contact information, just an example body of an email that you've used so that we can copy/paste for future requests.

shaidar:
Well the part I'm not very clear on is what Peter has access to and can do and what needs to be sent to our Cybersource contact since I've never seen the portal before. So typically things start with Peter and if he can't or doesn't have access to do something, then that gets forwarded to the contact. That's why there really isn't a generic email that I typically compose and thus the back and forth, however what has helped in the past was clarifying what the keys we're expecting look like as far as character lengths and as stated previously that `TRANSACTION_KEY` = `SOAP` key

blarghmatey:
Roger that. :+1:

PR #1233, opened 2020-05-19
Adding config for Bootcamps access to S3

blarghmatey:
----

#### What's this PR do?
Adds config to Heroku for accessing S3 to be used for Wagtail uploads, as well as adding logic to create the necessary buckets.

The Wagtail code for bootcamps was copied from xPRO but the necessary settings didn't get set up yet. This should resolve that, allowing media uploads to function properly.

### DIFF
--- a/pillar/heroku/bootcamps.sls
+++ b/pillar/heroku/bootcamps.sls
@@ -65,6 +65,9 @@ heroku:
   api_key: __vault__::secret-operations/global/heroku/api_key>data>value
   config_vars:
     ALLOWED_HOSTS: '["*"]'
+    AWS_ACCESS_KEY_ID:  __vault__:cache:aws-mitx/creds/read-write-delete-ol-bootcamps-app-{{ env_data.env_name }}>data>access_key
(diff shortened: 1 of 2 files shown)


shaidar:
Aren't you missing `BOOTCAMP_ECOMMERCE_USE_S3` and `AWS_S3_FILE_OVERWRITE`?

blarghmatey:
I'm not sure what the purpose of those settings are or what the intended values should be, so we can handle that in a separate PR.

salt/orchestrate/aws/s3_buckets/xpro.sls
```diff
@@ -1,5 +1,3 @@
-{% for env in ['mitxpro-qa', 'mitxpro-production'] %}
```
shaidar, line 1: What does this file have to do with bootcamps?

blarghmatey: Nothing, aside from the fact that I used it as a template for the code to create bootcamp buckets and I saw the bug while I was in there.

shaidar:
👍

PR #1234, opened 2020-05-20
Enable beats

markbreedlove:
Does this look OK? It's a small change, but I just wanted to check.
My reference for the index date format is https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html via https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html#custom-date-formats

### DIFF
--- a/pillar/elastic_stack/beats/init.sls
+++ b/pillar/elastic_stack/beats/init.sls
@@ -19,6 +19,7 @@ elastic_stack:
           hosts:
             - http://operations-elasticsearch.query.consul:9200
           compression_level: 3
+          index: "metricbeat-%{[agent.version]}-%{+yyyy.w}"
       modules:
         system:
           - module: system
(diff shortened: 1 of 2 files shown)


markbreedlove:
I think I actually need to add some more lines to specify lifecycle management options. I forgot that this is related to index templates, and Beats has config file options that I need to set.

markbreedlove:
I'm going to leave the lifecycle management options at their Metricbeat defaults to see how they work. [Their documentation](https://www.elastic.co/guide/en/beats/metricbeat/current/ilm.html) says that it will use ILM by default, and will create a policy named `metricbeat`.  I want to see how feasible it is to tune the retention policy from within Kibana.

blarghmatey:
:+1:

blarghmatey:
You'll need to manually run the beats state to get it installed. We should also probably change the collection period from 1 second to 5 or so to avoid flooding the disks.

blarghmatey:
:+1:

PR #1235, opened 2020-05-21
Change elk version includes

markbreedlove:
Include the pillar that defines the ELK stack version for all minions, depending only on whether they're in our QA or production inventory.

Now that we're trying to install Beats on most of our minions, we need the ELK version defined on them.

### DIFF
--- a/pillar/top.sls
+++ b/pillar/top.sls
@@ -7,7 +7,7 @@ base:
   'P@environment:(rc.*|.*-qa)':
     - match: compound
     - elastic_stack.version_qa
-  'not P@environment:(rc.*|.*-qa|ci)':
+  'not P@environment:(rc.*|.*-qa)':
     - match: compound
     - elastic_stack.version_production
   'roles:auth_server':
--- a/pillar/top.sls
+++ b/pillar/top.sls
@@ -5,7 +5,7 @@ base:
     - environment_settings
     - fluentd
   'P@environment:(rc.*|.*-qa|ci)':
-    - match: copound
+    - match: compound
     - elastic_stack.version_qa
   'not P@environment:(rc.*|.*-qa|ci)':
     - match: compound
--- a/pillar/top.sls
+++ b/pillar/top.sls
@@ -4,7 +4,7 @@ base:
     - common
     - environment_settings
     - fluentd
-  'P@environment:(rc.*|.*-qa|ci)':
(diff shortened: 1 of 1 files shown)


pillar/top.sls
```diff
@@ -4,6 +4,12 @@ base:
     - common
     - environment_settings
     - fluentd
+  'P@environment:(rc.*|.*-qa|ci)':
```
blarghmatey, line 7: We don't have any `CI` environments in any of the AWS infrastructure. That's a Heroku only concept.

pillar/top.sls
```diff
@@ -4,6 +4,12 @@ base:
...
+    - match: compound
+    - elastic_stack.version_qa
+  'not P@environment:(rc.*|.*-qa|ci)':
```
blarghmatey, line 10: You can just do a regex of `.*production.*` for the environment grain.

blarghmatey: I guess that's not entirely accurate because of the `operations` env...

pillar/top.sls
markbreedlove: I added it because I saw this:
```
proxy-mitxpro-ci:
    ci
```
Shall I skip that?

pillar/top.sls
markbreedlove: We have the environments `ocw` and `operations` in our production VPN.

pillar/top.sls
blarghmatey: Yeah, the proxy minions don't actually have any host that they interact with and only exist for populating settings on Heroku.

pillar/top.sls
markbreedlove: Should I treat those separately, perhaps?

blarghmatey: No, it's fine to do it as a single one. The original approach of using `not` is probably easiest.

pillar/top.sls
markbreedlove: OK. I've removed "ci" in commit 80439006.

blarghmatey:
:+1:

pillar/top.sls
markbreedlove: OK. I kept it as one condition, but I removed "ci" in commit 9d2548ca.

I'll have to use care when adding Beats in `salt/top.sls` and not have conditions that are too loose and install it on the wrong minions.

PR #1238, opened 2020-05-27
Use https for bootcamp URLs

noisecapella:
None

#### What's this PR do?
Updates URLs to use https. This is required for the bootcamp site URL because http URLs are redirected to https URLs which causes an error with the Jobma webhook.

N/A


shaidar:
👍

PR #1243, opened 2020-06-09
Added xpack to logging qa config

shaidar:
#### What's this PR do?
This PR configures the use of the basic (free) xpack license to accomplish the following:
- TLS encryption between ES nodes in cluster
- TLS encryption between Kibana, fluentd aggregators, metricbeats agents, and ES cluster

##### Note: The following tasks need to happen before merging this PR:
- A PR should also be submitted against Elasticsearch formula to create the cert key and files referenced in pillar.
- Values added to sdb
- Keys/values added to vault

This has been manually tested on the existing logging QA cluster

#### Useful URLs
- https://www.elastic.co/guide/en/elasticsearch/reference/master/configuring-tls.html#tls-active-directory
- https://www.elastic.co/guide/en/kibana/current/configuring-tls.html
- https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html
- https://www.elastic.co/guide/en/elasticsearch/reference/current/security-privileges.html
- https://www.elastic.co/subscriptions

### DIFF
--- a/pillar/consul/elasticsearch.sls
+++ b/pillar/consul/elasticsearch.sls
@@ -9,5 +9,6 @@ consul:
           - logging
         {% endif %}
         check:
-          http: 'http://localhost:9200/'
+          tcp: 'localhost:9200'
           interval: 10s
+          timeout: 3s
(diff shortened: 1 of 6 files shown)


shaidar:
@blarghmatey I'm tempted to close this for now since we won't be doing anything about it in the near-term

blarghmatey:
I think that makes sense. If we do anything in this direction I think it will be with the [Open Distro for Elasticsearch](https://opendistro.github.io/for-elasticsearch/)

PR #1245, opened 2020-06-22
Updating cloud profiles to use AMD instance types

blarghmatey:
AMD instance types are more cost effective, so moving to use those where possible. Also updated older generation types to improve cost

----

#### What's this PR do?
Changes instance types in cloud profiles to use more cost effective options.


shaidar:
👍

PR #1251, opened 2020-06-29
Moved jobma settings outside if conditional

shaidar:
#### What's this PR do?
Now that we have Jobma settings for production which I have added to vault, we no longer need to have settings under a conditional. I moved those settings into the main part, however I left the `if` conditional as I suspect we will be using it down the line and it might be easier to keep it there for devs when they submit PR's.


blarghmatey:
:+1:

PR #1253, opened 2020-07-02
Add OCW CMS's web dir to OCW sync_repo state

markbreedlove:
In the OCW sync_repo state, which deploys files to the OCW servers, update the behavior for the CMS servers to transfer static web files from the 'web' folder of the repo to the `/var/www/html` docroot directory on the server.

This is a different docroot folder than on the origin servers.

The "delete" option can be used with rsync, because, unlike on the origin servers,  there's no generated content (courses,  json files, rss feeds) that get put into that directory by other scripts.


blarghmatey:
:+1:

PR #1255, opened 2020-07-10
ProctorTrack config using edxapp public values

shaidar:
#### What's this PR do?
Removed previous changes for PT settings and used edxapp defined keys instead


blarghmatey:
:+1:

PR #1256, opened 2020-07-10
Fix grain deletion error

shaidar:
#### What's this PR do?
`build_ami` state was failing on the `grains.delkey` function with the following error:

`The key 'roles' exists but is a dict or a list. Use 'force=True' to overwrite.`

This changes the function to using the `grains.absent` state and force the removal of keys and values.

### DIFF
--- a/salt/orchestrate/edx/build_ami.sls
+++ b/salt/orchestrate/edx/build_ami.sls
@@ -217,9 +217,13 @@ delete_{{ grain }}_from_grains:
   salt.function:
     - tgt: 'edx*{{ ENVIRONMENT}}*base'
     - tgt_type: compound
-    - name: grains.delkey
-    - arg:
-        - {{ grain }}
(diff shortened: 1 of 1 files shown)


salt/orchestrate/edx/build_ami.sls
```diff
@@ -217,9 +217,13 @@ delete_{{ grain }}_from_grains:
...
     - tgt: 'edx*{{ ENVIRONMENT}}*base'
     - tgt_type: compound
-    - name: grains.delkey
-    - arg:
-        - {{ grain }}
+    - name: state.single
+    - arg: 
+      - grains.absent
+    - kwarg:
+        {{ grain }}
```
blarghmatey, line 224: This needs to be a key: value

shaidar: Fixed

PR #1257, opened 2020-07-13
Prevent fluentd pillar inclusion for devstack role

markbreedlove:

https://trello.com/c/4Op8y52z/148-migrate-and-upgrade-micromasters-sandbox

#### What's this PR do?

Since the `devstack` role has been prevented from including FluentD states in `salt/top.sls`, prevent it also from including fluentd pillar data `in pillar/top.sls`. Otherwise, we get a pillar rendering error when doing a highstate.

### DIFF
--- a/pillar/top.sls
+++ b/pillar/top.sls
@@ -4,6 +4,7 @@ base:
     - common
     - environment_settings
   '* and not proxy-* and not G@roles:devstack'
+    - match: compound
     - fluentd
   'P@environment:(rc.*|.*-qa)':
     - match: compound
--- a/pillar/top.sls
+++ b/pillar/top.sls
@@ -3,6 +3,7 @@ base:
     - match: compound
     - common
     - environment_settings
+  '* and not proxy-* and not G@roles:devstack'
     - fluentd
   'P@environment:(rc.*|.*-qa)':
     - match: compound


pillar/top.sls
```diff
@@ -3,6 +3,7 @@ base:
     - match: compound
     - common
     - environment_settings
+  '* and not proxy-* and not G@roles:devstack'
```
blarghmatey, line 6: You need to add the match type

markbreedlove: oh, yeah! I've just added it.

markbreedlove:
@blarghmatey Do you think this is OK now?

blarghmatey:
After taking another look at this, why is it getting a role of devstack? The devstack is what is used for local development. This should be getting a role of "sandbox". I guess it depends on how the instance is getting deployed, whether it's an all in a box, or if it's using the distributed deployment strategy. If it's all in a box then I guess this is fine.

markbreedlove:
Yeah, I originally thought "sandbox," too, but I had some conversation with you and @shaidar -- I believe it was last week -- when someone explained to me that I should be using "devstack" for all-in-one boxes.

blarghmatey:
:+1:

PR #1258, opened 2020-07-13
Added null to proctoring backend

shaidar:
#### What's this PR do?
Instructor dashboard on courses that doesn't use PT are receiving 500 since "null" value is no longer defined. This add it back


blarghmatey:
:+1:

PR #1259, opened 2020-07-14
Create file when tracking data is uploaded

shaidar:
#### What's this PR do?
Minor tweak to generate a file once tracking data has been zipped and uploaded to S3. Currently to check if those logs have been backed up prior to terminating an instance, we need to manually inspect S3. This just helps inspect it from the salt-master.


blarghmatey:
So, if the upload has been done then there should already be a tar.gz archive on disk. What would be good is to run a script to verify that the archive has all of the files present on the machine, because there are cases where we run the tracking log upload out of band of terminating the instance, so we want to be sure that we have all of the data.

shaidar:
Well, the tar file can be there, but then it could've failed at uploading to s3 for whatever reason, so the file being on disk, doesn't always guarantee that it's in S3.

blarghmatey:
Fair point. Might be worth doing both, also might be good to populate the sentinel file with the timestamp of when it was generated.

PR #1262, opened 2020-07-21
Added Akismet configuration for Open

rhysyngsun:
Part of https://github.com/mitodl/open-discussions/issues/3062

#### What's this PR do?
Adds configuration necessary for spam checking via Akismet introduced in https://github.com/mitodl/open-discussions/pull/3063

Configuration is applied to necessary environments and then tested with test data.

I added `AKISMET_BLOG_URL` as an additional value in `env_dict` since this value needs to match what is configured in the Akismet account. In practice this will be the same as the domain name for the app, but keep it as a separate value future-proofs this against an accidental breakage if the app moves.

### DIFF
--- a/pillar/heroku/discussions.sls
+++ b/pillar/heroku/discussions.sls
@@ -3,6 +3,7 @@
 
 {% set env_dict = {
     'ci': {
+      'AKISMET_BLOG_URL': 'https://discussions-ci.odl.mit.edu',
       'app_log_level': 'INFO',
       'app_name': 'odl-open-discussions-ci',
       'CLOUDFRONT_DIST': 'd28ic9ywb63ioi',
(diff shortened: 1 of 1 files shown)


shaidar:
👍

pillar/heroku/discussions.sls
```diff
@@ -108,6 +111,8 @@ heroku:
   app_name: {{ env_data.app_name }}
   api_key: __vault__::secret-operations/global/heroku/api_key>data>value
   config_vars:
+    AKISMET_API_KEY: __vault__::secret-operations/{{ env_data.vault_env_path }}/{{ business_unit }}/akismet>data>api_key
```
shaidar, line 114: If CI api_key is gonna be different than RC, then this should be changed from `env_data.vault_env_path` to `env_data.env_name`

rhysyngsun: done

shaidar:
👍

PR #1265, opened 2020-07-22
Add elastalert alert for SAML failure in LMS

markbreedlove:

https://trello.com/c/o9bAwz8Y/158-elastalert-for-saml-failure-in-lms

#### What's this PR do?

Adds an Elastalert alert for messages about authentication failures.

Is there anything that I could add to the description field to suggest to the recipient how to troubleshoot the issue?

### DIFF
--- a/pillar/elastalert.sls
+++ b/pillar/elastalert.sls
@@ -526,3 +526,30 @@ elastic_stack:
                   - query_string:
                       default_field: message
                       query: (received unregistered task) AND (message has been ignored)
+      - name: lms_saml_failure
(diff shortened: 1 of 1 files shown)


shaidar:
There's a scheduled state that runs on those instance that should fix this, so maybe add that to the alert as a fix? https://github.com/mitodl/salt-ops/blob/main/salt/edx/refresh_saml_provider_metadata.sls

markbreedlove:
Thanks, @shaidar! I have added some more information to the alert text and description fields.

PR #1266, opened 2020-08-03
Feature flag for spam exemptions

mbertrand:
Related to https://github.com/mitodl/open-discussions/issues/3086

#### What's this PR do?
Adds a new feature flag

N/A


shaidar:
If we're setting the value to `True` in all envs, I would think that we might as well just hardcode it in the pillar as opposed to setting it in each env.

mbertrand:
@shaidar thanks, done

shaidar:
👍

PR #1267, opened 2020-08-11
Adding pillar data and packer config for building Dagster AMI

blarghmatey:
In order to simplify our build and release process I would like to introduce Packer as our interface for creating deployable images. This PR bootstraps that effort with configuration and changes needed to get a Dagster AMI built.

----

https://trello.com/c/1fdWL6ju/130-deploy-dagster-to-production

#### What's this PR do?
Adds a Packer configuration for building a Dagster AMI

Run `packer build packer/dagster.json` from the salt-ops directory

We have typically used the Salt orchestration system for building and deploying instances, but it has historically been an error prone and manual process. This is a step in the direction of decomposing some of our process to rely on more best of breed tools and then integrating them to build a more stable whole.

### DIFF
--- a/pillar/dagster/init.sls
+++ b/pillar/dagster/init.sls
@@ -1,0 +1,50 @@
+# -*- mode: yaml -*-
+{% set environment = salt.grains.get('environment') %}
+
+dagster:
+  home: /opt/ol_data_pipelines
+  dagit:
+    path: /opt/ol_data_pipelines/bin
+    flags: -w etc/workspace.yaml
+  config:
+    instance:
+      scheduler:
+        module: dagster_cron.cron_scheduler
+        class: SystemCronScheduler
+      compute_logs:
+        module: dagster_aws.s3.compute_log_manager
+        class: S3ComputeLogManager
+        config:
+          bucket: dagster-{{ environment }}/compute-logs/
+      run_storage:
+        module: dagster_postgres.run_storage
+        class: PostgresRunStorage
+        config:
+          postgres_db:
+            username: __vault__:cache:postgres-{{ environment }}/creds/app
+            password: __vault__:cache:postgres-{{ environment }}/creds/app
+            hostname: dagster_db.service.{{ environment }}.consul
+            db_name: dagster
+            port: 5432
+      event_log_storage:
+        module: dagster_postgres.event_log
+        class: PostgresEventLogStorage
+        config:
+          postgres_db:
+            username: __vault__:cache:postgres-{{ environment }}/creds/app
+            password: __vault__:cache:postgres-{{ environment }}/creds/app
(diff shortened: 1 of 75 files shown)


pillar/dagster/init.sls
```diff
@@ -0,0 +1,50 @@
...
+        class: S3ComputeLogManager
+        config:
+          bucket: dagster-{{ environment }}/compute-logs/
+      run_storage:
+        module: dagster_postgres.run_storage
+        class: PostgresRunStorage
+        config:
+          postgres_db:
+            username: __vault__:cache:postgres-{{ environment }}/creds/app
+            password: __vault__:cache:postgres-{{ environment }}/creds/app
```
shaidar, line 25: Shouldn't the `app` be defined here?

salt/environment_settings.yml
```diff
@@ -145,6 +145,42 @@ sandbox_mitxpro_versions: &sandbox_mitxpro_versions
...
+        num_instances: 1
+        domains:
+          - pipelines-qa.odl.mit.edu
+        business_unit: data
+        security_groups:
+          - webapp-odl-vpn
+        size: r5a.large
+  data-production:
+    business_unit: data
+    network_prefix: 10.3
```
shaidar, line 168: Shouldn't this be a string?

salt/environment_settings.yml
```diff
@@ -145,6 +145,42 @@ sandbox_mitxpro_versions: &sandbox_mitxpro_versions
 
 
 environments:
+  data-qa:
+    business_unit: data
+    network_prefix: 10.2
```
shaidar, line 150: Shouldn't this be a string?

salt/top.sls
```diff
@@ -70,6 +70,10 @@ base:
     - match: grain
     - pulsar
     - pulsar.tests
+  'roles:dagster':
+    - match: grain
+    - dagster
+    - nginx
```
shaidar, line 76: Should the `data` environment be added to [here](https://github.com/mitodl/salt-ops/blob/main/salt/top.sls#L10) so that consul get installed with prod configs?

pillar/dagster/init.sls
blarghmatey: So, for the work I did on the VaultDatabase component in Pulumi I decided to just create a generic `app` role rather than creating a role defined based on the name of the app, which is what I'm leaning on here. Happy to reconsider that pattern if you or @markbreedlove has a preference.

salt/top.sls
blarghmatey: Eventually we will want to do that, but for the scope of this PR I'm mainly focusing on just getting the AMI to build. There will likely be other changes needed for the final deployment which we can bundle into a separate PR.

shaidar:
👍

PR #1268, opened 2020-08-13
add config for allowing ocw-next through CORS on open-rc

aliceriot:

for this PR to work correctly in 'prod': https://github.com/mitodl/hugo-course-publisher/pull/172

#### What's this PR do?

adds the hostname for ocw-next to the `OPEN_DISCUSSIONS_CORS_ORIGIN_WHITELIST` env var for discussions-rc.

### DIFF
--- a/pillar/heroku/discussions.sls
+++ b/pillar/heroku/discussions.sls
@@ -60,7 +60,8 @@
       'SOCIAL_AUTH_MICROMASTERS_LOGIN_URL': 'https://micromasters-rc.odl.mit.edu/login/edxorg/?next=/discussions/',
       'SOCIAL_AUTH_SAML_SP_ENTITY_ID': 'https://discussions-rc.odl.mit.edu/saml/metadata',
       'TIKA_SERVER_ENDPOINT': 'https://tika-rc-apps.odl.mit.edu',
-      'vault_env_path': 'rc-apps'
+      'vault_env_path': 'rc-apps',
+      'OCW_NEXT_URL': 'ocw-next.netlify.app'
       },
     'production': {
       'app_log_level': 'INFO',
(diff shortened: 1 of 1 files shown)


pillar/heroku/discussions.sls
```diff
@@ -180,7 +181,11 @@ heroku:
     OPEN_DISCUSSIONS_BASE_URL: {{ env_data.OPEN_DISCUSSIONS_BASE_URL }}
     OPEN_DISCUSSIONS_COOKIE_DOMAIN: {{ env_data.OPEN_DISCUSSIONS_COOKIE_DOMAIN }}
     OPEN_DISCUSSIONS_COOKIE_NAME: {{ env_data.OPEN_DISCUSSIONS_COOKIE_NAME}}
+    {% if env_data.OCW_NEXT_URL is defined %}
```
blarghmatey, line 184: I think this would need to be `if env_data.get('OCW_NEXT_URL')` as the `defined` test is for a given variable, whereas this is using attribute lookup.

aliceriot: cool, will change that

pillar/heroku/discussions.sls
```diff
@@ -181,7 +181,7 @@ heroku:
     OPEN_DISCUSSIONS_BASE_URL: {{ env_data.OPEN_DISCUSSIONS_BASE_URL }}
     OPEN_DISCUSSIONS_COOKIE_DOMAIN: {{ env_data.OPEN_DISCUSSIONS_COOKIE_DOMAIN }}
     OPEN_DISCUSSIONS_COOKIE_NAME: {{ env_data.OPEN_DISCUSSIONS_COOKIE_NAME}}
-    {% if env_data.OCW_NEXT_URL is defined %}
+    {% if env_data.get('OCW_NEXT_URL') is defined %}
```
blarghmatey, line 184: Sorry, wan't clear enough. In this case you can drop the `is defined` and rely on the truthiness of the return value of `.get` which will be `None` if the attribute isn't defined.

aliceriot: haha, no worries, I'll make the change now

blarghmatey:
:+1:

PR #1269, opened 2020-08-18
Update yarn and node versions to match value in package.json

shaidar:
#### What's this PR do?
Update yarn and node versions to the one specified in [package.json](https://github.com/mitodl/odl-video-service/blob/044fa9aa90cecbdc1a7c10eed7f12de93e988f6f/package.json#L117)


blarghmatey:
:+1:

PR #1272, opened 2020-08-20
Adding API configuration and provision Pulumi API user

blarghmatey:


shaidar:
Shouldn't [this](https://github.com/mitodl/salt-ops/blob/main/pillar/master/extra_config.sls#L21-L35) be removed?

shaidar:
👍

PR #1274, opened 2020-08-25
Fix Canvas integration settings

shaidar:
#### What's this PR do?
Canvas settings were being removed when config refresh was running as they don't appear to have been placed under the right heading. This should fix the issue.


blarghmatey:
:+1:

PR #1275, opened 2020-08-26
Added canvas config for production

shaidar:
#### What's this PR do?
- Removed Canvas config from next residential
- Added Canvas config to main residential sls pointing to vault which would include qa and prod envs


blarghmatey:
:+1:

PR #1277, opened 2020-08-28
Added OVS nginx redirection links

shaidar:
[Ticket#60903](https://odl.zendesk.com/agent/tickets/60903)

#### What's this PR do?
Adds a few redirect links of vids that used to be on techtv and were references in articles and need to still be resolvable.


blarghmatey:
:+1:

PR #1278, opened 2020-09-02
add OCW prod domain to CORS list for Open

aliceriot:

part of getting production OCW deployed.

#### What's this PR do?

We need to add domains to a CORS allow-list in order to make requests to the Open APIs from them. This does that for production.

read through it?

### DIFF
--- a/pillar/heroku/discussions.sls
+++ b/pillar/heroku/discussions.sls
@@ -91,7 +91,8 @@
       'SOCIAL_AUTH_MICROMASTERS_LOGIN_URL': 'https://micromasters.mit.edu/login/edxorg/?next=/discussions/',
       'SOCIAL_AUTH_SAML_SP_ENTITY_ID': 'https://discussions.odl.mit.edu/saml/metadata',
       'TIKA_SERVER_ENDPOINT': 'https://tika-production-apps.odl.mit.edu',
-      'vault_env_path': 'production-apps'
+      'vault_env_path': 'production-apps',
+      'OCW_NEXT_URL': 'ocw-beta.odl.mit.edu'
       }
 } %}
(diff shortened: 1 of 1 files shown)


pillar/heroku/discussions.sls
```diff
@@ -91,7 +91,8 @@
       'SOCIAL_AUTH_MICROMASTERS_LOGIN_URL': 'https://micromasters.mit.edu/login/edxorg/?next=/discussions/',
       'SOCIAL_AUTH_SAML_SP_ENTITY_ID': 'https://discussions.odl.mit.edu/saml/metadata',
       'TIKA_SERVER_ENDPOINT': 'https://tika-production-apps.odl.mit.edu',
-      'vault_env_path': 'production-apps'
+      'vault_env_path': 'production-apps',
+      'OCW_NEXT_URL': 'ocw-beta.odl.mit.edu'
```
markbreedlove, line 95: The variable that I learned about recently for governing CORS access was `OPEN_DISCUSSIONS_CORS_ORIGIN_WHITELIST`.  Is this in lieu of that variable? Does `OPEN_DISCUSSIONS_CORS_ORIGIN_WHITELIST` need to be set, too?

aliceriot: so we set this value in this dictionary here and then further on down below it's pulled out and added to the `OPEN_DISCUSSIONS_CORS_ORIGIN_WHITELIST` variable, alongside another value for Micromasters

aliceriot: sorry I don't know the terminology for all this stuff because I'm not so familiar with this salt stuff

markbreedlove: Oh, I see. That's fine, thanks.

PR #1281, opened 2020-09-10
Pointed ubcpi to edx fork

shaidar:
#### What's this PR do?
Version `0.6.4` of ubcpi isn't compatible with python3 and the newer version has some [import issues](https://github.com/ubc/ubcpi/blob/0.6.5/ubcpi/ubcpi.py#L20). Switched to using the edx fork which we are using on residential running Juniper.


blarghmatey:
:+1:

PR #1282, opened 2020-09-10
Added max file upload size setting to bootcamps

gsidebo:
Bootcamp PR: https://github.com/mitodl/bootcamp-ecommerce/pull/996
PR amending that setting and nginx config: https://github.com/mitodl/bootcamp-ecommerce/pull/1004

#### What's this PR do?
Adds a setting for the max file upload size (currently users can only upload resumes). This will be 10MB for all environments


shaidar:
In reviewing this and looking at the relevant [PR](https://github.com/mitodl/bootcamp-ecommerce/commit/0f4394e0043a45dadf9e17454136e6784fd2c0c0), I think it might be a bit misleading that there's a Heroku setting allowing us to set the upload limit, however it's already statically set in the nginx config at 10MB. Hence, changing the value in the Heroku config has no effect.

shaidar:
👍

PR #1283, opened 2020-09-11
Updated wildcard cert

shaidar:
#### What's this PR do?
We uploaded a new cert to ACM but didn't update the state to point to the new cert. This fixes that issue


blarghmatey:
:+1:

PR #1288, opened 2020-09-24
Add OPEN_DISCUSSIONS_FRONTPAGE_DIGEST_MAX_POSTS

rhysyngsun:
N/A

#### What's this PR do?
This bumps the limit of the number of posts that appear in the frontpage digest email from the default of 5 to 10.

Deploy it in CI and ensure the app starts (if the config is incorrect the app will fail to start)


shaidar:
👍

PR #1289, opened 2020-09-25
Automate ocw nextgen build

markbreedlove:
This automates the configuration of the OCW Next build machine.


blarghmatey (changes requested):

pillar/apps/ocw-next-qa.sls
```diff
@@ -0,0 +1,13 @@
+
+ocw-next:
+  website_bucket: ocw-beta-qa-course-site
+  source_data_bucket: open-learning-course-data-rc
+  search_api_url: //open.mit.edu/api/v0/search/
```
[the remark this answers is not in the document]
markbreedlove: 👍  Changed it to use `discussions-rc.odl.mit.edu`.

pillar/apps/ocw-next-production.sls
```diff
@@ -0,0 +1,13 @@
+
+ocw-next:
+  website_bucket: ocw-beta-production-course-site
+  source_data_bucket: open-learning-course-data-production
+  search_api_url: //open.mit.edu/api/v0/search/
+  ocw_to_hugo_git_ref: master
```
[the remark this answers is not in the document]
markbreedlove: OK. We'll have to do that when they've created the branch or tag. There's no release branch yet.

salt/apps/ocw/nextgen_build_install.sls
```diff
@@ -0,0 +1,75 @@
...
+    - rev: {{ ocw_next.ocw_to_hugo_git_ref }}
+    - force_checkout: True
+    - force_clone: True
+    - force_reset: True
+    - force_fetch: True
+    - update_head: True
+    - user: ocw
+    - require:
+      - pkg: ensure_os_package_prerequisites
+  git.latest:
```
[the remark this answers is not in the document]
markbreedlove: 👍  Broke those states apart in b114d093

salt/apps/ocw/nextgen_build_publish.sls
```diff
@@ -0,0 +1,31 @@
...
+run_hugo_course_publisher:
+  cmd.run:
+    - name: npm run build
+    # `cwd' is specified because it drops a log file here.
+    - cwd: /home/ocw/hugo-course-publisher
+    - runas: ocw
+    - env:
+      - NODE_OPTIONS: '-max_old_space_size=8192'
+    - require:
+      - cmd: run_ocw_to_hugo
```
blarghmatey, line 21: It looks like this state file is missing the step to handle the course static content?

markbreedlove: There's no more action needed, because the static course files (PDFs, images) are served from the raw data bucket. Fastly takes any path starting with `/coursemedia` and serves it from that bucket, instead of the website bucket.

[This line](https://github.com/mitodl/salt-ops/blob/b114d093febfd3d60abf0852d324d7174a724118/salt/apps/ocw/nextgen_build_publish.sls#L7) executes ocw-to-hugo with the options necessary to have it replace S3 bucket URLs with URLs that start with this `/coursemedia` path prefix.

salt/apps/ocw/nextgen_build_publish.sls
```diff
@@ -0,0 +1,31 @@
...
+    - runas: ocw
+
+run_hugo_course_publisher:
+  cmd.run:
+    - name: npm run build
+    # `cwd' is specified because it drops a log file here.
+    - cwd: /home/ocw/hugo-course-publisher
+    - runas: ocw
+    - env:
+      - NODE_OPTIONS: '-max_old_space_size=8192'
```
blarghmatey, line 19: We don't need this value. That was a relic of an effort to resolve a bug that was due to duplicated content and the Hugo process, not the Node process.

salt/top.sls
```diff
@@ -295,3 +295,10 @@ base:
     - edx.prod
     - rabbitmq.configure
     - edx.django_user
+  'roles:ocw-next-build':
```
blarghmatey, line 298: This needs to be `ocw-build` to match what is being set in Pulumi

pillar/top.sls
```diff
@@ -346,3 +346,6 @@ base:
     - fluentd.ocw_db
   'P@roles:ocw-(cms|db|origin|mirror) and G@ocw-environment:production':
     - datadog
+  'G@roles:ocw-next-build and G@ocw-environment:production':
```
blarghmatey, line 349: The ocw-environment isn't going to be set as this isn't getting deployed to the OCW environment. Also, we need to have a match statement for both QA and production environments to load the separate pillar files. Alternatively, we should combine the two pillar files into a single one using a mapping object to handle the differences.

pillar/apps/ocw-next-production.sls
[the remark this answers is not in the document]
markbreedlove: 👍  I have updated the branches in f1956172

salt/apps/ocw/nextgen_build_publish.sls
markbreedlove: Removed in 0711b181

pillar/top.sls
markbreedlove: Can you check my thinking here?
I'm seeing that the salt minion's environment grain(?) appears to be set [here](https://github.com/mitodl/ol-infrastructure/blob/17e4ebb49185683e227e26d96b3b00091b88e3c7/src/ol_infrastructure/applications/ocw_build/__main__.py#L153). If I understand this correctly, it's going to have an environment grain (`environment`?) that equals `applications-production`, based on what I see [here](https://github.com/mitodl/ol-infrastructure/blob/17e4ebb49185683e227e26d96b3b00091b88e3c7/src/ol_infrastructure/applications/ocw_build/__main__.py#L38). Is that right?

Therefore, I'm thinking this line in top.sls should say `G@roles:ocw-build and G@environment:applications-production`.

blarghmatey: Yes, that is correct

markbreedlove: Thanks!

Changed in f4a6b85a

blarghmatey (approved):

blarghmatey:
:+1:

PR #1291, opened 2020-09-28
Changed sandbox to use python 3.5

shaidar:
#### What's this PR do?
Change to use python 3.5 in the residential sandbox environment.


blarghmatey:
:+1: but we'll also want to revert https://github.com/mitodl/configuration/commit/57538d2036004080c4f7c44ca97c0d915bc604d9

PR #1292, opened 2020-09-29
Add date to compressed filename

shaidar:
#### What's this PR do?
Backup tracking logs uploaded to s3 by running the state could potentially overwrite existing files as the instances may end up having the same version number. This PR adds the current date to the tarred file name to avoid that scenario.


blarghmatey:
:+1:

PR #1293, opened 2020-10-01
Added config for NovoEd SSO via SAML in Bootcamps

gsidebo:
Related to https://github.com/mitodl/bootcamp-ecommerce/pull/1015

#### What's this PR do?
Adds settings for enabling SSO with NovoEd

We'll need new key and cert values for production. For other envs I can probably just send the key/cert that I've been using over Keybase. These are defined like key/cert settings values in other projects (i.e.: long string with line breaks replaced with `\n`)


pillar/heroku/bootcamps.sls
```diff
@@ -138,6 +141,9 @@ heroku:
     NOVOED_API_BASE_URL: https://api.novoed.com/
     NOVOED_API_KEY: __vault__::secret-operations/global/{{ business_unit }}/novoed>data>api_key
     NOVOED_API_SECRET: __vault__::secret-operations/global/{{ business_unit }}/novoed>data>api_secret
+    NOVOED_SAML_DEBUG: {{ env_data.NOVOED_SAML_DEBUG }}
+    NOVOED_SAML_KEY: __vault__::secret-operations/global/{{ business_unit }}/novoed>data>saml_key
+    NOVOED_SAML_CERT: __vault__::secret-operations/global/{{ business_unit }}/novoed>data>saml_cert
```
shaidar, line 146: Since the key and cert are different for RC/CI and Prod, I'd change the paths to this:
`__vault__::secret-{{ business_unit }}/{{ env_data.vault_env_path }}/novoed>data>saml_{key,cert}`

shaidar: @gsidebo `global` shouldn't be in the path since it's env dependent. I would just change it to look exactly like the path I listed

gsidebo: Whoops! Good catch. Fixed now

shaidar:
👍

PR #1294, opened 2020-10-02
Added setting for SAML base URL in Bootcamps

gsidebo:
Related to https://github.com/mitodl/bootcamp-ecommerce/pull/1034

#### What's this PR do?
Adds a production setting for the SAML base URL in Bootcamps (only needed for production – we can use the default base URL in other envs)


gsidebo:
@shaidar Do I need to define that env variable for CI and RC in the dictionary even though it can be empty for those environments? If so, I'll make that change

shaidar:
No, but I'd probably just move it to this section - https://github.com/mitodl/salt-ops/blob/main/pillar/heroku/bootcamps.sls#L109-L116

shaidar:
👍

PR #1296, opened 2020-10-06
Added NovoEd SAML login URL for Bootcamps

gsidebo:
https://github.com/mitodl/bootcamp-ecommerce/pull/1041

#### What's this PR do?
Adds a setting for the SP-initiated NovoEd SAML login URL 

We can only have a staging and a production environment for SAML, which is why I didn't specify this setting value for CI

### DIFF
--- a/pillar/heroku/bootcamps.sls
+++ b/pillar/heroku/bootcamps.sls
@@ -21,6 +21,7 @@
       'HUBSPOT_FOOTER_FORM_GUID': 'a8cffaac-4207-446a-b295-5f44656596f8',
       'MAILGUN_SENDER_DOMAIN': 'mail-rc.bootcamp.odl.mit.edu',
       'NOVOED_SAML_DEBUG': True,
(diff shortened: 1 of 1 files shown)


pillar/heroku/bootcamps.sls
```diff
@@ -146,6 +148,7 @@ heroku:
     NOVOED_SAML_DEBUG: {{ env_data.NOVOED_SAML_DEBUG }}
     NOVOED_SAML_KEY: __vault__::secret-{{ business_unit }}/{{ env_data.vault_env_path }}/novoed>data>saml_key
     NOVOED_SAML_CERT: __vault__::secret-{{ business_unit }}/{{ env_data.vault_env_path }}/novoed>data>saml_cert
+    NOVOED_SAML_LOGIN_URL: {{ env_data.NOVOED_SAML_LOGIN_URL }}
```
shaidar, line 151: That would fail for CI as it was be undefined. Don't think there would be harm in hooking up CI to staging as well so maybe just add the url to ci as well.

gsidebo: Done 👍

shaidar:
👍

PR #1297, opened 2020-10-07
Install Node 12.x on OCW build machine

markbreedlove:
Install Node 12 instead of the default. The `ocw-to-hugo` and `hugo-course-publisher` apps need version 12, which can be different than the Debian distribution's version.

### DIFF
--- a/pillar/apps/ocw-next-production.sls
+++ b/pillar/apps/ocw-next-production.sls
@@ -8,6 +8,7 @@ ocw-next:
 
 node:
   version: 12.18.4
-  install_from_ppa: True
-  ppa:
-    repository_url: https://deb.nodesource.com/node_12.x
+  pkg:
+    use_upstream_repo: True
+    archive:
+      uri: https://deb.nodesource.com/node_12.x
--- a/pillar/apps/ocw-next-production.sls
+++ b/pillar/apps/ocw-next-production.sls
@@ -9,4 +9,5 @@ ocw-next:
 node:
   version: 12.18.4
   pkg:
+    version: 12
     use_upstream_repo: True
--- a/pillar/apps/ocw-next-production.sls
+++ b/pillar/apps/ocw-next-production.sls
@@ -10,5 +10,3 @@ node:
   version: 12.18.4
   pkg:
     use_upstream_repo: True
-    archive:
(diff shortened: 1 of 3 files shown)


blarghmatey:
Isn't this already handled with the Node formula?

markbreedlove:
But of course there is a flaw in this pr: I didn't take out the node formula from top.sls. I'll check for a version option, hoping I can just keep the formula ...

markbreedlove:
I specified version 12.18.4 in the pillar but it did not take: https://github.com/mitodl/salt-ops/blob/153aabab207973689cb4d543eafd592c062e2f30/pillar/apps/ocw-next-qa.sls#L10

So I will investigate why that happened first.

blarghmatey:
I think you need to add `use_upstream_repo: true` in the node pillar https://github.com/mitodl/node-formula/blob/main/pillar.example#L11

markbreedlove:
I think that 1640e5b2 might be a little better. I can't find any advice on using this repo in the formula's docs, and there's just an example pillar, that doesn't cover my use-case.

pillar/apps/ocw-next-production.sls
```diff
@@ -8,6 +8,7 @@ ocw-next:
 
 node:
   version: 12.18.4
-  install_from_ppa: True
-  ppa:
-    repository_url: https://deb.nodesource.com/node_12.x
+  pkg:
+    use_upstream_repo: True
+    archive:
```
blarghmatey, line 13: From looking at the formula I don't think there's any need for the `archive` attribute. It looks like it uses the version definition to populate the appropriate repo URL

markbreedlove: OK, just removed it in 5d0ec698

blarghmatey:
I just pulled in a PR that was open on the upstream repo that fixes management of the repo definition. https://github.com/saltstack-formulas/node-formula/pull/53/commits/30ec6532b9a9b71586ee9ea65473c27ac53881ba

That should just require us to set the major version under the package definition now.

```
node:
  version: 12.8.4
  pkg:
    version: 12
    use_upstream_repo: True
```

markbreedlove:
OK. c6f73374 has the update.

blarghmatey:
:+1:

PR #1299, opened 2020-10-13
Change OCW Next CORS domain for Heroku configuration

markbreedlove:
Is this OK, or do we want to preserve the other alternative domains for the Next sites, since the domains are still hooked up? This only allows one.


blarghmatey:
Yeah, we're going to want to at least keep ocw-beta working on production, unless we decide to force a redirect to the new domain. I defer to @pdpinch on that decision.

pdpinch:
I'm hoping that we will soon redirect ocw-beta to ocwnext, but it's best to stay flexible for now.

markbreedlove:
@blarghmatey Do you think commit a56468b1 is OK? This would allow some flexibility.

blarghmatey:
Yes, this will work. :+1:

PR #1300, opened 2020-10-13
Fix OCW Next webhook script logging if lock error

markbreedlove:
Fix the OCW Next build server's webhook script so that it does not remove the lock directory if it's failing because a lock directory already exists!


blarghmatey:
:+1:

PR #1301, opened 2020-10-13
Add retry capability to OCW next webhook script

markbreedlove:
Add some simple "retry" functionality to the Caddy webhook script.

Have it run itself again if it looks like another invocation came along in the middle of the current run. This could mean that there's another code commit that needs to make it to the server.

Also replace some tab characters with space-character sequences.


blarghmatey:
:+1:

PR #1303, opened 2020-10-14
Add Fastly cache purge to OCW Next webhook script

markbreedlove:
Attempt to purge the Fastly cache after pushing content to the web bucket.

### DIFF
--- a/pillar/apps/ocw-next-production.sls
+++ b/pillar/apps/ocw-next-production.sls
@@ -5,8 +5,8 @@ ocw-next:
   search_api_url: //open.mit.edu/api/v0/search/
   ocw_to_hugo_git_ref: release
   hugo_course_publisher_git_ref: release
-  fastly_api_token: __vault__::secret-open-courseware/production-apps/fastly-api-token>data>value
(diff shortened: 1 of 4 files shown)


pillar/apps/ocw-next-production.sls
```diff
@@ -5,6 +5,8 @@ ocw-next:
   search_api_url: //open.mit.edu/api/v0/search/
   ocw_to_hugo_git_ref: release
   hugo_course_publisher_git_ref: release
+  fastly_api_token: __vault__::secret-open-courseware/production-apps/fastly-api-token>data>value
+  fastly_service_id: __vault__::secret-open-courseware/production-apps/fastly-service-id>data>value
```
blarghmatey, line 9: I would combine these two values in one path, so that it would be something like:
```
  fastly_api_token: __vault__::secret-open-courseware/production-apps/fastly-api>data>token
  fastly_service_id: __vault__::secret-open-courseware/production-apps/fastly-api>data>service_id
```

markbreedlove: I've changed those paths as indicated, and have created the corresponding Vault secrets.

blarghmatey:
:+1:

PR #1305, opened 2020-10-20
Fix subnet names after having them renamed

shaidar:
#### What's this PR do?
Fixed subnet names after having them renamed through Pulumi


blarghmatey:
:+1:

PR #1306, opened 2020-10-21
Changed xPro VPC name to match new naming

shaidar:
#### What's this PR do?
xPro VPC names have changed since the import into Pulumi, so we just needed to reflect the change in the env_settings file.


blarghmatey:
:+1:

PR #1307, opened 2020-10-21
Manage and run ocw-to-hugo through hugo-course-publisher NPM target

markbreedlove:
This revises the webhook script so that it uses `hugo-course-publisher`'s `import:ocw` NPM target to pull a `ocw-to-hugo` from NPM and pull down the source data, instead of us doing those things ourselves.

I've just run it in successfully QA by manually pasting this in and updating the environment variables.

### DIFF
--- a/salt/apps/ocw/templates/webhook-publish.sh.jinja
+++ b/salt/apps/ocw/templates/webhook-publish.sh.jinja
@@ -1,9 +1,6 @@
 #!/usr/bin/env bash
 
 LOG_FILE=/opt/ocw/webhook-publish.log
-SOURCE_DATA_BUCKET={{ source_data_bucket }}
-SOURCE_DATA_DIR=/opt/ocw/open-learning-course-data
-MARKDOWN_DIR=/opt/ocw/hugo-course-publisher/site/content
 SITE_OUTPUT_DIR=/opt/ocw/hugo-course-publisher/dist/  # Should end in '/'
 WEBSITE_BUCKET={{ website_bucket }}
 FASTLY_API_TOKEN={{ fastly_api_token }}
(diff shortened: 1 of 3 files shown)


blarghmatey:
:+1:

salt/apps/ocw/templates/webhook-publish.sh.jinja
```diff
@@ -14,6 +11,16 @@ lock_dir=/tmp/webhook-publish-lock
 # that came in during the period of the current run.
 retry_file=/tmp/webhook-publish-retry
 
+# Configuration for hugo-course-publisher
+OCW_TO_HUGO_INPUT=/opt/ocw/open-learning-course-data
+OCW_TO_HUGO_DOWNLOAD=1
```
gumaerc, line 16: You don't need to set this flag I think, as `ocw-to-hugo` is not handling the course download and you are running `aws sync` to sync `open-learning-course-data-production` right?

gumaerc (changes requested):
just one issue with your env variables

salt/apps/ocw/templates/webhook-publish.sh.jinja
gumaerc: Although this made me think, since we're going to need to support updating only a single / a few courses in the future, we could use this functionality to accomplish that.  I'll need to add support for not blowing away the built site every time a build runs, but in theory we could use the download functionality already built into `ocw-to-hugo` to just only download a subset of updated courses, process them into markdown and build replacement course sites.

markbreedlove: I removed that `aws sync`, which used to be on lines 54 through 58, in order to simplify things. It's nice that that could make publishes of subsets feasible in the future.

gumaerc: Ah, I didn't look closely enough and didn't see that.  Carry on!

gumaerc (approved):
:+1:

salt/apps/ocw/templates/webhook-publish.sh.jinja
```diff
@@ -40,33 +47,21 @@ else
...
-aws s3 sync s3://$SOURCE_DATA_BUCKET/ $SOURCE_DATA_DIR --delete --only-show-errors
-if [ $? -ne 0 ]; then
-    error_and_exit "Failed to pull source data"
-fi
-
-log_message "Running ocw-to-hugo"
-cd /opt/ocw/ocw-to-hugo
-node src/bin/index.js -i $SOURCE_DATA_DIR -o $MARKDOWN_DIR \
-    --strips3 --staticPrefix /coursemedia
+log_message "Pulling ocw-to-hugo and source data"
```
gumaerc, line 58: A little bit of a nitpick, but technically `ocw-to-hugo` is pulled in during the `yarn install` command above.  This command just executes it based on the environment variables.

PR #1309, opened 2020-10-30
Updated configs to point to new certs

shaidar:
#### What's this PR do?
Updates configs to point to new wildcard certs.


blarghmatey:
:+1:

PR #1310, opened 2020-11-05
Tune reddit memory usage beacon

markbreedlove:
Tune the beacon that is used for the reactor that restarts Reddit processes when the used memory gets to a certain  percentage. This is being tuned to align with our updated Datadog monitor.


blarghmatey:
:+1:

PR #1312, opened 2020-11-09
Koa related changes for xpro sandbox

shaidar:
#### What's this PR do?
A few changes related to Koa and it running on Ubuntu 20. Also cleaned up some python2 references as we're no longer running that on any of our edx instances.


blarghmatey:
:+1:

PR #1313, opened 2020-11-10
Add enhanced Sentry support for edX apps

markbreedlove:
Add enhanced Sentry support for the MITx and xPro edX applications, with the `EOX core` plugin.

See https://github.com/eduNEXT/eox-core

Is this redundant with the existing `raven` dependency?


blarghmatey:
Yes, we can remove `raven` from the dependencies list in favor of this plugin.

markbreedlove:
OK, just removed it.

blarghmatey:
:+1:

PR #1314, opened 2020-11-16
Changes related to Koa deployment

shaidar:
#### What's this PR do?
- Added comments to keys/values that can be eventually removed once all instances are on Koa
- Added conditional for edxapp private requirements as those have changes in Koa


blarghmatey:
:+1:

PR #1316, opened 2020-11-19
Added koa pillar to xpro-qa

shaidar:
#### What's this PR do?
Adds koa pillar state file to mitxpro-qa.

### DIFF
--- a/pillar/top.sls
+++ b/pillar/top.sls
@@ -236,9 +236,9 @@ base:
     - edx
     - edx.sandbox
     - edx.ansible_vars
-  'G@roles:sandbox and P@environment:mitxpro-qa':
-    - match: compound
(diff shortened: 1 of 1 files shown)


pillar/top.sls
```diff
@@ -236,9 +236,9 @@ base:
     - edx
     - edx.sandbox
     - edx.ansible_vars
-  'G@roles:sandbox and P@environment:mitxpro-qa':
-    - match: compound
     - edx.ansible_vars.theme
+  'G@edx_codename:koa':
+    - match: compound
```
blarghmatey, line 241: We should place this further down the list as the pillar merging is last one wins. If it is placed above other ansible vars pillars then it is likely to get overwritten.

blarghmatey:
:+1:

PR #1317, opened 2020-11-23
Modified python path to absolute one

shaidar:
#### What's this PR do?
In Koa, the `/edx/bin/python.edxapp` symlink no longer exists. This PR uses the absolute python path and should work for Koa and previous releases.

As a side note, we also might want to consider using the migrations scripts that exists in `/edx/bin`


blarghmatey:
:+1:

PR #1320, opened 2020-11-24
Added option to run restore scripts

shaidar:
#### What's this PR do?
Gives us the option to specify whether we want to run the restore scripts as part of the restore instance build or not


blarghmatey:
:+1:

PR #1321, opened 2020-11-30
Update OCW Next webhook script to recognize code changes

markbreedlove:
Update the OCW Next webhook build script to recognize when a git pull has resulted in a change, and to only publish if necessary.

There is a "full" option now to tell the script to run even without a code change, in case that is necessary, but the default behavior is now for it to stop if there has been no new commit.

### DIFF
--- a/salt/apps/ocw/templates/webhook-publish.sh.jinja
+++ b/salt/apps/ocw/templates/webhook-publish.sh.jinja
@@ -45,14 +45,25 @@ else
     rm $retry_file
 fi
 
+cd /opt/ocw/hugo-course-publisher \
+    || error_and_exit "Can not cd to hugo-course-publisher!"
+
+orig_commit=`git rev-parse HEAD`
+log_message "Pulling hugo-course-publisher"
+git pull || error_and_exit "Can not pull hugo-course-publisher"
+new_commit=`git rev-parse HEAD`
+log_message "$orig_commit -> $new_commit"
+if [ "x$orig_commit" != "x$new_commit" ]; then
+    log_message "Pulled new commit, so continuing with publish."
+else
+    error_and_exit "No new commit or got a rev-parse error, so stopping here."
+fi
+
 log_message "Pulling source data"
 aws s3 sync s3://$SOURCE_DATA_BUCKET/ $OCW_TO_HUGO_INPUT/ --delete --only-show-errors
 if [ $? -ne 0 ]; then
     error_and_exit "Failed to pull source data"
 fi
-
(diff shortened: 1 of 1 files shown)


salt/apps/ocw/templates/webhook-publish.sh.jinja
```diff
@@ -45,24 +60,37 @@ else
...
+    || error_and_exit "Can not cd to hugo-course-publisher!"
+
+orig_commit=`git rev-parse HEAD`
+log_message "Pulling hugo-course-publisher"
+git pull || error_and_exit "Can not pull hugo-course-publisher"
+new_commit=`git rev-parse HEAD`
+log_message "$orig_commit -> $new_commit"
+
+if [ $option == "full" ]; then
+    if [ "$orig_commit" != "$new_commit" ]; then
```
blarghmatey, line 73: Shouldn't these conditionals be swapped?

salt/apps/ocw/templates/webhook-publish.sh.jinja
```diff
@@ -45,24 +60,37 @@ else
...
+        log_message "Pulled new commit, so continuing with publish."
+    else
+        error_and_exit "No new commit or got a rev-parse error, so stopping here."
```
markbreedlove, line 76: 🤔 On second thought, maybe this should not exit with an error code.

salt/apps/ocw/templates/webhook-publish.sh.jinja
markbreedlove: The logic is for it to not bother checking if it's run with the "full" option. The only good stuff inside this block is where it exits the script.

markbreedlove: Oops! It should be `!=`, not `==` ...

blarghmatey: Then shouldn't it be `$option != "full"`?

markbreedlove: Yeah, just changed it in commit c803bb41

salt/apps/ocw/templates/webhook-publish.sh.jinja
blarghmatey: I agree that it should just `exit 0` in the event of an unchanged commit.

markbreedlove: Changed in commit a6dd1c76

blarghmatey:
:+1:

PR #1322, opened 2020-12-02
update OCW build to support new data template structure

gumaerc:
Related to https://github.com/mitodl/hugo-course-publisher/issues/326

#### What's this PR do?
This PR updates the OCW build scripts to output to the proper path after the recent changes to store course metadata using Hugo Data Templates.  It also now clears out the `site/data/courses` folder before a fresh build in the same way it clears out `site/content`

Not certain how code is normally tested here.

### DIFF
--- a/salt/apps/ocw/nextgen_build_publish.sls
+++ b/salt/apps/ocw/nextgen_build_publish.sls
@@ -4,7 +4,7 @@ run_ocw_to_hugo:
   cmd.run:
     # Installing ocw-to-hugo without the `-g' switch leaves us without an
     # `ocw-to-hugo' executable, so we use the path to `index.js'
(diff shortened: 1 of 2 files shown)


gumaerc:
I'm realizing now that the change to that `nextgen_build_publish.sls` script might not be necessary because `ocw-to-hugo` is now being run through `npm run import:ocw` and the output path is set within the script that it runs.  Maybe `run_ocw_to_hugo` doesn't need to be there at all?  Either way, the change that clears out `site/data/courses/*` is necessary.

markbreedlove:
You are right; I've been meaning to remove that nextgen_build_publish Salt state. You can back out the change to that particular file, if you like, because I'm just going to issue a separate commit soon to remove the file.

gumaerc:
@markbreedlove I backed out that change to the salt state file, all set to merge pending your approval.

markbreedlove (approved):
All's well.

markbreedlove:
I have deployed this to QA and production.

PR #1323, opened 2020-12-02
Change MITx proctoring provider to "null"

markbreedlove:
Change the proctoring provider in the MITx Studio advanced settings. This is causing an error in MITx Studio.

Fixes https://github.mit.edu/mitx/Spring-2021/issues/1


blarghmatey:
:+1:

PR #1324, opened 2020-12-03
Remove old OCW Next states that we don't need

markbreedlove:
The work performed by these states is better left to the webhook-publish.sh script.


blarghmatey:
:+1:

PR #1325, opened 2020-12-04
Fix secgroups based on Pulumi changes

shaidar:
#### What's this PR do?
We changed how some of the main secgroups are named through Pulumi. This PR addresses that change for the deploy state


salt/orchestrate/apps/deploy.sls
```diff
@@ -1,5 +1,5 @@
 {% set env_settings = salt.cp.get_url("https://raw.githubusercontent.com/mitodl/salt-ops/main/salt/environment_settings.yml", dest=None)|load_yaml %}
-{% set ENVIRONMENT = salt.environ.get('ENVIRONMENT', 'rc-apps') %}
+{% set ENVIRONMENT = salt.environ.get('ENVIRONMENT', 'applications-qa') %}
```
blarghmatey, line 2: Pulumi treats the env name as applications-qa, but there are still too many assumptions built into the pillar data for us to switch to that naming in salt land.

blarghmatey (changes requested):

shaidar:
Too early for this PR

PR #1328, opened 2020-12-10
Enabled volume encryption on edx instances

shaidar:
#### What's this PR do?
Change enables EC2 EBS volume encryption using the default AWS managed KMS key on Open edX instances.


blarghmatey:
:+1:

PR #1332, opened 2020-12-22
Updated mount point and DNS for edX MySQL DB for xPro RDS update

blarghmatey:
In order to allow for upgrading the DB without disrupting the running system we changed the DNS entry and mount point for the new RDS instance during the xPro upgrade to enable storage encryption.


shaidar:
👍

PR #1333, opened 2020-12-22
Updated mitxpro_autoscaling secgroups

shaidar:
#### What's this PR do?
Updated asg state to use new secgroups generated by Pulumi and the QA AMI

### DIFF
--- a/salt/orchestrate/aws/mitxpro_autoscaling.sls
+++ b/salt/orchestrate/aws/mitxpro_autoscaling.sls
@@ -1,6 +1,10 @@
 {% for app_name in ['edxapp', 'edx-worker'] %}
 {% set env_settings = salt.cp.get_url("https://raw.githubusercontent.com/mitodl/salt-ops/main/salt/environment_settings.yml", dest=None)|load_yaml %}
 {% set ENVIRONMENT = salt.environ.get('ENVIRONMENT', 'mitxpro-production') %}
+
+# Using QA AMI build to deploy production instances 
+{% set AMI_ENV = salt.environ.get('AMI_ENV', 'mitxpro-qa') %}
+
 {% set purpose = salt.grains.get('purpose', 'xpro-production') %}
 {% set env_data = env_settings.environments[ENVIRONMENT] %}
 {% set VPC_NAME = env_data.vpc_name %}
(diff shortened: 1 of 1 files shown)


salt/orchestrate/aws/mitxpro_autoscaling.sls
```diff
@@ -9,18 +13,14 @@
...
 {% set sns_topic = env_data.provider_services[app_name].sns.topic ~ '-' ~ ENVIRONMENT ~ '-autoscaling' %}
 {% set edx_codename = purpose_data.versions.codename %}
-{% if app_name == 'edxapp' %}
-{% set security_groups = ['edx'] %}
-{% else %}
-{% set security_groups = ['edx-worker'] %}
-{% endif %}
-{% do security_groups.extend(['master-ssh', 'consul-agent', 'default']) %}
+{% set env_suffix = ENVIRONMENT.strip('-')[1] %}
+{% set security_groups = ['{}-salt-minion'.format(ENVIRONMENT), '{}-consul-agent'.format(ENVIRONMENT), 'default', 'mitxpro-edxapp-access-{}'.format(env_suffix), 'mitxpro-edxapp-db-access-{}'.format(env_suffix)]) %}
```
blarghmatey, line 17: We don't want to assign the db security group to the edx instances. Also, might be good to split this across multiple lines for readability.

salt/orchestrate/aws/mitxpro_autoscaling.sls
```diff
@@ -1,6 +1,10 @@
 {% for app_name in ['edxapp', 'edx-worker'] %}
 {% set env_settings = salt.cp.get_url("https://raw.githubusercontent.com/mitodl/salt-ops/main/salt/environment_settings.yml", dest=None)|load_yaml %}
 {% set ENVIRONMENT = salt.environ.get('ENVIRONMENT', 'mitxpro-production') %}
+
+# Using QA AMI build to deploy production instances 
+{% set AMI_ENV = salt.environ.get('AMI_ENV', 'mitxpro-qa') %}
```
blarghmatey, line 6: Don't forget that we need to manually set the AMI ID in the SDB in production.

shaidar: Why does that matter since we're passing the AMI to the launch configuration

blarghmatey: Because it's trying to pull the ID from SDB, and the SDB systems in production and QA can't talk to each other, so we need to update the value in the production SDB for it to get fetched properly.

shaidar: The only place I see it using the sdb is for the `release_number` and that appears useless as it's not used anywhere in that file. As far as the AMI, I only see it using the AMI name and uses that to create the launch config. What am I missing?

shaidar: Nevermind, I got it

blarghmatey: Nevermind, I was thinking about how it's used in the residential orchestrate, didn't realize it was doing a lookup of the actual AMI.

blarghmatey:
:+1:

PR #1340, opened 2021-01-13
Update xPro redirects for learner certificates

markbreedlove:
This fixes a redirect that we used to return for xPro learners' course certificates, which sent them to edx.org. We're now sending them to our certificates.mitxpro.mit.edu site. Additionally, it adds a redirect for xPro program certificates; again, sending them to the new certificates domain. The Cloudfront distribution for certificates.mitxpro.mit.edu has been updated with a Behavior to send requests for `/credentials` to our redirect server so that it can issue the redirect, because redirects are a pain to configure in S3, if not impossible for this particular pattern-matching use-case.

This is already set on the server. The changes here document what's already been modified.


blarghmatey:
:+1:

PR #1344, opened 2021-01-15
Make OCW Next build script upload ocw-to-hugo's output

markbreedlove:
This adds an `aws S3 sync` of the Markdown and JSON files that come out of `ocw-to-hugo`. They are pushed up to an S3 bucket that is to be consumed by `ocw-studio`.

### DIFF
--- a/pillar/apps/ocw-next-production.sls
+++ b/pillar/apps/ocw-next-production.sls
@@ -1,6 +1,7 @@
 
 ocw-next:
   website_bucket: ocw-website-applications-production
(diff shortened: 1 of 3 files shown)


blarghmatey:
:+1:

blarghmatey:
For the QA instance can you also have it upload to the CI bucket?

markbreedlove:
I can probably do that, but would it be possible instead for the CI ocw-studio to just pull from the RC bucket?

blarghmatey:
Yeah, that's fine

PR #1345, opened 2021-01-15
Updating to use QA bucket creds for CI environment for the time being

blarghmatey:


markbreedlove:
That looks OK. The corresponding Vault credentials contain an IAM user that has access to the right bucket.

shaidar:
👍

PR #1348, opened 2021-01-20
Removing some leftover Koa conditionals

blarghmatey:
- Explicitly setting the ES7 host for Forum to fix failing check during build
- Removing conditionals around Koa since it's running everywhere now


shaidar:
👍

PR #1349, opened 2021-01-21
Write out commit hash files in OCW Next build

markbreedlove:
In webhook-publish.sh (the webhook build script for OCW Next), write out files with commit hashes of the associated repositories so that our build bot can check them.


blarghmatey:
:+1:

PR #1350, opened 2021-01-22
Added NovoEd feature flag to bootcamp RC/prod

gsidebo:
N/A

#### What's this PR do?
Sets the NovoEd integration flag to `True` in bootcamps RC and prod

Leaving the flag as `False` in the CI config is intentional. NovoEd integration is only configured to work in RC and prod


blarghmatey:
:+1:

PR #1351, opened 2021-01-22
Change ocw next webhook logging

markbreedlove:
Improve the OCW Next webhook build script's logging, following reports of issues where the logs had been truncated and we lacked diagnostic information.

The logs are now put in their own `logs` directory, and are rotated with logrotate. Not truncating them also allows us to consume them more reliably with a logging agent like Vector or FluentD.


blarghmatey:
:+1:

PR #1352, opened 2021-01-25
Initial cleanup of hacks.sls

shaidar:
#### What's this PR do?
Removed the following hacks since they are no longer needed or have been added to `ansible_vars`:
- EMAIL_USE_DEFAULT_FROM_FOR_BULK -> added to `ansible_vars`
- RAVEN_CONFIG -> replaced through the use of `eox-core[sentry]`
- JWT config for xpro -> handled through `ansible_vars`
- Removed XPRO_BASE_URL -> added config to edx-platform and configure it through ansible_vars

### DIFF
--- a/salt/edx/hacks.sls
+++ b/salt/edx/hacks.sls
@@ -1,10 +1,4 @@
 {% set purpose = salt.grains.get('purpose', 'xpro-qa') %}
-{% set heroku_xpro_env_url_mapping = {
-    'sandbox': 'https://xpro-ci.odl.mit.edu',
-    'xpro-qa': 'https://xpro-rc.odl.mit.edu',
-    'xpro-production': 'https://xpro.mit.edu'
-  } %}
-{% set heroku_env = heroku_xpro_env_url_mapping['{}'.format(purpose)] %}
 
 {% if salt.file.directory_exists('/edx/var/edxapp/staticfiles/studio/templates') %}
 ensure_license_selector_template_is_in_expected_location:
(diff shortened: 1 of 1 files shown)


salt/edx/hacks.sls
```diff
@@ -71,21 +52,5 @@ add_xpro_base_url_to_{{ app }}_production_file:
     - text: XPRO_BASE_URL = '{{ heroku_env }}'
```
blarghmatey, line 52: We should commit this directly to edx-platform and maintain it as a cherry pick rather than appending to the file at deploy time. We'll want to add it to production.py so that it reads the value in from lms.yaml and then add that value in the lms_env_extra in the ansible vars.

shaidar: Does it make sense though to have it be in the edx-platform when it's specific to xpro?

blarghmatey: It would be on the xpro branch

shaidar:
@blarghmatey Think this should be enough for now as the static assets would probably need some additional testing/time to verify whether they're still needed or not.

salt/edx/hacks.sls
```diff
@@ -45,47 +33,10 @@ copy_select_static_assets_to_static_subfolder:
...
-  file.append:
-    - name: /edx/app/edxapp/edx-platform/{{ app }}/envs/production.py
-    - text: RAVEN_CONFIG = ENV_TOKENS.get('RAVEN_CONFIG', {})
-{% endfor %}
-
 {% if 'mitxpro' in salt.grains.get('environment') %}
 add_social_auth_https_redirect_to_lms_production_file:
   file.append:
     - name: /edx/app/edxapp/edx-platform/lms/envs/production.py
     - text: SOCIAL_AUTH_REDIRECT_IS_HTTPS = ENV_TOKENS.get('SOCIAL_AUTH_REDIRECT_IS_HTTPS', True)
```
blarghmatey, line 40: We should add this directly to edx-platform along with the xpro_base_url rather than relying on appending to it at build time.

shaidar: This looks like it's defined/used in multiple places in the configuration repo and set as `false`

blarghmatey:
We can merge this as is for now, but should circle back on the `SOCIAL_AUTH_REDIRECT_IS_HTTPS` setting.

PR #1353, opened 2021-01-26
Add Vector agent to OCW Next build server

markbreedlove:
Add the Vector log agent to the OCW Next build server and have it ship directly to Elasticsearch.

This is a trial and if it goes well we will add aggregators and more agents.

Though I haven't been able to test the Salt states, I've been able to test the configuration file itself with a couple of VirtualBox VMs that simulate an agent and the Elasticsearch service. I've also tested the systemd service configuration.

### DIFF
--- a/pillar/vector/init.sls
+++ b/pillar/vector/init.sls
@@ -1,0 +1,2 @@
+vector:
+  configuration:
--- a/pillar/vector/init.sls
+++ b/pillar/vector/init.sls
@@ -1,2 +1,0 @@
-vector:
-  configuration:
--- a/salt/vector/install.sls
+++ b/salt/vector/install.sls
@@ -28,9 +28,7 @@ ensure_state_of_systemd_service_file:
   file.managed:
     - name: /etc/systemd/system/vector.service
     - source: salt://vector/files/vector.service
-  {% if salt.grains.get('init') == 'systemd' %}
   cmd.wait:
     - name: systemctl daemon-reload
     - watch:
         - file: ensure_state_of_systemd_service_file
-  {% endif %}
--- a/salt/vector/install.sls
+++ b/salt/vector/install.sls
@@ -1,6 +1,21 @@
-install_vector_repo:
+ensure_package_prerequisite_installations:
+  pkg.installed:
+    pkgs:
+      - debian-keyring
+      - debian-archive-keyring
+      - apt-transport-https
+      - ca-certificates
+      - gnupg
+
+install_vector_repo_key:
   cmd.run:
-    - name: curl -1sLf 'https://repositories.timber.io/public/vector/cfg/setup/bash.deb.sh' | bash
+    - name: curl -1sLf "https://repositories.timber.io/public/vector/cfg/gpg/gpg.3543DB2D0A2BC4B8.key" | apt-key add -
+
+update_apt_sources_list:
+  pkgrepo.managed:
+    - humanname: Vector
+    - name: deb https://repositories.timber.io/public/vector/deb/debian {{ salt.grains.get('oscodename', 'stable') }} main
+    - refresh_db: True
 
 ensure_vector_package_state:
   pkg.installed:
--- a/salt/vector/install.sls
+++ b/salt/vector/install.sls
@@ -13,3 +13,9 @@ ensure_state_of_systemd_service_file:
   file.managed:
     - name: /etc/systemd/system/vector.service
     - source: salt://vector/files/vector.service
+  {% if salt.grains.get('init') == 'systemd' %}
+  cmd.wait:
+    - name: systemctl daemon-reload
+    - watch:
+        - file: ensure_state_of_systemd_service_file
+  {% endif %}
--- a/salt/vector/install.sls
+++ b/salt/vector/install.sls
@@ -30,5 +30,5 @@ ensure_state_of_systemd_service_file:
     - source: salt://vector/files/vector.service
   cmd.wait:
     - name: systemctl daemon-reload
-    - watch:
+    - onchanges:
         - file: ensure_state_of_systemd_service_file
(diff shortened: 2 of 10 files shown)


markbreedlove:
We currently have a `fluentd_tag` field to identify a category for each log message. If we move away from FluentD, this name is going to become obsolete. Do you mind if I start using a new field named "category?" I considered "tag," but "tag" means something specific in Vector related to metrics (as opposed to messages), so I want to avoid any confusion with that.

pillar/vector/init.sls
```diff
@@ -0,0 +1,2 @@
+vector:
+  configuration:
```
blarghmatey, line 2: There's not really a need for an `init.sls` file if it's not doing anything. The `init.sls` is just for providing a default baseline that gets merged into specific configs for different instances.

salt/vector/install.sls
```diff
@@ -0,0 +1,15 @@
+install_vector_repo:
+  cmd.run:
+    - name: curl -1sLf 'https://repositories.timber.io/public/vector/cfg/setup/bash.deb.sh' | bash
```
blarghmatey, line 3: Rather than running a curl-bash we should either set up the package repository, or just download the compiled binary, especially since we're creating the systemd unit anyway.

Digging into the script the Apt repository is at
```
deb https://repositories.timber.io/public/vector/deb/debian buster main

deb-src https://repositories.timber.io/public/vector/deb/debian buster main
```

https://vector.dev/docs/setup/installation/manual/from-archives/

salt/vector/install.sls
```diff
@@ -0,0 +1,15 @@
...
+  pkg.installed:
+    - name: vector
+    - refresh: True
+    - require:
+      - cmd: install_vector_repo
+
+ensure_state_of_systemd_service_file:
+  file.managed:
+    - name: /etc/systemd/system/vector.service
+    - source: salt://vector/files/vector.service
```
blarghmatey, line 15: When creating a systemd unit file you have to also run `systemctl daemon-reload` for systemd to pick up the new service definition.

blarghmatey:
To the question of fluentd_tag vs tag vs categories etc. I think that category is a bit misleading and we might be better served with calling it `labels`. As an additional question, does vector support applying either multiple or nested tags/labels?

pillar/vector/init.sls
markbreedlove: ok. I'll remove this.

salt/vector/install.sls
markbreedlove: I wondered about that because I've done so elsewhere, but did not look into the script. It's not doing anything special outside of setting the apt repository like we usually do?

markbreedlove: FWIW, Vector's installation documentation says to run the script to install it; but I guess they can say whatever they want to make it look easier.

salt/vector/install.sls
markbreedlove: Thanks, will add that

markbreedlove:
Re labels, categories, tags, etc:

The `category` is just a field that I'm adding to a record: a property named `category` which is a property of the JSON like any other, so it could be an array; so I guess we could start using multiple tokens of some sort in place of what we currently call `fluentd_tag`.

But a "tag" in Vector is a specific thing related to metrics, and I have not read their metrics documentation. I had come across the term in reading some introductory text, with the explanation that it was something related to metrics specifically.

Outside of that, I don't recall there being anything specific about labels or whatever. You just add whatever custom properties you want to the log event.

salt/vector/install.sls
blarghmatey: Yeah, the script just sets up the apt repo, nothing much beyond that.

markbreedlove:
I've changed `category` to `labels`, and it works as an array.

I performed a test that gives me the following with an additional label applied:

```
    "labels": [
      "ocw_build",
      "testlabel"
    ],
```

pillar/vector/init.sls
markbreedlove: Removed in 34f969e1

salt/vector/install.sls
markbreedlove: Do you think that d7635657 will work?

salt/vector/install.sls
```diff
@@ -13,3 +13,9 @@ ensure_state_of_systemd_service_file:
   file.managed:
     - name: /etc/systemd/system/vector.service
     - source: salt://vector/files/vector.service
+  {% if salt.grains.get('init') == 'systemd' %}
```
blarghmatey, line 16: There's not really a need for this check. We're already assuming systemd by the contents of the unit file.

salt/vector/install.sls
```diff
@@ -13,3 +13,9 @@ ensure_state_of_systemd_service_file:
...
+  cmd.wait:
+    - name: systemctl daemon-reload
+    - watch:
```
blarghmatey, line 19: You can change this to `onchanges`

salt/vector/install.sls
markbreedlove: OK. I'd copied the logic from the cassandra state.

salt/vector/install.sls
markbreedlove: I've tried to handle this in 60883a26

The script customizes the apt sources list based on the particular version of Debian and the architecture.

salt/vector/install.sls
markbreedlove: I think I was emulating another state; probably `salt/edx/gitreload.sls`. I'll check the usage for `onchanges` ...

markbreedlove: Changed in 7b3d921d

blarghmatey:
:+1:

PR #1359, opened 2021-02-04
Removed Alcali

shaidar:
#### What's this PR do?
We didn't really end up using Alcali and this PR removes all references to it.


blarghmatey:
:+1:

PR #1360, opened 2021-02-06
Removed data http-check-integration

shaidar:
#### What's this PR do?
This is no longer needed now that we're using Pingdom for http checks.


blarghmatey:
:+1:

PR #1362, opened 2021-02-10
Adding xpro app domains to login whitelist for xPro edX

blarghmatey:


pillar/edx/ansible_vars/xpro.sls
```diff
@@ -5,18 +5,21 @@
...
 {% set support_email = 'support@xpro.mit.edu' %}
-{% set heroku_xpro_env_url_mapping = {
+{% set heroku_xpro_env_domain_mapping = {
     'sandbox': 'https://xpro-ci.odl.mit.edu',
-    'xpro-qa': 'https://xpro-rc.odl.mit.edu',
+    'xpro-qa': 'https://rc.xpro.mit.edu',
     'xpro-production': 'https://xpro.mit.edu'
   } %}
-{% set heroku_env = heroku_xpro_env_url_mapping['{}'.format(purpose)] %}
+{% set heroku_env = heroku_xpro_env_domain_mapping['https://{}'.format(purpose)] %}
```
shaidar, line 13: Wouldn't `purpose` get the value for the key in `heroku_xpro_env_domain_mapping` which in the default case would be xpro-qa and that would return `https://rc.xpro.mit.edu` and you'd end up with double `https`?

pillar/edx/ansible_vars/xpro.sls
```diff
@@ -5,18 +5,21 @@
...
 {% set purpose_data = env_data.purposes[purpose] %}
 {% set LMS_DOMAIN = purpose_data.domains.lms %}
 {% set CMS_DOMAIN = purpose_data.domains.cms %}
+{% set edxapp_domains = purpose_data.domains.values()|list %}
```
shaidar, line 17: Doesn't this end up being a list where some domains have the `https` scheme and the ones pulled from env_settings not having it?

shaidar:
👍

PR #1365, opened 2021-02-11
Adding config to prevent abusive scans for PHP vulnerabilities

blarghmatey:
We have been getting Sentry errors due to people scanning for wordpress and PHP endpoints on our edX servers. This adds nginx rules to return a 404 for any wp-content or wp-admin routes and anything ending in .php


shaidar:
👍

PR #1367, opened 2021-02-26
Add vector agent for MITx residential

markbreedlove:
This swaps out the FluentD service for Vector, on the MITx residential servers.

There are a couple of things that you can provide feedback on even if you don't know Vector's configuration very well:

1)
I am changing the "tags," which we are now calling "labels" because Vector uses the term "tag" to mean something specific to metrics. The labels are now named like `edx_nginx_access` instead of `edx.nginx.access`. I hope that's not too much of an issue. There are no saved Kibana searches that use "tags" so I hope this isn't disruptive.

2)
I've gone back and forth on this, but I am changing index names. It doesn't make sense anymore to have them named `logstash-*` so I'm removing that prefix. It also doesn't seem necessary to throw in the whole environment name, like `logstash-mitx-production-*` because we never have production and QA in the same Elasticsearch logging cluster. The indices are broken up a little differently. There are separate indices for Nginx access and error logs because their fields are so different and I think fewer fields per index is a good thing. The "tracking" log gets its own index, as well, because of its different field content. The "edx app" stderr logs can still go into one index because their fields are mostly the same, even when their line formats vary. If we really need searches or graphs that work across old and new indices (a broad historical range) then this could get in the way; but if we can tolerate separate queries for the older and newer data ranges, then this should work. Eventually the old indices will be expired and this won't matter.

A couple of log files have ceased to exist, presumably with the edX upgrade to Koa. I've checked all of the logfiles represented here against reality, on both QA and Production, and I've verified all of the regexes with https://pythex.org and sample lines from production log files.


markbreedlove:
Seems like a pattern, but I have thought of something right after hitting the submit button. If there is any value in searching across _all_ log indices with one query -- maybe to find some needle in the haystack when you don't know which part of it to look in? -- then maybe I should prefix all of the index names with something so that you can use an Index Pattern like our existing `logstash-*`. I have seldom found that Index Pattern useful, but it's a thought. There are mapping conflicts between a lot of existing Elasticsearch indices, making that Index Pattern problematic (for example, there could be multiple logs with a "status" field, with different meanings or datatypes); but I thought I'd ask anyway.

markbreedlove:
Also note that the FluentD configuration was using multiline matches to catch random stderr output and stacktraces, but I've stopped doing that, because it's tricky, and also because I don't think it used to work properly in the first place.

markbreedlove:
I should probably have this remove `pillar/fluentd/mitx.sls` ...

blarghmatey:
This looks good to me, though I'm curious as to the motivation for removing the `logstash-` prefix from the index names. Is there any technical reason, or just for aesthetic purposes? If it's just aesthetic and causes some measure of switching cost then it might be worth just keeping it.

On the note of removing the `-qa`/`-production` suffixes from the index names, I would argue that it's wasted effort as it's easier to just pass through the exact environment value that the instances are configured with rather than editorializing.

markbreedlove:
The reason is not as much about aesthetics as efficiency: it just takes up less space and it's less misleading. But, on the other hand, it could be helpful to have some kind of prefixes to distinguish from the non-logging indices on the cluster, like the administrative ones.

My motivation in removing `-qa` and `-production` was to remove something unnecessary so that we don't have to do any string interpolation at all.

pillar/vector/mitx.sls
```diff
@@ -0,0 +1,371 @@
...
+          - /edx/var/log/supervisor/cms_*stderr.log
+
+      worker_lms_stderr_log:
+        type: file
+        include:
+          - /edx/var/log/supervisor/lms_*stderr.log
+
+      {% endif %}
+
+      tracking_log:
```
shaidar, line 62: I think this should go under the `edx` role

pillar/vector/mitx.sls
```diff
@@ -0,0 +1,371 @@
...
+
+      # edX worker's "cms" supervisor stderr log,
+      # e.g. /edx/var/log/supervisor/cms_default_4-stderr.log
+      # is formatted as in these examples:
+      #
+      # [2021-02-25 16:29:59,191: ERROR/ForkPoolWorker-4] git_auto_export.tasks.async_export_to_git[b6320d3b-d208-4ea4-ae53-4d0fbedc247c]: Failed async course content export to git (course id: course-v1:MITx+8.011r_8+2021_Spring): Unable to push changes.  This is usually because the remote repository cannot be contacted
+      # [2021-02-25 18:05:02,524: INFO/MainProcess] Received task: lms.djangoapps.discussion.tasks.update_discussions_map[9413651f-6863-4f62-bc61-85f565515568]  ETA:[2021-02-25 18:05:32.505810+00:00]
+      # [2021-02-25 16:31:14,432: ERROR/ForkPoolWorker-4] Error running git push command: b"To github.mit.edu:MITx-Studio2LMS/content-mit-8011r_8-2021_Spring.git\n ! [rejected]  [...]
+      #
+      worker_cms_stderr_log_parser:
```
shaidar, line 242: nitpick: Should we call this edx_worker.. given that the role is `edx-worker`

pillar/vector/mitx.sls
```diff
@@ -0,0 +1,371 @@
...
+          {% if 'edx' in salt.grains.get('roles') %}
+          - cms_stderr_sampler
+          - lms_stderr_sampler
+          {% endif %}
+          {% if 'edx-worker' in salt.grains.get('roles') %}
+          - worker_stderr_timestamp_renamer
+          {% endif %}
+        type: elasticsearch
+        endpoint: 'http://operations-elasticsearch.query.consul:9200'
+        index: mitx-%Y.%W
```
shaidar, line 362: I feel that this index name doesn't match the naming convention used for the other logs. Might want to either have multiple indices or if we're gonna use one index for all those logs, maybe use some more descriptive for it `mitx-stderr...`

markbreedlove:
What if I name the new indices starting with `logs-*`? That would take away the reference to Logstash, would shorten the index names, and would retain a prefix to distinguish log-related indices from administrative or configuration-related ones; plus we can still have Index Patterns that work across both the pre- and post-Vector data.

pillar/vector/mitx.sls
markbreedlove: I had applied this to both the `edx` and `edx-worker` roles (i.e. outside of the conditionals) because a `find` command reveals that they exist on both. `find /edx/var/log/tracking -name "*.log" -mtime -1 -ls` shows this log on one of the production `edx-worker-mitx-production-residential-live` minions and on one of the `draft` instances. Do you think I should ignore them?

pillar/vector/mitx.sls
markbreedlove: I'd removed "edx_" from the beginning of all of the `source`s because in the context of this file we know that they're edx-related, and the IDs are unique within the file.

pillar/vector/mitx.sls
markbreedlove: I agree; adding `stderr` would make its nature more clear.

markbreedlove:
TODO: I've just realized that I skipped the gitreload log. I'm adding that ...

markbreedlove:
Status so far, following the feedback given above:

1. I've added back an index name prefix, `logs-*`. I don't think there will be any switching cost this way, with some modifications to existing Index Patterns. I've analyzed our current Index Patterns and current saved reports, and this be disruptive. We still retain a prefix that sets log indices apart from configuration and administrative ones.
2. I've added the gitreload logs, which I'd skipped.

Things I haven't changed yet:

1. I haven't prefixed the YAML properties under `sources` with `edx_`.
2. The index names still don't have "-production" or "-qa" like they used to because it saves one more `import` and one more string interpolation. The environments won't be mixed on a given Elasticsearch cluster, so I think this will be OK.

Feel free to push for changes to those last two items, if you like! I've just assumed that maybe the strategies I had been using will become more clear, but if you don't think they'll work well then I can make adjustments.

markbreedlove:
I'm going to go back on what I said above about multiline matches, because it looks like our logs have too much important information in unformatted stacktraces. I'm taking another look at the Vector docs and want to try something.

markbreedlove:
I've added some multiline matching to bring in stacktrace data from a number of our logs.

This Rust-style regex tester helps a lot: https://rustexp.lpil.uk/

I've tested most of the configuration directives with my VM setup.

blarghmatey:
:+1:

PR #1368, opened 2021-03-02
Adding config to force edX logouts to redirect to xPro app

blarghmatey:
Also fixes the path to the Sentry DSN for xPro

-------

https://github.com/mitodl/edx-platform/pull/241


shaidar:
👍

PR #1366, opened 2021-02-24
add mailchimp settings for newsletter on home page

pdpinch:
#### What are the relevant tickets?

fixes https://github.com/mitodl/ocw-www/issues/52

#### What's this PR do?

Adds Mailchimp settings, which I hope will be available to the ocw-www build. 

#### How should this be manually tested?

Visit the ocwnext home page and enter your email address in the newsletter form:

image

PR #1372, opened 2021-03-03
fix typo in mailchimp_user_id key

pdpinch:

none

#### What's this PR do?

spells "mailchimp" correctly

same as #1366


noisecapella:
This PR looks good but I'm wondering if there are other changes that need to be made. For example, `gtm_account_id` is a similar variable and there changes in the build script for it which I don't see for the mailchimp variable

pdpinch:
Moot, now that we went with https://github.com/mitodl/ocw-www/issues/62

PR #1373, opened 2021-03-03
Fix Vector installation for Ubuntu

markbreedlove:
Here is a quick PR to fix the Vector `install` state for an Ubuntu minion.

The package repository that I had hardcoded does not work for Ubuntu, with "debian" as the o/s and "focal" as the (nonexistent for Debian) release codename.

The `/etc/apt/sources.list.d/timber-vector.list` value in my Ubuntu VM is:
```
deb https://repositories.timber.io/public/vector/deb/ubuntu focal main
```

### DIFF
--- a/salt/vector/install.sls
+++ b/salt/vector/install.sls
@@ -1,11 +1,10 @@
+{% from "vector/map.jinja" import vector with context %}
+
+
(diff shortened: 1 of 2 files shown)


blarghmatey:
:+1: looks good

PR #1374, opened 2021-03-04
Add sink for MITx auth log data

markbreedlove:
I'd neglected to add a sink for the auth.log data in my last Vector PR.

This adds that sink and sets a new paradigm of adding all auth.log events to a single index for all of our servers. I figured this would make it easier to manage permissions on security data in the future.


blarghmatey:
:+1:

PR #1375, opened 2021-03-04
Add environment field to MITx Vector events

markbreedlove:
This one adds an `environment` field to events, and changes the name of some transforms because they are doing more now
than adding the "label" field.


blarghmatey:
:+1:

PR #1376, opened 2021-03-04
Swap Vector for FluentD on MongoDB minions

markbreedlove:
This adds the Vector agent to MongoDB minions, and changes/fixes the topfile selectors for both the `mongodb` and `edx` roles.

The main thing I could use input on is the index naming pattern: `logs-mongodb-%Y.%W`. As envisioned, this would hold MongoDB logs for all `mongodb` hosts, including XPro and MITx. Is this going to be OK, or do we want the ability to preserve MITx Residential's MongoDB logs for a different duration than XPro's? In that case, I'd want to name the indexes based on the environment name in addition to "mongodb."

### DIFF
--- a/pillar/top.sls
+++ b/pillar/top.sls
@@ -92,7 +92,7 @@ base:
   'roles:mongodb':
     - match: grain
     - mongodb
-    - fluentd.mongodb
+    - vector.mongodb
     - consul.mongodb
   mongodb*production*:
     - datadog.mongodb-integration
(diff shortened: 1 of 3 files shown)


blarghmatey:
I think the bigger question is being able to segment logs by environment to be able to try and identify any systemic issues that span systems in a given application. As long as the environment labels are enough for that then I think that combining the mongo logs is fine.

markbreedlove:
OK, great, thanks!

PR #1378, opened 2021-03-05
Add quotes to nginx app_metric referer field

markbreedlove:
Add quotes to the 'referer' field in the logs, or else it gets parsed incorrectly by Vector's built-in 'logfmt_parser' Transform when a URL has certain special characters.


blarghmatey:
:+1:

PR #1379, opened 2021-03-09
Vector: send MITx tracking logs to S3, not Elasticsearch

markbreedlove:
This changes the MITx Vector configuration to send the tracking logs to an S3 bucket, instead of Elasticsearch.

The logs are batched by day (or in 2 GB increments, whichever comes first).

The formatting results in folders and files named in this fashion:

```
2021-03-09/tracking-mitx-qa_2021-03-09.16:23:58_1615307099-663dd84c-1c8a-4aac-bc40-c7e8a788d543.log.gz
```

The unique identifiers are added by default by Vector's `aws_s3` sink to prevent collisions when multiple hosts are involved.

### DIFF
--- a/pillar/vector/mitx.sls
+++ b/pillar/vector/mitx.sls
@@ -432,15 +432,6 @@ vector:
         fields:
           time: "@timestamp"
 
-      tracking_log_field_adder:
-        inputs:
-          - tracking_log_timestamp_renamer
-        type: add_fields
-        fields:
-          labels:
-            - edx_tracking
-          environment: {{ salt.grains.get('environment') }}
-
       auth_log_parser:
         inputs:
           - auth_log
@@ -516,17 +507,23 @@ vector:
         index: logs-mitx-stderr-%Y.%W
         healthcheck: false
 
-      elasticsearch_tracking:
-        inputs:
-          - tracking_log_field_adder
-        type: elasticsearch
-        endpoint: 'http://operations-elasticsearch.query.consul:9200'
-        index: logs-mitx-tracking-%Y.%W
-        healthcheck: false
-
       elasticsearch_authlog:
         inputs:
           - auth_log_field_adder
         type: elasticsearch
         endpoint: 'http://operations-elasticsearch.query.consul:9200'
         index: logs-authlog-%Y.%W
+
+      s3_tracking:
+        inputs:
+          - tracking_log_timestamp_renamer
+        type: aws_s3
+        bucket: {{ salt.grains.get('environment') }}-tracking-logs
+        region: us-east-1
+        key_prefix: "%F/tracking-{{ salt.grains.get('environment') }}_%F.%T_"
+        encoding:
+          codec: ndjson
+        batch:
+          timeout_secs: {{ 60 * 60 * 24 }}
+          max_bytes: {{ 1024 * 1024 * 1024 * 2 }}
+        healthcheck: false
--- a/pillar/vector/mitx.sls
+++ b/pillar/vector/mitx.sls
@@ -536,9 +536,9 @@ vector:
         inputs:
           - tracking_log_timestamp_renamer
         type: aws_s3
-        bucket: {{ salt.grains.get('environment') }}-tracking-logs
+        bucket: odl-residential-tracking-data
         region: us-east-1
-        key_prefix: "%F/tracking-{{ salt.grains.get('environment') }}_%F.%T_"
+        key_prefix: "%F-%H_"
         encoding:
           codec: ndjson
         batch:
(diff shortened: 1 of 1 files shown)


shaidar:
Given the sensitivity/importance of those logs, we might want to consider shipping those every hour instead of per day

markbreedlove:
@shaidar That should be fine. I defer to you two about the interval. It would be easy to change, or to add a pillar property to control it.

pillar/vector/mitx.sls
```diff
@@ -516,17 +507,23 @@ vector:
...
           - auth_log_field_adder
         type: elasticsearch
         endpoint: 'http://operations-elasticsearch.query.consul:9200'
         index: logs-authlog-%Y.%W
+
+      s3_tracking:
+        inputs:
+          - tracking_log_timestamp_renamer
+        type: aws_s3
+        bucket: {{ salt.grains.get('environment') }}-tracking-logs
```
blarghmatey, line 521: We need to keep the S3 destination the same as it currently is, so they should be going to the buckets defined here: https://github.com/mitodl/salt-ops/blob/main/pillar/fluentd/server.sls#L6-L7

markbreedlove: OK, no problem. Would there by any issue with my changing the file naming scheme, or should I make sure the files are still named like `%Y-%m-%d-%H_[01].gz`? (Where `[01]` seems to be an identifier for the aggregator.)

The issue with that is we don't have Vector aggregators yet, and all the logs will be coming from many individual instances. If the format of the identifier doesn't matter and can be a UUID, this will be easy to implement.

pillar/vector/mitx.sls
```diff
@@ -516,17 +507,23 @@ vector:
...
+        region: us-east-1
+        key_prefix: "%F/tracking-{{ salt.grains.get('environment') }}_%F.%T_"
+        encoding:
+          codec: ndjson
+        batch:
+          timeout_secs: {{ 60 * 60 * 24 }}
```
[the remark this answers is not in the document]
markbreedlove: For some reason I'd misread the Trello card and had been thinking daily. It's easy to change to hourly.

pillar/vector/mitx.sls
blarghmatey: The identifier is arbitrary and doesn't matter so it should work fine with a UUID

markbreedlove: How is QA handled in the FluentD configuration above?

blarghmatey: It's not. For this setup it might be worth setting up a sampling rate to send ~5 or 10% of logs to S3 so we can smoke test it as new versions of Vector get deployed, etc.

blarghmatey: No, I meant send the QA tracking logs to a QA bucket, but only send ~5% of the log lines since we don't care as much about actually capturing that data.

pillar/vector/mitx.sls
```diff
@@ -530,3 +531,17 @@ vector:
...
+
+      s3_tracking:
+        inputs:
+          - tracking_log_timestamp_renamer
+        type: aws_s3
+        bucket: odl-residential-tracking-data
+        region: us-east-1
+        key_prefix: "%F-%H_"
+        encoding:
+          codec: ndjson
```
blarghmatey, line 543: Would it be possible to also Gzip the files that are uploaded?

pillar/vector/mitx.sls
```diff
@@ -530,3 +531,17 @@ vector:
         type: elasticsearch
         endpoint: 'http://operations-elasticsearch.query.consul:9200'
         index: logs-authlog-%Y.%W
+
+      s3_tracking:
+        inputs:
+          - tracking_log_timestamp_renamer
+        type: aws_s3
+        bucket: odl-residential-tracking-data
```
blarghmatey, line 539: The logs are currently uploaded to a `/logs` subdirectory.

pillar/vector/mitx.sls
markbreedlove: Yes, Vector Gzips them by default.

pillar/vector/mitx.sls
markbreedlove: :+1:

pillar/vector/mitx.sls
markbreedlove: Resolved in Slack call: have separate buckets for QA and production, but don't worry about sampling QA, since it doesn't have a lot of traffic.

blarghmatey:
:+1:

blarghmatey:
:+1:

PR #1380, opened 2021-03-10
Commenting out policy definitions to avoid overwrites

blarghmatey:
As we are transitioning to Pulumi we want to avoid having Salt overwrite any policy definitions that we are managing elsewhere. This comments out policy definitions from orchestrate scripts to avoid those collisions. The reason for commenting rather than deleting is to make it easier to identify and translate policies that need to be migrated.


shaidar:
👍

PR #1382, opened 2021-03-12
Convert Vector states to use Vector's new DSL

markbreedlove:
This updates the existing Vector configurations to use Vector's new Vector Remap Language, which has suddenly superceded the transforms that used to be available.

### DIFF
--- a/pillar/vector/mitx.sls
+++ b/pillar/vector/mitx.sls
@@ -450,7 +450,6 @@ vector:
         source: |
           parsed, err = parse_json(.message)
           if parsed != null {
-            .raw_message = .message
             del(.message)
             ., err = merge(., parsed)
             .labels = ["edx_tracking"]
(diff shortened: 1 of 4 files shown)


markbreedlove:
I'm going to have it preserve a `time` field in addition to `@timestamp` to make it easier to eventually change our index patterns and get away from the funny `@timestamp` nomenclature. Commit coming ...

blarghmatey:
Looks good :+1:

blarghmatey:
Looks good :+1:

markbreedlove:
Thanks. I'm sorry, but I forget exactly what we ended up deciding. I've added a "raw_message" field to the tracking log, but did we want to do that for any other ones? We'd discussed the storage implications, and I think we didn't want to do that for everything.

blarghmatey:
I think that for now we should skip on adding the raw message to tracking logs so as to avoid fouling any of the downstream consumer logic.

blarghmatey:
Let's merge this as is and then circle back on raw messages as we get further along with the rest of our logging.

PR #1384, opened 2021-03-16
Change tracking logs S3 check

markbreedlove:
Get rid of the malfunctioning Salt scheduled task and replace it with a shell script on cron.

Is this halfway sane?

We can add the xPro tracking logs bucket when we get that functioning for xPro again.


salt/master_utils/install_tracking_s3_monitor.sls
```diff
@@ -0,0 +1,24 @@
...
+        #!/usr/bin/env bash
+        BUCKETS="odl-residential-tracking-data"
+        # -e means fail the script on error ...
+        set -e
+        today=`date +"%F"`
+        for bucket in $BUCKETS; do
+            # ... and this will return error status for no matching objects:
+            aws s3 ls s3://$bucket/logs/$today
+        done
+        curl -s '{{ __vault__::secret-operations/global/healthchecks/mitx-tracking-s3>data>value }}'
```
blarghmatey, line 15: This needs to be passed as pillar data. The vault pillar renderer doesn't work in state files.

markbreedlove: Would the `master.config` pillar be the best place? Or one of the neighboring ones under `master`?

markbreedlove: I've made an attempt to fix it in commit 98a25d34
Can change it, if necessary!

blarghmatey:
:+1:

PR #1385, opened 2021-03-16
Bump up OCW_ITERATOR_CHUNK_SIZE on RC to avoid memory issues

mbertrand:
N/A

#### What's this PR do?
Bumps up `OCW_ITERATOR_CHUNK_SIZE` for open-discussions to 5000 (from 1000) to avoid reboots caused by excessive memory consumption when running the `backpopulate_ocw_data` command.


blarghmatey:
:+1:

PR #1386, opened 2021-03-18
Fix Vector auth log parsing

markbreedlove:
Fix Vector's auth log parsing, which was failing because the timestamps
did not have the year. Instead, we will use Vector's `parse_syslog()`
function, which figures out the year for us.


blarghmatey:
:+1:

PR #1387, opened 2021-03-23
Upgrade Elasticsearch and Kibana to Open Distro

markbreedlove:
This upgrades Elasticsearch and Kibana from Elastic.co's licence-restricted distribution to Open Distro.

We are working to reduce the time and energy we're investing in Saltstack, so this is done as simply as possible, with most of the work performed by a shell script.


blarghmatey:
Looks good

PR #1389, opened 2021-03-25
Improve Open Distro upgrade script and orchestration

markbreedlove:
- Fix the shell script so that it jumps through more hoops
- Update the orchestration so that it does a full cluster shutdown

I can update you with some of the crazy things I had to go through to make the upgrade work in QA, if you like.

Highlights:

* We can use the discovery-ec2 plugin for discovery... for now... until they break it or ES.co decides to make it not work with Open Distro. It's being pulled from the Es.co repository, I think.
* The uninstallation and re-installation of Java does not work the way a reasonable person would expect it to
* Index Lifecycle Management has been superceded by Open Distro's index management plugin, which is incompatible with ILM even though it does the same things. It has a different JSON syntax for configuration. We have to just delete ILM policies and recreate their equivalents ourselves in Kibana after the upgrade.
* If you don't delete the ILM settings from your index `_settings`, the "lifecycle" object gets moved to a new "archived" object in the settings ... which causes `400 Bad Request` errors when you try to update any settings on that index... and you can't delete it unless you use a little trick that is out-of-scope for this PR description. So we just delete the ILM settings before doing the upgrade.
* You have to delete the `.kibana` indices to avoid running into this awesome error when you start the cluster back up: `org.elasticsearch.bootstrap.StartupException: java.lang.IllegalStateException: unable to upgrade the mappings for the index [[.kibana_3/0
56zS2BBQtS-9vBzf2qTsQ]]` (Which renders the cluster unstartable.)

### DIFF
--- a/salt/elastic_stack/files/upgrade_es.sh
+++ b/salt/elastic_stack/files/upgrade_es.sh
@@ -2,9 +2,12 @@
 
 set -e
 
-systemctl stop elasticsearch
+systemctl stop elasticsearch || true
 cp -p /etc/elasticsearch/elasticsearch.yml /etc/elasticsearch/elasticsearch.yml.backup-pre-upgrade
+umount /var/lib/elasticsearch
 dpkg -P elasticsearch
+mkdir /var/lib/elasticsearch
+mount /var/lib/elasticsearch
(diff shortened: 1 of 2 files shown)


blarghmatey:
:+1:

PR #1390, opened 2021-03-30
Add Vector to XPro edX servers

markbreedlove:
Take the Vector configuration for MITx and make it useful for XPro, as well.


blarghmatey:
:+1:

PR #1392, opened 2021-04-01
Added Touchstone/SAML settings for OCW Studio

rhysyngsun:
Part of https://github.com/mitodl/ocw-studio/issues/62

#### What's this PR do?
Adds environment variables needed to get

Deploy to CI/RC and ensure the app still runs

### DIFF
--- a/pillar/heroku/ocw-studio.sls
+++ b/pillar/heroku/ocw-studio.sls
@@ -8,9 +8,12 @@
       'env_name': 'ci',
       'GTM_ACCOUNT_ID': 'GTM-5JZ7X78',
       'MAILGUN_SENDER_DOMAIN': 'ocw-ci.mail.odl.mit.edu',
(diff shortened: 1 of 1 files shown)


shaidar:
@rhysyngsun Looks good. Gonna request the certs from IS&T and once I have those in vault, will update you and we can get this merged and deployed.

blarghmatey:
For the SAML certs we just generate them and provide the public key to them. We don't need to have them create it.

rhysyngsun:
@shaidar I think we need to have the saml metadata API in place before we can contact IS&T, that's being added in https://github.com/mitodl/ocw-studio/pull/165

shaidar:
Yes we do as they'll have to make a request to it before they are able to add it. Working on the certs and adding to vault

shaidar:
👍

PR #1393, opened 2021-04-02
Use ocw-to-hugo version referenced by ocw-course-hugo-theme

noisecapella:
**Note**: do not merge until we add ocw-to-hugo as a dependency to ocw-course-hugo-theme first (https://github.com/mitodl/ocw-course-hugo-theme/pull/79)

None

#### What's this PR do?
Previously the ocwnext deployment script used the latest version of `ocw-to-hugo` from the `release-candidate` branch. To make the build process more explicit, we want to reference the version of `ocw-to-hugo` in `ocw-course-hugo-theme` specifically. This PR updates the build script to use that version of `ocw-to-hugo`, and remove some unused settings related to `ocw-to-hugo`.

TBD -- is this a necessary step for reviews in salt-ops?

#### Other context
I'm not sure what this line means for ocw-to-hugo or how this file is used in general. Should we just delete this block or is it still useful? https://github.com/mitodl/salt-ops/blob/main/salt/apps/ocw/nextgen_build_install.sls#L51

### DIFF
--- a/pillar/apps/ocw-next-production.sls
+++ b/pillar/apps/ocw-next-production.sls
@@ -4,7 +4,6 @@ ocw-next:
   ocw_to_hugo_bucket: ocw-to-hugo-output-production
   source_data_bucket: open-learning-course-data-production
   search_api_url: //open.mit.edu/api/v0/search/
-  ocw_to_hugo_git_ref: release
   ocw_www_git_ref: release
   ocw_course_hugo_starter_git_ref: release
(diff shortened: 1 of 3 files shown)


markbreedlove:
Regarding `nextgen_build_install.sls`: I will review that file and probably delete the state that you reference. I think it's no longer necessary, and predates the webhook shell script.

Regarding review steps: it's not ideal, but we'll have to either place a copy of the script from this PR on the QA server and try it out, or we will have to merge this and try it out on QA and then make further commits to fix things.

The changes look OK to me, without knowing exactly how the new commands work.

PR #1394, opened 2021-04-05
Fix admin API access, apps Elasticsearch cluster

markbreedlove:
Fix "localhost" access to all of the Elasticsearch APIs on the Apps Elasticsearch clusters. We've installed a new version of ReadonlyRest, which seems to work better with this new `hosts_local` property that we haven't used before.


blarghmatey:
:+1:

PR #1399, opened 2021-04-13
Removed sentry dsn from salt-master

shaidar:
#### What's this PR do?
Removed sentry config from salt-master as at times salt errors are causing a spike in sentry events thus depleting our subscription plan.


blarghmatey:
:+1:

PR #1400, opened 2021-04-16
remove obsolete RESTRICT_ENROLL_NO_ATSIGN_USERNAMES

pdpinch:
As far as I can tell, the code for this setting was removed in Ironwood, when we stopped using CAS.

none

#### What's this PR do?

remove unused ansible setting

Can a user with a Touchstone Collaborator account still self enroll in any course on Residential MITx? (Can Touchstone collaborator accounts log in at all any more?) 

IIRC, when we moved from CAS to SAML we had to switch to a different mechanism to prevent Touchstone Collaborators from self-enrolling in courses -- like, maybe https://github.com/mitodl/edx-platform/pull/228


blarghmatey:
:+1:

PR #1402, opened 2021-04-20
Set USE_X_FORWARDED_HOST to True for all xPRO environments

noisecapella:
None

#### What's this PR do?
Sets `USE_X_FORWARDED_HOST` to True for CI and RC environments for xPRO. The code for creating the Cybersource redirect URL uses the `Host` header which meant the redirect URL had a different domain, causing the user to need to login again on purchase (on RC). We need to set `USE_X_FORWARDED_HOST` so that the `X_FORWARDED_HOST` header is used instead, which contains the correct host value.

- Login to rc.xpro.mit.edu
- Go to https://rc.xpro.mit.edu/checkout/?product=90 and make a purchase. Use the fake Cybersource credit card information
- You should be redirected to a URL starting with `rc.xpro.mit.edu`.

### DIFF
--- a/pillar/heroku/xpro.sls
+++ b/pillar/heroku/xpro.sls
@@ -35,7 +35,6 @@
       'SHEETS_REFUND_ERROR_COL': 13,
       'SHEETS_REFUND_SKIP_ROW_COL': 14,
       'vault_env_path': 'rc-apps',
-      'USE_X_FORWARDED_HOST': True,
       'VOUCHER_COMPANY_ID': 1
       },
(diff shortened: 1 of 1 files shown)


shaidar:
Since it's true for all envs, I would just change [this](https://github.com/mitodl/salt-ops/blob/main/pillar/heroku/xpro.sls#L228) `True`

noisecapella:
@shaidar ready for another look

shaidar:
👍

PR #1406, opened 2021-04-27
Adding configuration for installing the sysadmin dashboard plugin

blarghmatey:


pillar/edx/ansible_vars/residential.sls
```diff
@@ -174,6 +174,7 @@ edx:
         RESTRICT_ENROLL_SOCIAL_PROVIDERS:
           - mit-kerberos
         ENABLE_THIRD_PARTY_ONLY_AUTH: True
+        ENABLE_SYSADMIN_DASHBOARD: False
```
shaidar, line 177: Why False?

blarghmatey: Because that turns off the built-in feature that's still in the edx-platform repository and lets the plugin take over.

shaidar:
👍

PR #1408, opened 2021-04-28
Add sentry dsn for ocw studio

markbreedlove:
This makes a small update to the Vault path for the Sentry DSN value for ocw-studio. It adds an `ocw-studio` purpose to the path, in case OCW adds more apps.

I've also hardcoded the string `open-courseware` that is used for the business unit, to make it a little quicker to read the file, since that value never changes.


blarghmatey:
:+1:

PR #1411, opened 2021-04-30
Added forum as a service in consul

shaidar:
#### What's this PR do?
This is related to getting the ['edx-username-changer`](https://github.com/mitodl/edx-username-changer) app to work in xPRO. The edx worker node uses the API to make a call to the forum service which is assumes is running locally. However, our worker instances are on separate nodes from the edxapp and so we need to have the forum service running on the edxapp node be accessible/reachable from the worker instances.


blarghmatey:
:+1:

PR #1414, opened 2021-05-10
Update pkg versions

shaidar:
#### What's this PR do?
Updates some of the os pkgs that are no longer available on focal and fail with the ansible run


blarghmatey:
:+1:

PR #1417, opened 2021-05-14
Removed iam role creation

shaidar:
#### What's this PR do?
When running orchestrate states, salt is messing with the existing roles managed by Pulumi now. This should address that issue.


blarghmatey:
:+1:

PR #1424, opened 2021-06-07
Added eox config to studio.yml

shaidar:
#### What's this PR do?
Adds eox config to studio.yml since those values are required in both lms and cms


blarghmatey:
:+1:

PR #1426, opened 2021-06-10
Removed datadog.http-check-integration as it's no longer used

shaidar:
#### What's this PR do?
Removes the `datadog.http-check-integration` as it's no longer and we have already deleted the file however we didn't remove it from the top pillar file.


blarghmatey:
:+1:

PR #1432, opened 2021-06-28
add GITHUB_WEBHOOK_BRANCH variables

gumaerc:
Related to https://github.com/mitodl/ocw-studio/issues/341

#### What's this PR do?
This PR adds the `GITHUB_WEBHOOK_BRANCH` variables to the RC and production config for `ocw-studio`.  They are used for filtering webhook signals sent to `ocw-studio` for updating starter configs based on a github repo, filtering it against the defined branch.

N/A


shaidar:
The key needs to be referenced under config vars so that it gets added there. I would also add one for CI which would be `master`

gumaerc:
@shaidar Thanks, I'll check that reference.  I hadn't added one to CI because the underlying code in `ocw-studio` is set up to fall back to the `default_branch` property reported by `PyGithub`.  The branch this is referring to isn't a branch of `ocw-studio` but rather a branch of any given repo on Github that has added a webhook to call `ocw-studio`.  The branch there might not be `master`, it might be `main`.  Therefore, I think it's safer to just leave it blank so it defaults to whatever gets reported as the default branch.

shaidar:
@gumaerc Without having a value for CI, pillar rendering would fail when it tries to resolve `env_data.GITHUB_WEBHOOK_BRANCH` as it's undefined

gumaerc:
@shaidar By :+1: did you mean to approve the PR?  Is this good to go?

shaidar:
@gumaerc Yeah that's our convention that it looks good and u can go ahead and merge it

PR #1433, opened 2021-06-30
Removed digital credential run from production

briangrossman:
#### What's this PR do?
Given that we the timeline for digital credentials has been delayed, removed the run that was set to support digital credentials. We will add a run back once things are ready.

thumb_when-you-are-on-time-rvc-j-www-rvcj-com-but-14452433


blarghmatey:
:+1:

PR #1438, opened 2021-07-14
More env variables for ocw-studio related to publishing notifications

mbertrand:
Related to https://github.com/mitodl/ocw-studio/pull/381

#### What's this PR do?
Adds more required vars for ocw-studio related to publishing notifications


pillar/heroku/ocw-studio.sls
```diff
@@ -51,6 +53,8 @@
       'GITHUB_WEBHOOK_BRANCH': 'release',
       'MAILGUN_SENDER_DOMAIN': 'ocw.mail.odl.mit.edu',
       'OCW_STUDIO_BASE_URL': 'https://ocw-studio.odl.mit.edu/',
+      'OCW_STUDIO_DRAFT_URL': '???',
+      'OCW_STUDIO_LIVE_URL': 'https://ocwnext.odl.mit.edu/',
```
mbertrand, line 57: Assuming `OCW_STUDIO_LIVE_URL` should equal the production URL, not sure what `OCW_STUDIO_DRAFT_URL` should be

pillar/heroku/ocw-studio.sls
```diff
@@ -70,6 +74,7 @@ heroku:
   api_key: __vault__::secret-operations/global/heroku/api_key>data>value
   config_vars:
     ALLOWED_HOSTS: '["*"]'
+    API_BEARER_TOKEN: __vault__::secret-open-courseware/ocw-studio/{{ environment }}/api-bearer-token>data>value
```
mbertrand, line 77: Made a guess here, it will have to be created in vault

pillar/heroku/ocw-studio.sls
blarghmatey: Where are these being used? Is this for the pipeline to set the base URL in the rendered HTML? The DRAFT_URL is whatever we are using as the preview domain. Also, we should have different values for RC vs production, and currently ocwnext.odl is the output of the single build server. Might be worth getting @pdpinch 's opinion on what we want the URL to be going forward, at least until we make the final cutover to ocw.mit

pillar/heroku/ocw-studio.sls
blarghmatey: What is this used for?

pillar/heroku/ocw-studio.sls
blarghmatey: Ignore the RC vs prod part of the comment, just looked at this in context.

mbertrand: These are the URL's where the draft and live outputs will be, which are used in the email notifications sent to site admins:  https://user-images.githubusercontent.com/187676/125663229-878bd030-a275-4d27-9b4a-86e14b6cc692.png

pillar/heroku/ocw-studio.sls
mbertrand: Authenticating against a new API webhook endpoint (`Authentication: Bearer <API_BEARER_TOKEN>`) to make sure it's coming from a legitimate source.  The pipeline will be sending a request to this endpoint on either success or failure to notify site admins.

blarghmatey: I'm assuming you mean an endpoint in the OCW site? Ideally we won't write that token into the pipeline definition itself and instead use the var syntax to fetch it from Vault instead.

mbertrand: I can make it work that way, but it needs to be available to OCW Studio too so that it can [make sure they match](https://github.com/mitodl/ocw-studio/pull/381/files#diff-16563ed2f0c515e79c29ce5e9eabd02fe39202da9bee751da6bdb640d1ab600bR330).

mbertrand: Let me know which vault var name it will be and I will update the pipeline definition in the OCW Studio PR to use it.

blarghmatey: I have not added the token value. What format is the app expecting for that? e.g. length requirements/constraints, UUID vs random string, etc.

As to the draft and live URL those aren't set currently but eventually that will be ocw.mit.edu for live. I'd defer to @pdpinch for what we want to use currently.

mbertrand: It will be a bearer token sent in the header so it can be any hard-to-guess string of any reasonable length.

mbertrand: A uuid would be fine but not strictly necessary

blarghmatey:
:+1:

PR #1439, opened 2021-07-15
make sure new vars are copied to heroku

mbertrand:
N/A

#### What's this PR do?
Make sure new vars are copied to heroku

### DIFF
--- a/pillar/heroku/ocw-studio.sls
+++ b/pillar/heroku/ocw-studio.sls
@@ -79,7 +79,7 @@ heroku:
     CONTENT_SYNC_PIPELINE: content_sync.pipelines.concourse.ConcourseGithubPipeline
     CONCOURSE_URL: {{ env_data.CONCOURSE_URL }}
     CONCOURSE_USERNAME: oldevops
-    CONCOURSE_PASSWORD: __vault__::secret-concourse/web>data>data>admin_password
+    CONCOURSE_PASSWORD: __vault__::secret-concourse/data/web>data>data>admin_password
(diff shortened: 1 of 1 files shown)


pillar/heroku/ocw-studio.sls
```diff
@@ -77,6 +77,7 @@ heroku:
     AWS_STORAGE_BUCKET_NAME: 'ol-ocw-studio-app-{{ env_data.env }}'
     CONTENT_SYNC_BACKEND: content_sync.backends.github.GithubBackend
     CONTENT_SYNC_PIPELINE: content_sync.pipelines.concourse.ConcourseGithubPipeline
+    CONCOURSE_URL: {{ env_data.CONCOURSE_URL }}
     CONCOURSE_USERNAME: oldevops
     CONCOURSE_PASSWORD: __vault__::secret-concourse/web>data>data>admin_password
```
mbertrand, line 82: This doesn't seem to be carried over to heroku, any idea why?  Is the value missing from vault?

blarghmatey: It's because the path isn't quite right. It needs to be `secret-concourse/data/web`

blarghmatey:
:+1:

pillar/heroku/ocw-studio.sls
mbertrand: Thanks, just changed it to ```
CONCOURSE_PASSWORD: __vault__::secret-concourse/data/web>data>data>admin_password
```

PR #1440, opened 2021-07-19
Remove ocw-to-hugo-hash.txt

noisecapella:
None

#### What's this PR do?
We are writing out three hash files, including `ocw-to-hugo-hash.txt`, which communicate the git hash of the deployed code. We used to use the latest `ocw-to-hugo` from the main branch of that repository. However we switched to `ocw-to-hugo` which is referenced and installed as a package in `ocw-hugo-themes`. The `ocw-to-hugo-hash.txt` file now serves no purpose.


blarghmatey:
:+1:

PR #1441, opened 2021-07-19
Run git clean -f as part of git_fetch_and_reset

noisecapella:
None

#### What's this PR do?
Run `git clean -f` when fetching ocw-www and ocw-hugo-themes so that we can ensure a clean slate on builds


blarghmatey:
:+1:

PR #1453, opened 2021-08-30
Added youtube oauth config

shaidar:
[Issue#486](https://github.com/mitodl/ocw-studio/issues/486)

#### What's this PR do?
This mainly deploys the configs requested in the issue linked above. For QA/RC/CI we will be using the existing ovs YouTube API in order not to go through configuring a new app with GCP and aren't concerned too much about hitting limits. For prod, the recommended approach is to the use the OCW account.


blarghmatey:
:+1:

PR #1454, opened 2021-09-07
Updating edX pillar to use our own Sentry plugin

blarghmatey:
Updating to use our tightly scoped plugin for Sentry integration on edX platform


shaidar:
👍

PR #1457, opened 2021-09-08
Error on failure to build ocw-www

noisecapella:
None

#### What's this PR do?
Causes the build to break if there was an error deploying ocw-www


shaidar:
👍

PR #1470, opened 2021-10-01
Added EMAIL_SUPPORT env value for mitxonline production

gsidebo:
https://github.com/mitodl/mitxonline/issues/241

#### What's this PR do?
Adds the correct env value for the `EMAIL_SUPPORT` setting in mitxonline production 

There isn't a CI/RC email address that I know of, hence the production-only setting


blarghmatey:
:+1:

PR #1481, opened 2021-11-08
THREEPLAY_CALLBACK_KEY

abeglova:
https://github.com/mitodl/ocw-studio/issues/752

#### What's this PR do?
Adds THREEPLAY_CALLBACK_KEY for https://github.com/mitodl/ocw-studio/pull/777

The value of THREEPLAY_CALLBACK_KEY can be set to any randomly generated secret string, as long as it's a different one for each environment


pillar/heroku/ocw-studio.sls
```diff
@@ -158,6 +158,7 @@ heroku:
     SOCIAL_AUTH_SAML_SP_PUBLIC_CERT: __vault__::secret-operations/{{ env_data.vault_env_path }}/{{ business_unit }}/saml>data>public_cert
     STATUS_TOKEN: __vault__:gen_if_missing:secret-{{ business_unit }}/{{ app }}/{{ environment }}/django-status-token>data>value
     THREEPLAY_API_KEY: __vault__::secret-operations/global/{{ business_unit }}/ocw-studio/threeplay_api_key>data>value
+    THREEPLAY_CALLBACK_KEY: __vault__::secret-operations/global/{{ business_unit }}/ocw-studio/threeplay_api_key>data>value
```
blarghmatey, line 161: If we just want an arbitrary random string then you can actually write it as `__vault__:gen_if_missing:secret-operations...` which will create an arbitrary random value at that location in Vault on first read.

blarghmatey:
:+1:

Issue #1482, opened 2021-11-08
ocw publish script should not upload ocw-to-hugo output


gumaerc:
Since the EC2 based OCW build servers are being deprecated, the webhook publish script should not upload `ocw-to-hugo` output to S3.  This task will be moved into its own Concourse pipeline and the output there will be specifically configured to be properly imported into `ocw-studio` and run on a more regular basis.

PR #1483, opened 2021-11-08
remove ocw-to-hugo output s3 upload

gumaerc:
Closes https://github.com/mitodl/salt-ops/issues/1482

#### What's this PR do?
This PR removes the step from OCW's `webhook-publish.sh.jinja` that uploads the `ocw-to-hugo` output to S3.  This task is being moved into its own Concourse pipeline so it can be run on a regular basis and be configured specfically for import into `ocw-studio`, rather than having to make its output compatible with both the EC2 server's deployment strategy and that.

N/A


blarghmatey:
:+1:

PR #1488, opened 2021-12-15
Updated xqueue-watcher configs to map to live/staging split

blarghmatey:
For our deployment of the Maple release we are splitting the live and staging
environments into separate VPCs. This means that for residential use of xqueue-watcher
we need to deploy a separate group of watchers for those courses that use the staging
environment. This updates the configuration to reduce the amount of extraneous logic so
that we can deploy dedicated watchers per environment.

- Gets rid of iteration over purposes that is difficult to understand
- Adds new environment definitions to include the staging environments of QA and
- Production
- Updates xqueue-watcher versions to match the latest code

### DIFF
--- a/pillar/edx/xqwatcher_600.sls
+++ b/pillar/edx/xqwatcher_600.sls
@@ -3,6 +3,13 @@
 {% set env_data = env_settings.environments[environment] %}
 {% set xqwatcher_venv_base = '/edx/app/xqwatcher/venvs' %}
 {% set python3_version = 'python3.8' %}
+{% set env_map = {
+  "mitx": "production",
+  "mitx-staging": "master"
+} %}
+
+{% set watcher_git_ref = env_map[environment.rsplit("-", 1)[0]] %}
+{% set env_prefix = environment.rsplit("-", 1)[-1] %}
 
 edx:
   xqwatcher:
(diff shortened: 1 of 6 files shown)


salt/environment_settings.yml
```diff
@@ -427,27 +365,29 @@ environments:
...
-            min_number: 6
-            max_number: 10
-            type: r5a.large
-          edx-worker:
-            min_number: 3
-            max_number: 7
-            type: t3a.large
-        security_groups:
-          - edx
+          xqwatcher: 41ef7dff8e73f2ef5a5ffee20c4a5b822eb95341
```
shaidar, line 379: ?

shaidar: Guessing that's a commit hash

salt/environment_settings.yml
```diff
@@ -427,27 +365,29 @@ environments:
...
+          edx_config_repo: https://github.com/edx/configuration
+          edx_config_version: open-release/maple.master
+        courses:
+          - name: 600x
+            num_instances: 2
```
shaidar, line 384: Do we need 2 for qa?

salt/environment_settings.yml
blarghmatey: Yeah, that's the hash of xqwatcher that we want to install

salt/environment_settings.yml
blarghmatey: Probably not. I think we can also remove some of the other watchers for courses that don't use staging.

pillar/edx/xqwatcher_600.sls
```diff
@@ -3,19 +3,24 @@
 {% set env_data = env_settings.environments[environment] %}
 {% set xqwatcher_venv_base = '/edx/app/xqwatcher/venvs' %}
 {% set python3_version = 'python3.8' %}
+{% set env_map = {
+  "mitx": "production",
+  "mitx-staging": "master"
+} %}
```
shaidar, line 9: Are those tags correct?

shaidar:
👍

PR #1489, opened 2022-01-07
Made changes related to removal of salt-master RDS DB

shaidar:
#### What's this PR do?
Deleted salt-master RDS DB and this is just to cleanup any code that uses that DB


blarghmatey:
:+1:

PR #1490, opened 2022-01-11
Switched digtial credentials auth mode to 'code'

rhysyngsun:
N/A

#### What's this PR do?
Changes the configured `auth_mode` for digital credentials to `"code"`.

I've already manually set this on RC and it's been verified by the DC team, so other than valid yaml it should be ok.


blarghmatey:
:+1:

PR #1491, opened 2022-01-13
Fix redis cluster name resolution

shaidar:
#### What's this PR do?
Running into issues on OVS where consul isn't able to resolve the hostname of the Redis service. As a temporary workaround, this change get the cluster-name and add it to the OVS config that is deployed on the EC2 instances.


blarghmatey:
:+1:

PR #1492, opened 2022-01-14
Adding default Vector config, removing Datadog, and general cleanup

blarghmatey:
- Adding a default Vector config to collect host metrics
- Removed all assignment of Datadog to host targeting for install and configuration
- Removed references for edX and MongoDB in top files since we are no longer running those via Salt


pillar/top.sls
```diff
@@ -223,54 +180,12 @@ base:
...
-    - vector.edx
-    - datadog.nginx-integration
-    - datadog.supervisord-integration
-  'P@roles:(edx|edx-worker)$and not G@roles:edx-analytics':
-    - match: compound
-    - vector.edx
-  'P@roles:(edx|edx-worker) and not G@edx_codename:tumbleweed':
-    - match: compound
-    - edx.ansible_vars.theme
   'G@roles:sandbox and P@environment:mitx(pro)?-qa':
```
shaidar, line 183: Should this be removed as well?

blarghmatey: Yeah, probably

salt/top.sls
```diff
@@ -181,52 +163,11 @@ base:
     - match: compound
     - elasticsearch
     - elasticsearch.plugins
-  'G@roles:elasticsearch and P@environment:(micromasters|production-apps)':
-    - match: compound
-    - datadog
-    - datadog.plugins
   'roles:mongodb':
```
shaidar, line 166: Do we still need this?

salt/top.sls
```diff
@@ -275,27 +216,13 @@ base:
     - match: compound
     - nginx
     - nginx.certificates
-  'P@roles:ocw-(cms|db|origin|mirror) and G@ocw-environment:production':
-    - datadog
   'roles:sandbox':
```
shaidar, line 219: We can probably remove this too

PR #1495, opened 2022-01-28
Move prod ocw-studio to github.mit.edu

mbertrand:
N/A

#### What's this PR do?
Moves git production settings for ocw-studio from github.com to github.mit.edu


blarghmatey:
:+1: once the app ID is updated

PR #1509, opened 2022-02-24
ocw next webhook keys

abeglova:
https://github.com/mitodl/ocw-studio/issues/1005

#### What's this PR do?
This PR adds OCW_NEXT_WEBHOOK_KEY to open-discussions and OPEN_DISCUSSIONS_WEBHOOK_KEY to ocw-studio. They can be any string but should be the same for each environment. I'm a little confused about how vault is organized - it seems like the paths in discussions.sls follow a different pattern than the paths in ocw-studio.sls. Please double check that I got them right

### DIFF
--- a/pillar/heroku/discussions.sls
+++ b/pillar/heroku/discussions.sls
@@ -199,6 +199,7 @@ heroku:
     OCW_LEARNING_COURSE_BUCKET_NAME: open-learning-course-data-{{ env_data.env_name }}
     OCW_NEXT_AWS_STORAGE_BUCKET_NAME: {{ env_data.OCW_NEXT_AWS_STORAGE_BUCKET_NAME }}
     OCW_NEXT_LIVE_BUCKET: {{ env_data.OCW_NEXT_LIVE_BUCKET }}
+    OCW_NEXT_SEARCH_WEBHOOK_KEY: __vault__::secret-{{ business_unit }}/global/update-search-data-webhook-key>data>value
     OCW_UPLOAD_IMAGE_ONLY: {{ env_data.OCW_UPLOAD_IMAGE_ONLY }}
     OLL_ALT_URL: https://openlearninglibrary.mit.edu/courses/
     OLL_API_ACCESS_TOKEN_URL: https://openlearninglibrary.mit.edu/oauth2/access_token/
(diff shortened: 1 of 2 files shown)


pillar/heroku/ocw-studio.sls
```diff
@@ -156,6 +156,7 @@ heroku:
     OCW_STUDIO_LOG_LEVEL: {{ env_data.OCW_STUDIO_LOG_LEVEL }}
     OCW_STUDIO_SUPPORT_EMAIL: {{ env_data.OCW_STUDIO_SUPPORT_EMAIL }}
     OCW_STUDIO_USE_S3: True
+    OPEN_DISCUSSIONS_WEBHOOK_KEY: __vault__::secret-operations/global/{{ business_unit }}/ocw-studio/open-discussions-webhook-key>data>value
```
shaidar, line 159: I would change this to `__vault__:gen_if_missing:secret-{{ business_unit }}/{{ app }}/{{ env_data.env }}/open-discussions-webhook-key>data>value`

pillar/heroku/discussions.sls
```diff
@@ -199,6 +199,7 @@ heroku:
     OCW_LEARNING_COURSE_BUCKET_NAME: open-learning-course-data-{{ env_data.env_name }}
     OCW_NEXT_AWS_STORAGE_BUCKET_NAME: {{ env_data.OCW_NEXT_AWS_STORAGE_BUCKET_NAME }}
     OCW_NEXT_LIVE_BUCKET: {{ env_data.OCW_NEXT_LIVE_BUCKET }}
+    OCW_NEXT_WEBHOOK_KEY: __vault__::secret-{{ business_unit }}/{{ env_data.env_name }}/ocw-next-webhook-key>data>value
```
shaidar, line 202: I would change this to ``__vault__:gen_if_missing:secret-{{ business_unit }}{{ env_data.env_name }}/ocw-next-webhook-key>data>value`

pillar/heroku/ocw-studio.sls
shaidar: Missed the point that they have to be the same value between apps. So would change it to `__vault__::secret-{{ business_unit }}/global/update-search-data-webhook-key>data>value`

I would also probably change the name of the key to be the same in both apps. Would probably make it easier down the line to identify that the value should be the same.

pillar/heroku/discussions.sls
shaidar: Missed the point that they have to be the same value between apps. So would change it to `__vault__::secret-{{ business_unit }}/global/update-search-data-webhook-key>data>value`.

I would also probably change the name of the key to be the same in both apps. Would probably make it easier down the line to identify that the value should be the same.

shaidar:
👍

PR #1515, opened 2022-03-07
Change to allow vector to run on older ubuntu systems that use upstart.

### LINKED ISSUE #1514: init.d startup configuration for vector on older linux releases
vector doesn't get registered as a service on some older linux distros by default. Need to write salt code to setup an init.d config to manage and run the agent. Shouldn't be too difficult and there is a ref implementation here: https://github.com/vectordotdev/vector/blob/master/distribution/init.d/vector

Ardiea:
Closes #1514 

#### What's this PR do?
Fixes the vector setup on reddit servers which run with upstart rather than systemd.

### DIFF
--- a/salt/vector/files/vector.conf
+++ b/salt/vector/files/vector.conf
@@ -1,0 +1,11 @@
+description "Event processing framework"
+author "Mike Davidson"
+start on filesystem or runlevel [2345]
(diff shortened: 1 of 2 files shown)


blarghmatey:
:+1:

PR #1525, opened 2022-06-08
add environment variable for open exchange rates

umarmughal824:
https://github.com/mitodl/mitxonline/pull/590

#### What's this PR do?
Add environment variable for open exchange rates

### DIFF
--- a/pillar/heroku/mitxonline.sls
+++ b/pillar/heroku/mitxonline.sls
@@ -84,6 +84,8 @@ heroku:
     MITX_ONLINE_SECURE_SSL_HOST: {{ env_data.MITXONLINE_SECURE_SSL_HOST }}
     MITX_ONLINE_USE_S3: True
     NODE_MODULES_CACHE: False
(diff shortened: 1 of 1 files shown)


pillar/heroku/mitxonline.sls
```diff
@@ -84,6 +84,8 @@ heroku:
     MITX_ONLINE_SECURE_SSL_HOST: {{ env_data.MITXONLINE_SECURE_SSL_HOST }}
     MITX_ONLINE_USE_S3: True
     NODE_MODULES_CACHE: False
+    OPEN_EXCHANGE_RATES_APP_ID: {{ env_data.OPEN_EXCHANGE_RATES_APP_ID }}
```
blarghmatey, line 87: You'll need to also add the concrete value of the app ID in the dictionary at the top of the file.

blarghmatey (changes requested):

pillar/heroku/mitxonline.sls
umarmughal824: I don't have the APP ID

PR #1526, opened 2022-06-09
update various ocw URLs

gumaerc:
N/A

#### What's this PR do?
This PR updates a number of URLs used when populating the env for OCW related apps like `ocw-studio` and `open-discussions`.  More specifically, uses of `ocwnext-rc.odl.mit.edu` and `ocwnext.odl.mit.edu` have been replaced with `live-qa.ocw.mit.edu` and `ocw.mit.edu` respectively.  The old domains have been left in the CORS allow list in `open-discussions` for the time being.

N/A


blarghmatey:
:+1: looks good!

gumaerc:
@blarghmatey I didn't update any of the documentation in `fastly_cdn.md` because I'm not sure exactly how y'all have Fastly configured when it refers to TLS domains, so maybe you could update that if it needs to be updated?

blarghmatey:
:+1:

PR #1529, opened 2022-07-07
enable infinite-corridor on ci and rc, not prod

ChristopherChudzicki:
For https://github.com/mitodl/open-discussions/issues/3613 and related work

#### What's this PR do?
This PR sets `ENABLE_INFINITE_CORRIDOR=True` on ci and prod, but False on product.

This env variable controls whether or not the `/infinite` page on open-discussions is served or not.

@abeglova Tested this as part of https://github.com/mitodl/open-discussions/pull/3601  :

> Check new ENABLE_INFINITE_CORRIDOR env variable:
> 
> 1. Visit http://yourname.od.odl.local:8063/infinite ... expect 404 ... *Correction: Actually a blank page... maybe we need a 404. But blank page is what open.mit.edu/woofwoofmeow shows, too.*
> 2. Set ENABLE_INFINITE_CORRIDOR=True in env file and docker compose restart watch
> 3. Visit http://yourname.od.odl.local:8063/infinite ... should work


pillar/heroku/discussions.sls
```diff
@@ -85,6 +87,7 @@
       'CLOUDFRONT_DIST': 'd2mcnjhkvrfuy2',
       'CORS_URLS': ['https://micromasters.mit.edu', 'https://ocwnext.odl.mit.edu', 'https://ocw-beta.odl.mit.edu', 'https://ocw-preview.odl.mit.edu', 'https://ocw-published.odl.mit.edu', "https://draft.ocw.mit.edu", "https://www.ocw.mit.edu", "https://ocw.mit.edu", 'https://live.ocw.mit.edu'],
       'DEBUG': False,
+      'ENABLE_INFINITE_CORRIDOR': False,
```
blarghmatey, line 90: This sets the values in the Jinja dictionary, but you'll still need to use it to set the key in the YAML dict that gets used for populating the Heroku settings.

PR #1531, opened 2022-07-27
Set MITX_ONLINE_SUPPORT_EMAIL for RC

asadali145:
Fixes https://github.com/mitodl/mitxonline/issues/729

#### What's this PR do?
Sets MITX_ONLINE_SUPPORT_EMAIL for RC.

Can check Value from the server.

MITX_ONLINE_SUPPORT_EMAIL environment variable was not set for RC. It turned out that this variable was only being used for the `production` environment due to a conditional block in the configs. In this PR, MITX_ONLINE_SUPPORT_EMAIL is being set outside the `production` environment conditional block.


Ardiea (approved):
Hi @asadali145! Looks good. In the future you can remove the optional blocks entirely from the description if they are not relevant and maybe provide a bit more context as to what is going on. Mostly just for future reference so someone coming back to this can quickly ascertain what is going on without having to reference the actual code. Thanks!

PR #1532, opened 2022-07-27
Added settings for MITx Online refunds

rhysyngsun:
https://github.com/mitodl/mitxonline/issues/750

#### What's this PR do?
This adds configuration to MITx Online to support the refund functionality via Google Sheets.

We'll test in RC first ([PR that uses these settings](https://github.com/mitodl/mitxonline/pull/723) and then we can configure production.

Note: there's some non-devops setup (namely creating the google sheets themselves) that will be required to be done first.


blarghmatey:
:+1:

PR #1533, opened 2022-08-11
Override MITx Online google sheets defaults

rhysyngsun:
Fixes https://github.com/mitodl/mitxonline/issues/825

#### What's this PR do?
This sets some settings that weren't set / were using defaults that don't work with how the spreadsheet is setup.

I've manually applied these changes to RC myself, so running this against that should produce no change (although maybe the first flag in this PR might change a bit).


blarghmatey:
:+1:

Ardiea (approved):

PR #1534, opened 2022-08-29
New open-discussions env variables for importing MITx Online courses

mbertrand:
https://github.com/mitodl/open-discussions/issues/3637
https://github.com/mitodl/open-discussions/issues/3636

#### What's this PR do?
Adds env variables required to import MITx Online courses into the open-discussions search index.

N/A

### DIFF
--- a/pillar/heroku/discussions.sls
+++ b/pillar/heroku/discussions.sls
@@ -197,10 +197,12 @@ heroku:
     MICROMASTERS_COURSE_URL: https://{{ etl_micromasters_host }}/api/v0/courseruns/
     MICROMASTERS_EXTERNAL_LOGIN_URL: https://{{ env_data.MICROMASTERS_BASE_URL}}/discussions
     MITPE_BASE_URL: https://professional.mit.edu/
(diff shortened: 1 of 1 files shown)


shaidar:
Since this only applies to production, might be best to filter on that with something like [this](https://github.com/mitodl/salt-ops/blob/2e7ba69a0deb9be47b90c2cc72f8287d6f5e8f2d/pillar/heroku/discussions.sls#L156):
```
{% if env_data.env_name == 'production' %}
```

mbertrand:
> Since this only applies to production, might be best to filter on that with something like this

I was thinking we'd want the env variables on RC too for testing purposes, to make sure the production mitx-online courses get imported properly.

shaidar:
In that case, I would just add it under [this](https://github.com/mitodl/salt-ops/blob/2e7ba69a0deb9be47b90c2cc72f8287d6f5e8f2d/pillar/heroku/discussions.sls#L156) conditional

shaidar:
👍

PR #1535, opened 2022-09-12
Updated open discussions salt stacks to point to AWS Opensearch.

Ardiea:
Relates to https://github.com/mitodl/ol-infrastructure/pull/990

#### What's this PR do?
This will point each heroku app at ES instances running in AWS Opensearch rather than our own managed EC2 instances.

In CI + QA

N/A

N/A

#### Screenshots (if appropriate)
N/A

N/A


pillar/heroku/discussions.sls
```diff
@@ -9,9 +9,9 @@
       'CLOUDFRONT_DIST': 'd28ic9ywb63ioi',
       'CORS_URLS': ['https://micromasters-ci.odl.mit.edu', 'https://ocwnext-rc.odl.mit.edu', 'https://draft-qa.ocw.mit.edu', 'https://live-qa.ocw.mit.edu'],
       'DEBUG': False,
-      'ELASTICSEARCH_INDEX': 'discussions-ci',
+      'ELASTICSEARCH_INDEX': 'discussions-rc',
```
Ardiea, line 12: We don't have a *-ci set of indices at the moment but doing this allows us to use the -rc data in both environments for now. We can probably changed this back to -ci and just run the reindex all command in CI to generate the data at the right names.

PR #1539, opened 2022-09-21
Added some additional settings for google sheets

rhysyngsun:
Related to changes for https://github.com/mitodl/mitxonline/issues/968

#### What's this PR do?
Adds some additional configuration settings to avoid some errors.

- Changes can be applied to RC, a value for `MITOL_GOOGLE_SHEETS_ADMIN_EMAILS` is not needed in RC
- In production, `MITOL_GOOGLE_SHEETS_ADMIN_EMAILS` does require a value, it should be a python list literal of email addresses. I _think_ that is supposed to be at least the service account's email address, but the documentation I have around this is a bit vague so we might need to trial-and-error this a bit.


blarghmatey:
:+1:

PR #1554, opened 2023-02-07
Added hostnames to bootcamps allowed_hosts

shaidar:
- Part of the work being done in this issue https://github.com/mitodl/hq/issues/426
- PR also removes `BOOTCAMP_ADMISSION_BASE_URL` as there is no reference to it in the code
#### What's this PR do?
Adds the hostnames to the ALLOWED_HOSTS config var

Tested this by making the change to all three applications

(Optional)

(Optional)

#### Screenshots (if appropriate)
(Optional)

(Optional)

### DIFF
--- a/pillar/heroku/bootcamps.sls
+++ b/pillar/heroku/bootcamps.sls
@@ -6,7 +6,7 @@
     'ci': {
       'app_name': 'bootcamp-ecommerce-ci',
       'env_name': 'ci',
-      'BOOTCAMP_ADMISSION_BASE_URL': 'https://admissions-test.herokuapp.com',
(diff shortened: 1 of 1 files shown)


pillar/heroku/bootcamps.sls
```diff
@@ -88,7 +91,7 @@ heroku:
   app_name: {{ env_data.app_name }}
   api_key: __vault__::secret-operations/global/heroku/mitx-devops-api-key>data>value
   config_vars:
-    ALLOWED_HOSTS: '["*"]'
+    ALLOWED_HOSTS: {{ env_data.ALLOWED_HOSTS }}
```
rhysyngsun, line 94: Will this correctly serialize as a string value instead of a list object?

PR #1562, opened 2023-05-08
config: add config for course run sync to be every hour (MITxOnline)

arslanashraf7:
https://github.com/mitodl/hq/issues/1211

#### What's this PR do?
- Adds values for cron job hours to run the job every hour. Previously it was using a value of `0` by default right from settings. Now it needs to be overridden in Heroku as *. We need to run the task every hour.

- To test it locally, you can run this task w.r.t minutes and check that the sync task runs every (Mins) that you've set. For this you can set the value for `CRON_COURSERUN_SYNC_HOURS` to be in decimal.
- The sync_courseruns should run every hour moving onward, but that's actually testable in xPRO.

(Optional)

We made a similar change in https://github.com/mitodl/salt-ops/pull/1560 for xPRO.


blarghmatey:
:+1:

PR #1563, opened 2023-05-08
fix: temporarily run the course run sync taks on just fridays(MITxPRO)

arslanashraf7:
Temporary fix for https://github.com/mitodl/hq/issues/1406

- edx course details API is returning incorrect/cached start-end dates for the courses and that's effecting the behavior on xPRO, The dates in xPRO sync every hour, and even when we set the right dates manually the hourly task overrides them with the wrong ones in edX.

#### What's this PR do?
- Sets the task frequency to be run on Friday only, So that meanwhile we can figure out what's going on on edX side. This will stop the task for some days, meanwhile, we can see what's going on in edX.

- Check that the value set is the readable and right config value for celery tasks.
- The task only runs on Fridays


arslanashraf7:
@cachob FYI

PR #1564, opened 2023-05-10
Revert "fix: temporarily run the course run sync taks on just fridays(MITxPRO)"

arslanashraf7:
Reverts mitodl/salt-ops#1563

In the original PR, we changed the course run sync frequency from every hour daily to just every hour on `Friday`. The reason for doing that was that we were facing issues with edX dates. The course details API of our edX instance was returning wrong/cached dates and for that reason, they were getting overridden in xPRO making the course dates to be out of sync.

Now, The course details API from our edX instance seems to work fine and we are in a position to revert this change.


arslanashraf7:
This PR can be reviewed, But I'm adding "Do not merge" label because @cachob wanted to keep an eye on the dates to see if they start malfunctioning again.

arslanashraf7:
@mitodl/devops could you take a look at this?

Ardiea (approved):

PR #1565, opened 2023-05-15
Adding ENV_NAME variable to ocw-studio

pt2302:
Part of https://github.com/mitodl/ocw-hugo-themes/issues/1125.

#### What's this PR do?
Adds the environment variable `ENV_NAME` to `ocw-studio`, which is necessary for differentiating between QA and production in defining the `noindex` meta tags to avoid indexing QA by search engines.

Verify that `ENV_NAME` is being defined correctly in Heroku, in particular to `rc` and `production`, as appropriate.


pdpinch:
Is this a new environment variable? Do we not have anything else that already plays the same role?  For example, how are we setting the "ENV" for sentry? 

image

Ardiea (approved):

Ardiea:
I'm not entirely sure where the sentry env names come from based on this file alone, but he is re-using a current piece of data for this new variable.

PR #1577, opened 2023-08-14
Vault path reconciliation for mitopen

Ardiea:
Cousin : https://github.com/mitodl/salt-ops/pull/1577

Refactoring for new value paths for mitopen and hard-coding 'mitopen' as the business unit in all paths rather than using a variable that never varies.

# Screenshots (if appropriate):

- [ ] Desktop screenshots
- [ ] Mobile width screenshots

### DIFF
--- a/pillar/heroku/mitopen.sls
+++ b/pillar/heroku/mitopen.sls
@@ -107,7 +107,7 @@ heroku:
     ENABLE_INFINITE_CORRIDOR: {{ env_data.ENABLE_INFINITE_CORRIDOR }}
     GA_G_TRACKING_ID: {{ env_data.GA_G_TRACKING_ID }}
     GA_TRACKING_ID: {{ env_data.GA_TRACKING_ID }}
-    GITHUB_ACCESS_TOKEN: __vault__::secret-mitopen/secrets>data>data>global>odlbot-github-access-token
+    GITHUB_ACCESS_TOKEN: __vault__::secret-operations/global/odlbot-github-access-token>data>value
     INDEXING_API_USERNAME: {{ env_data.INDEXING_API_USERNAME }}
     MAILGUN_FROM_EMAIL: 'MIT Open <no-reply@{{ env_data.MAILGUN_SENDER_DOMAIN }}'
     MAILGUN_KEY: __vault__::secret-operations/global/mailgun-api-key>data>value
@@ -127,7 +127,7 @@ heroku:
     OCW_NEXT_AWS_STORAGE_BUCKET_NAME: {{ env_data.OCW_NEXT_AWS_STORAGE_BUCKET_NAME }}
     OCW_NEXT_BASE_URL: {{ env_data.OCW_NEXT_BASE_URL }}
     OCW_NEXT_LIVE_BUCKET: {{ env_data.OCW_NEXT_LIVE_BUCKET }}
-    OCW_NEXT_SEARCH_WEBHOOK_KEY: __vault__::secret-mitopen/secrets>data>data>global>update-search-data-webhook-key
+    OCW_NEXT_SEARCH_WEBHOOK_KEY: __vault__::secret-operations/global/update-search-data-webhook-key>data>value
     OCW_UPLOAD_IMAGE_ONLY: {{ env_data.OCW_UPLOAD_IMAGE_ONLY }}
     OLL_ALT_URL: https://openlearninglibrary.mit.edu/courses/
     OLL_API_ACCESS_TOKEN_URL: https://openlearninglibrary.mit.edu/oauth2/access_token/
--- a/pillar/heroku/mitopen.sls
+++ b/pillar/heroku/mitopen.sls
@@ -153,7 +153,7 @@ heroku:
     MITOPEN_ENVIRONMENT: {{ env_data.env_name }}
     MITOPEN_FROM_EMAIL: MITOpen <mitopen-support@mit.edu>
     MITOPEN_FRONTPAGE_DIGEST_MAX_POSTS: 10
-    MITOPEN_JWT_SECRET: __vault__:gen_if_missing:secret-mitopen/secrets>data>data>jwt_secret
+    MITOPEN_JWT_SECRET: __vault__::secret-mitopen/secrets>data>data>jwt_secret
     MITOPEN_LOG_LEVEL: {{ env_data.app_log_level }}
     MITOPEN_SUPPORT_EMAIL: {{ env_data.MITOPEN_SUPPORT_EMAIL }}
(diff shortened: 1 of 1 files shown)


pillar/heroku/mitopen.sls
```diff
@@ -93,22 +92,22 @@ heroku:
...
+    EDX_API_CLIENT_ID: __vault__::secret-mitopen/secrets>data>data>edx-api-client>id
+    EDX_API_CLIENT_SECRET: __vault__::secret-mitopen/secrets>data>data>edx-api-client>secret
     EDX_API_URL: https://api.edx.org/catalog/v1/catalogs/10/courses
-    EMBEDLY_KEY: __vault__::secret-operations/{{ business_unit }}/embedly_key>data>value
+    EMBEDLY_KEY: __vault__::secret-operations/mitopen/embedly_key>data>value
     ENABLE_INFINITE_CORRIDOR: {{ env_data.ENABLE_INFINITE_CORRIDOR }}
     GA_G_TRACKING_ID: {{ env_data.GA_G_TRACKING_ID }}
     GA_TRACKING_ID: {{ env_data.GA_TRACKING_ID }}
-    GITHUB_ACCESS_TOKEN: __vault__::secret-{{ business_unit }}/global/odlbot-github-access-token>data>value
+    GITHUB_ACCESS_TOKEN: __vault__::secret-mitopen/secrets>data>data>global>odlbot-github-access-token
```
blarghmatey, line 110: We should probably just use `secret-operations` for any global values like this

pillar/heroku/mitopen.sls
```diff
@@ -128,12 +127,12 @@ heroku:
     OCW_NEXT_AWS_STORAGE_BUCKET_NAME: {{ env_data.OCW_NEXT_AWS_STORAGE_BUCKET_NAME }}
     OCW_NEXT_BASE_URL: {{ env_data.OCW_NEXT_BASE_URL }}
     OCW_NEXT_LIVE_BUCKET: {{ env_data.OCW_NEXT_LIVE_BUCKET }}
-    OCW_NEXT_SEARCH_WEBHOOK_KEY: __vault__::secret-{{ business_unit }}/global/update-search-data-webhook-key>data>value
+    OCW_NEXT_SEARCH_WEBHOOK_KEY: __vault__::secret-mitopen/secrets>data>data>global>update-search-data-webhook-key
```
blarghmatey, line 130: Again, for values shared across projects/apps we should just use the `secret-operations` mount

shaidar, line 130: Would probably remove `global`

pillar/heroku/mitopen.sls
```diff
@@ -154,12 +153,12 @@ heroku:
...
     MITOPEN_FROM_EMAIL: MITOpen <mitopen-support@mit.edu>
     MITOPEN_FRONTPAGE_DIGEST_MAX_POSTS: 10
-    MITOPEN_JWT_SECRET: __vault__:gen_if_missing:secret-{{ business_unit }}/jwt_secret>data>value
+    MITOPEN_JWT_SECRET: __vault__:gen_if_missing:secret-mitopen/secrets>data>data>jwt_secret
     MITOPEN_LOG_LEVEL: {{ env_data.app_log_level }}
     MITOPEN_SUPPORT_EMAIL: {{ env_data.MITOPEN_SUPPORT_EMAIL }}
     MITOPEN_USE_S3: True
     OPENSEARCH_DEFAULT_TIMEOUT: 30
-    OPENSEARCH_HTTP_AUTH: __vault__::secret-{{ business_unit }}/elasticsearch>data>http_auth
+    OPENSEARCH_HTTP_AUTH: __vault__::secret-mitopen/secrets>data>data>elasticsearch>http_auth
```
blarghmatey, line 161: `opensearch`?

pillar/heroku/mitopen.sls
```diff
@@ -168,17 +167,17 @@ heroku:
     PGBOUNCER_MAX_CLIENT_CONN: {{ env_data.PGBOUNCER_MAX_CLIENT_CONN }}
     PGBOUNCER_MIN_POOL_SIZE: {{ env_data.PGBOUNCER_MIN_POOL_SIZE }}
     PROLEARN_CATALOG_API_URL: https://prolearn.mit.edu/graphql
-    SECRET_KEY: __vault__:gen_if_missing:secret-{{ business_unit }}/django-secret-key>data>value
+    SECRET_KEY: __vault__:gen_if_missing:secret-mitopen/secrets>data>data>django-secret-key
```
blarghmatey, line 170: This won't work properly with nested values and `gen_if_missing`. Might be good to keep this as a separate key in the mount to avoid things getting accidentally clobbered.

Ardiea: Can I just get rid of `gen_if_missing` instead? It appears to just be an arbitrary alpha-numeric string.

blarghmatey: Yes, that is an option. The behavior of gen_if_missing was to simplify the work of populating secrets when we didn't actually care about the value.

Ardiea: I like that better. It is a one-time thing and I have a little function in my shell to generate random noise like this. I will replace the values for these three rather than carry them over from the existing mounts.

pillar/heroku/mitopen.sls
Ardiea: So, I'm happy to do that, but..... nothing under `secret-operations` is currently managed. That is, everything there appears to have been populated by hand. 

I can make us a managed space (probably a new mount rather than trying to pull `secret-operations` into management) for global+shared secrets and then begin populating those global+shared secrets with a `substructure` stack or something. 

I don't think it is a good idea to create a global+shared mount and then populate it with values willy-nilly from all kinds of various `application` stacks like this one, if that makes sense.

pillar/heroku/mitopen.sls
blarghmatey: There's also always the option of generating secrets with `python -c "import secrets; print(secrets.token_urlsafe())"`

pillar/heroku/mitopen.sls
blarghmatey: I think that is a very sensible argument and approach. We have a general need of migrating our existing static secret mounts into more holistic management, as well as moving off of `generic` and `kv1` mounts where possible, but that's a bigger effort than necessary for this specific need.

Ardiea: Okay I will populate these into `secret-operations` manually for now and open an issue to setup a `substructure` stack and begin populating it.

PR #1586, opened 2023-09-13
fix: remove trailing comma - ocw studio

HussainTaj-arbisoft:
None

Removes trailing comma from integer values. The comma makes it a string value which is not what the studio expects.

# How can this be tested?
This can be tested by deploying the config to RC.


ChristopherChudzicki (approved):
👍

PR #1597, opened 2023-10-23
Add MICROMASTERS_CATALOG_API_URL to mitopen env

mbertrand:
Related to https://github.com/mitodl/mit-open/pull/144

Adds `MICROMASTERS_CATALOG_API_URL` to env settings for mit-open, value based on this pre-existing line:

```
{% set etl_micromasters_host = salt.sdb.get('sdb://consul/open-{}-etl-micromasters-host'.format(environment)) %}
```

Devops question: Is `sdb://consul/open-{}-etl-micromasters-host` set for mit-open?  Should be the same as for open-discussions.


Ardiea:
> Devops question: Is `sdb://consul/open-{}-etl-micromasters-host` set for mit-open? Should be the same as for open-discussions.

This is a good question to which I do not have an answer yet. I'm not familiar with this syntax to begin with so I will look into it further.

Ardiea (commented):
👍

PR #1613, opened 2024-01-22
feat: add feature flag for enterprise page

asadali145:

A part of https://github.com/mitodl/hq/issues/1370

Enables the feature flag for the enterprise page on RC and CI.

### How can this be tested?

Can be tested by enabling the flag locally.


pillar/heroku/xpro.sls
```diff
@@ -102,6 +104,7 @@
       'MITXPRO_BASE_URL': 'https://xpro.mit.edu',
       'MITXPRO_SECURE_SSL_HOST': 'xpro.mit.edu',
       'ENABLE_ORDER_RECEIPTS': True,
+      'FEATURE_ENABLE_ENTERPRISE': False,
```
arslanashraf7, line 107: I think we can enable this as well. This would not be auto-deployed but once we test it on RC we can take these configs to prod without a new PR.

arslanashraf7 (changes requested):

pillar/heroku/xpro.sls
asadali145: Enabled in https://github.com/mitodl/salt-ops/pull/1613/commits/b7631f46f4b231874c648b04cb238694a944a251

arslanashraf7 (approved):
👍 I'll defer this to @mitodl/devops for the final approval.

blarghmatey (approved):

PR #1617, opened 2024-02-02
refactor: remove zendesk tags

asadali145:

A part of https://github.com/mitodl/hq/issues/3366

Removes the configurable values of ZENDESK_DOMAIN_VERIFICATION_TAG_VALUE as we have hardcoded.


arslanashraf7 (approved):
👍 . We might need to remove this from the Heroku manually In case these don't get removed from Heroku on an auto deployment.

PR #1618, opened 2024-02-05
Removing old pillar heroku app configs. Added a noteson where to find new config items. 

Ardiea:
https://github.com/mitodl/ol-infrastructure/issues/2074

Added docs around moved configs. Removed xpro and mitopen pillar configs.


blarghmatey:
This looks good, I'd just want to double check that we captured the change for xPRO that is causing the merge conflict.

